PR #5, opened 2026-05-06
Skip FA4 registration on B300 (SM 10.3)

antiagainst:
FA4's CUTE kernels only support SM 10.0 variants (B100/B200/GB200). On B300/GB300 (SM 10.3 / sm_103), the arch enum falls outside the [sm_100, sm_110f] range checked in FlashAttentionForwardSm100, causing an AssertionError at call time. Restrict registration to ArchVersion(10, 0) so B300 falls back to other available kernels.


zhyncs (commented):
https://github.com/Dao-AILab/flash-attention/issues/2376#issuecomment-4102205342

zhyncs (approved):

PR #6, opened 2026-05-06
perf(cache): overlap target and draft KV loadback independently 

LorrinWWW:
## Summary

When a request's KV is restored from the host KV store, attention reads each
layer behind a per-layer counter so the forward overlaps the host→device load.
Previously the draft model's KV shared the target's counter and was interleaved
on the load stream (base[0], draft[0], base[1], …), so the target's per-layer
waits resolved only after the draft load too — the target effectively waited
through the entire loadback, including the draft. That's cheap when the draft is
a small slice of the transfer (gpt-oss EAGLE3 ≈ 1/37) but costly for MLA-base +
MHA-draft models (Kimi-K2.5: the regular-attention draft is ≈32% of each
loadback), where the target needlessly stalls on the draft.

This gives the draft its own loadback pool and per-layer counter, loaded after
the base, so target and draft overlap their own loads independently and the
target is no longer gated on the draft. It's modeled on the right axis:
`CacheKind` stays the cache type (KV/Mamba) and target-vs-draft is an orthogonal
`is_draft` role, so the executor keys pools by id (`kv`, `kv.draft`, `mamba`, …)
and a draft of any kind needs no special kind; writeback is unchanged (draft KV
is written back by its target pool). On a Kimi-like profile this brings the
target's KV ~22–32% sooner in the loadback-bound regime (time-to-target-KV-ready
15.6ms → 12.1ms in a micro-benchmark) with identical total work; negligible for
small drafts. Validated by host-executor unit tests (13/13), the B200
prefix-cache-e2e suite (gpt-oss-20b, 3/3), and byte-identical loadback output.

E.g.
```
  time ─────────────────────────────────────────────────────────────────────▶

  BEFORE  (shared counter, draft interleaved after base[0])
    load: │b0│▓▓draft▓▓│b1│b2│b3│ ............. │bN-1│
                                                     ▲
          target KV ready ───────────────────────────┘   ← waits the WHOLE load,
                                                              incl. the fat draft

  AFTER  (this PR: draft = own pool/counter, loaded after base)
    load: │b0│b1│b2│b3│ ............. │bN-1│▓▓draft▓▓│
                                           ▲          ▲
          target KV ready ─────────────────┘          └─ draft KV ready
                                           ↑
                            target usable here — not gated on draft
          (≈22–32% sooner for MLA-base + MHA-draft; total work unchanged)
```


LorrinWWW:
~rebase make this PR tiny now, but it still looks like to be needed.~
The current main wait both layer 0 for draft and target, which essentially disabled layerwise overlapping.

PR #8, opened 2026-05-06
Add KV cache events to scheduler

Aphoh:
## Summary

Adds KV cache mutation events to the live C++ scheduler path. The scheduler now records reusable device prefix-cache `BlockStored` and `BlockRemoved` events, exposes them through `drain_kv_events()`, and the Python runtime publishes them through the existing ZMQ/msgpack KV event publisher when -`-kv-events-config` enables the stream.

Also adds the `--kv-events-config` CLI option, vLLM-style `enable_kv_cache_events` config handling, scheduler/native Python bindings for KV events, focused C++ and Python tests, and docs for the event config and wire format.

## Test Plan
The following all pass
- `pytest -q test/runtime/test_cli_config_compat.py::TestCLIConfigCompat::test_kv_events_config_arg test/
    runtime/test_kv_events.py`
- `tokenspeed_scheduler_tests --gtest_filter='KVPrefixCacheEventTestSuite.*:SchedulerKvCacheEventTestSuite.*'`
- Manual GPU smoke test with Llama-3.2-3B-Instruct, ZMQ KV events enabled, and a subscriber writing received event batches to `/tmp/kv_events_received.jsonl`.


python/tokenspeed/runtime/pd/kv_events.py
```diff
@@ -78,6 +78,48 @@ class KVEventBatch(EventBatch):
     events: list[Union[BlockStored, BlockRemoved, AllBlocksCleared]]
 
 
+def scheduler_kv_event_to_wire_event(
+    event: Any,
+) -> Union[BlockStored, BlockRemoved]:
+    """Translate a scheduler-native KV event into the ZMQ wire struct."""
+    if hasattr(event, "token_ids"):
```
LorrinWWW: The C++ binding already exposes kind ("BlockStored" / "BlockRemoved"). Shall we switch to `event.kind` and avoid hasattr? potentially there would be new events that have `token_ids`

python/tokenspeed/runtime/engine/scheduler_utils.py
```diff
@@ -67,6 +68,15 @@ def make_config(
...
     cfg.prefetch_threshold = prefetch_threshold
+    if enable_kv_cache_events:
+        if not hasattr(cfg, "enable_kv_cache_events"):
+            raise RuntimeError(
+                "KV cache events require a tokenspeed_scheduler extension with "
+                "SchedulerConfig.enable_kv_cache_events support."
+            )
+        cfg.enable_kv_cache_events = True
+    elif hasattr(cfg, "enable_kv_cache_events"):
+        cfg.enable_kv_cache_events = False
```
LorrinWWW: Maybe we don't need to be too defensive here if we have updated the scheduler already. Let's just do `cfg.enable_kv_cache_events = enable_kv_cache_events`
```suggestion
    cfg.enable_kv_cache_events = enable_kv_cache_events
```

LorrinWWW:
Let's fix lint as well. I'm good to move forward with the requested changes.

python/tokenspeed/runtime/pd/kv_events.py
Aphoh: done!

python/tokenspeed/runtime/engine/scheduler_utils.py
Aphoh: done!

Aphoh:
@LorrinWWW Lint/conflicts should be resolved. Let me know if there are any additional changes you'd like.

LorrinWWW:
Sorry just realized `BuildBlockEventsForNode` walks RootToLeaf(target) and re-hashes every ancestor page just to derive parent_hash, and Insert calls it once per newly-stored node, so inserting N new pages re-hashes the full prefix N times, O(N^2 * page_size) FNV mixes. But the disabled path is well gated, so this only bites when events are enabled. I'd be fine landing as-is with a TODO(perf), but the fix is also relatively: cache block_hash on TreeNode. I approved if you wanna defer - in that case we can leave an issue to track, or it would be great to turn around quickly. @Aphoh wdyt?

LorrinWWW (approved):

PR #10, opened 2026-05-07
MiniMax-M2 H100 FP8 optimizations

FlamingoPg:
## Summary

MiniMax-M2 H100 FP8 serving cleanup and optimization for the TP8/EP8 path.

This PR focuses on `MiniMaxAI/MiniMax-M2.5` and keeps MiniMax-M2.7 config compatibility covered. It does **not** include lazy prefill CUDA graph / PCG; that runtime change is intentionally split out for a separate review.

## Recommended Launch

```bash
python3 -m tokenspeed.api_server \
  MiniMaxAI/MiniMax-M2.5 \
  --quantization fp8 \
  --host 127.0.0.1 \
  --port 30000 \
  --gpu-memory-utilization 0.85 \
  --attn-tp-size 8 \
  --expert-parallel-size 8 \
  --attention-backend fa3 \
  --max-cudagraph-capture-size 32 \
  --reasoning-parser minimax-append-think \
  --enable-allreduce-fusion \
  --comm-fusion-max-num-tokens 4096 \
  --enable-symm-mem
```

Cache-free benchmark runs add:

```bash
  --no-enable-prefix-caching \
  --disable-kvstore
```

For MiniMax-M2.7, use the same command and replace the model path with `MiniMaxAI/MiniMax-M2.7` or a local checkpoint path.

Benchmark matrix used for regression:

- Prompt/output: `1k/8k`, `8k/1k`, `4k/4k`
- Batch sizes: `1, 2, 4, 8, 16, 32`
- Metrics: TTFT, TPOT, E2E

## What Changed

Correctness and compatibility:

- Keep MiniMax-M2 `lm_head` unquantized while transformer weights use FP8 block quantization.
- Preserve public MiniMax-M2.7 config metadata without wiring speculative/MTP metadata into main-model serving.
- Skip speculative layers structurally when checkpoint layer ids appear after `num_hidden_layers`.
- Fix MiniMax fused Q/K RMSNorm for strided Q/K views by passing Python-visible row strides into the CUDA binding.
- Fix FP8 per-token-group scale layout conversion for downstream MoE/GEMM use.
- Add `minimax-append-think` as an explicit reasoning parser option.

Performance:

- Add FlashInfer CUTLASS FP8 MoE backend and prefer it for supported H100 FP8 MoE layers; Triton remains the fallback.
- Add MiniMax biased grouped top-k Triton route kernel for the common `topk=8, group=1` shape.
- Add FA3 decode scheduler metadata reuse for plain MHA/GQA decode CUDA graph replay.
- Add FlashInfer SM90 small-M FP8 block-scale GEMM and cache selected dense FP8 GEMM kernels after weight loading.
- Keep FP32 router GEMM small-M gating inside the router wrapper instead of hard-coding it in model code.

## Review Hotspots

- MiniMax Q/K RMSNorm fused comm path and CUDA binding stride handling.
- FP8 MoE backend selection: FlashInfer CUTLASS primary path, Triton fallback preserved.
- Dense FP8 GEMM selection and the FlashInfer small-M path with internal activation quantization.
- MiniMax top-k routing registry traits and fallback behavior.
- FA3 decode scheduler metadata buffer lifecycle for CUDA graph replay.

## Accuracy Regression

| Run | Dataset | Num | Score | Notes |
| --- | ---: | ---: | ---: | --- |
| AIME25 sampling, server seed `82503480` | aime25 | 30 | 0.9000 | Matches the validated score |
| GSM8K deterministic, server seed `82503480` | gsm8k | 1319 | 0.8946 | Full EvalScope 4-shot CoT, `do_sample=false`, `temperature=0`, `max_tokens=1024` |

## Performance Regression

Current PR baseline was rerun after PCG was split out.

| Scope | Frameworks | TokenSpeed PCG? | Result |
| --- | --- | --- | --- |
| Current PR baseline | TokenSpeed only | No | Full cache-free matrix completed; TTFT/TPOT remain in the expected post-cleanup range |

## Follow-ups

- Lazy prefill CUDA graph / PCG.
- Scheduler prefill budget split: separate total scheduled prefill-token budget from per-request chunk cap.
- Broader AllReduce+Norm fusion is deferred; current profiling did not justify adding it to this PR.
- Further Triton MoE fallback cleanup/tuning can be reviewed separately from the FlashInfer CUTLASS primary path.

## Validation

- `git diff --check` passed.
- Changed Python files compile with `python -m py_compile`.
- H100 runtime, performance, and accuracy validation were run in the H100 container environment.


python/tokenspeed/runtime/execution/cuda_graph_wrapper.py
```diff
@@ -408,9 +408,13 @@ def _init_replay_metadata(
         seq_lens: torch.Tensor,
         req_to_page: torch.Tensor,
         forward_mode: ForwardMode,
+        mamba_pool_indices: torch.Tensor | None = None,
         **kwargs,
     ):
         """Graph-replay path — update persistent cuda-graph buffers in place."""
+        if self.input_buffers.has_mamba and mamba_pool_indices is not None:
+            kwargs["mamba_pool_indices"] = mamba_pool_indices
```
LorrinWWW: is there a reason to hack mamba metadata here?

FlamingoPg: Adding the concrete repro context here: I verified this on H100 with MiniMax-M2.5 using the PR launch shape (`--attention-backend fa3` + decode CUDA graph). MiniMax does not create a Mamba pool (`num_mamba_slots=0`), but without this guard the replay path still forwards `mamba_pool_indices=None` into `FlashAttentionBackend.init_forward_metadata_replay_cuda_graph()`, whose signature does not accept that kwarg. The first replay request then crashes with:

```text
TypeError: FlashAttentionBackend.init_forward_metadata_replay_cuda_graph() got an unexpected keyword argument 'mamba_pool_indices'
```

With the guard restored, the same smoke starts successfully and `/v1/chat/completions` returns 200.

So the intent is not to special-case MiniMax or mutate Mamba metadata. The replay wrapper should only build and forward Mamba-specific replay kwargs when the executor actually owns Mamba buffers; non-Mamba backends should not receive that kwarg at all. I agree the inline condition can read a bit ad-hoc, so the cleaner shape is to keep this filtering at the wrapper boundary (or behind a small helper) rather than making every non-Mamba backend accept and ignore arbitrary Mamba kwargs.

PR #20, opened 2026-05-07
Fix DeepSeek V3 tokenizer and fp8 KV guard

debo3:
## Summary

Fix two DeepSeek V3/V3.1 serving failure modes found while testing DeepSeek-V3.1-Terminus with the `tokenspeed_mla` backend.

First, DeepSeek V3-family tokenizer configs can advertise `LlamaTokenizerFast` even though the model ships a byte-level BPE `tokenizer.json`. Loading through `AutoTokenizer` can decode byte-level markers literally, producing output like `Ġattention` and `ĊĊ`. This PR routes DeepSeek V3 architectures through the verbatim fast-tokenizer path so decoding matches the tokenizer JSON.

Second, DeepSeek V3 with `tokenspeed_mla` and `--kv-cache-dtype fp8_e4m3` requires KV-cache scaling factors. Without them, the server can start but produce corrupted text. This PR adds an early validation error when that unsafe configuration is used without `--quantization-param-path`.

## Test Plan

Validated with focused regression tests:

```bash
source .venv/bin/activate
pytest -q test/runtime/test_hf_transformers_utils.py test/runtime/test_weight_loader_prefetch.py

### DIFF
--- a/python/tokenspeed/runtime/execution/weight_loader.py
+++ b/python/tokenspeed/runtime/execution/weight_loader.py
@@ -34,6 +34,13 @@ from tokenspeed.runtime.utils.torch_memory_saver_adapter import TorchMemorySaver
 
 logger = get_colorful_logger(__name__)
 
+_DEEPSEEK_V3_ARCHITECTURES = frozenset(
+    {
+        "DeepseekV3ForCausalLM",
+        "DeepseekV3ForCausalLMNextN",
+    }
+)
+
 
 class WeightLoader:
     """Handles model weight loading from disk.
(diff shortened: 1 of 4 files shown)


LorrinWWW:
Thanks for the PR. I wonder if the validation should happen in the attention backend / even kernel side, what do you think? @borontion

borontion:
> Thanks for the PR. I wonder if the validation should happen in the attention backend / even kernel side, what do you think? @borontion

yeah it will be better if we can validate this in tokenspeed_mla backend.

PR #22, opened 2026-05-07
perf: chunked-prefill prefix cache update for non-hybrid models

LorrinWWW:
## Summary

The commit broadens what was InsertHybridCache (gated on hybrid_cache != nullptr — Mamba/hybrid only) into InsertPrefixCache (gated on kv_prefix_cache != nullptr). Before this fix, plain transformer models never inserted mid-chunk KV pages into the radix tree during chunked prefill, only hybrid models did. The function was renamed and the guard moved so non-hybrid models now also get prefix-cache updates per chunk.

```
  begin plan:
    snapshot prefix-cache matches for first-chunk candidates
    schedule/apply requests, build ops
    publish completed chunks as usual
    newly published chunks are visible to the next plan's snapshot
  return FlatForwardOperation
```


LorrinWWW:
@XucSh Can you take a look for this one? thx!

XucSh:
Sorry I missed this PR! It looks like this performance optimization is necessary.

LorrinWWW:
I'll update to latest main

LorrinWWW:
@codex review

LorrinWWW:
duplicated to #381

PR #24, opened 2026-05-08
kv_cache_host: AMD per-layer loop for KVStore device→host backup (fixes MI355X fault)

andyluo7:
## Summary

Without this patch, **the KVStore (host-side L2 KV cache) faults at the HSA layer on AMD MI355X (gfx950 / CDNA4) under any multi-request workload**, killing the server within ~10 seconds. KVStore + prefix caching are both on by default, so AMD users currently have to run with `--disable-kvstore --no-enable-prefix-caching` and lose those features entirely.

This PR routes the AMD path through the existing single-layer kernel in a Python loop. The NVIDIA path is byte-identical.

## Symptom

```
Memory access fault by GPU node-3 (Agent handle: 0x...) on address 0x...
GPU coredump: Directory \"/coredumps not writable or does not exist
GPU core dump failed
Fatal Python error: Aborted
```

Reproduces with default flags within seconds of any multi-request workload (e.g. `tokenspeed bench serve --num-prompts 32`).

## Root cause (triage)

Triage with `HIP_LAUNCH_BLOCKING=1 AMD_SERIALIZE_KERNEL=3 AMD_LOG_LEVEL=2` traced the fault to:

```
event_loop.py:1084  run_event_loop
event_loop.py:949   event_loop_overlap
event_loop.py:531   _submit_cache_ops
memory_executor.py:195  submit_plan
host_executor.py:259    flush
host_executor.py:285    _start_writing
kv_cache_host.py:378    backup_from_device_all_layer        <-- this PR
ops/kvcache/triton.py:385  transfer_kv_all_layer
ops/kvcache/triton.py:_kv_transfer_all_layer_kernel        <-- root cause
```

`_kv_transfer_all_layer_kernel` uses a tensor-of-pointers indirection (`tl.load(k_ptr_src_ptr + layer).to(tl.pointer_type(tl.uint32))`) to iterate over per-layer KV tensors. Combined with writes to host-pinned memory mapped into the GPU VA space, this pattern appears to mistranslate in the Triton AMD backend for gfx950 — the resulting GPU kernel issues a memory access that HSA flags as a fault.

The NVIDIA-only optimized variant (`_kv_transfer_all_layer_cs32_kernel`) uses inline ASM and is gated by `_is_nvidia`, so AMD always traverses the buggy path. Stripping the kernel's `cache_modifier=\".cg\"`/`\".cs\"` hints (an early hypothesis) did not help; the indirection itself is the issue.

A proper long-term fix should address the underlying Triton kernel so the all-layer form works on AMD too. This PR is the safe interim fix: keep the NVIDIA fast path, route AMD through the per-layer kernel that already exists in the same module and uses straight tensor pointers.

## Patch

```python
if _is_nvidia:
    transfer_kv_all_layer(...)            # unchanged on NVIDIA
else:
    # AMD: per-layer Python loop avoids the all-layer Triton kernel
    # that faults on MI355X (pointer-of-pointer indirection +
    # writes to host-pinned memory). Each per-layer kernel call
    # uses straight tensor pointers and works fine.
    for _layer_id in range(self.layer_num):
        transfer_kv_per_layer(
            src_k=device_pool.get_key_buffer(_layer_id),
            dst_k=self.k_data_refs[_layer_id],
            src_v=device_pool.get_value_buffer(_layer_id),
            dst_v=self.v_data_refs[_layer_id],
            src_indices=device_indices,
            dst_indices=host_indices,
            item_size=self.token_stride_size,
        )
```

### MI355X (mia1-p01-g06, lmsysorg/sglang-rocm:v0.5.11-rocm720-mi35x-20260505)

`gpt-oss-120b` MXFP4, TP=1, all defaults (KVStore + prefix caching ON), CUDA graphs ON:

| Concurrency | Output tok/s | TPOT median | TTFT mean | Failed |
|---:|---:|---:|---:|---:|
| **before** (any concurrency) | crash within ~10s | — | — | all |
| 32 | 1818.6 | 16.12 ms | 413 ms | 0 |
| 64 | 2362.1 | 23.73 ms | 532 ms | 0 |
| 128 | 2567.9 | 24.68 ms | 1327 ms | 0 |

Peak output throughput at c=128 = 5105 tok/s.

### B200 (Shadeform, lightseekorg/tokenspeed-runner:cu130-torch-2.11.0)

Server boots and serves correctly with this patch applied. The `_is_nvidia` branch is taken; the original `transfer_kv_all_layer` call runs unchanged. Verified with TinyLlama-1.1B-Chat producing coherent generations.

## Test plan

- [ ] CI passes (lint + tests)
- [ ] On NVIDIA: serve any model — confirm no behavior change vs `main`
- [ ] On AMD: `tokenspeed bench serve --num-prompts 32 --input-len 512 --output-len 256` — confirm 0 failed requests (was: crash)

## Related

Likely the same root cause as the existing skip note in `test/runtime/test_inline_detokenizer_e2e.py` (\"GPU memory access fault inside `reset_valid_cache_length` on AMD MI355X\"). Both fault at the same call class (host-pinned memory access from a Triton kernel with pointer-array indirection). Following up upstream is left to a separate PR (probably needs Triton-AMD-backend coordination).

### DIFF
--- a/python/tokenspeed/runtime/cache/kv_cache_host.py
+++ b/python/tokenspeed/runtime/cache/kv_cache_host.py
@@ -375,6 +375,17 @@ class MHATokenToKVPoolHost(HostKVCache):
     ):
         if io_backend == "kernel":
             if self.layout == "layer_first":
+                # Pass per-layer tensor refs alongside the pointer arrays.
+                # transfer_kv_all_layer routes through them on AMD where the
+                # all-layer Triton kernel's pointer-of-pointer indirection
+                # faults against host-pinned destinations. NVIDIA path is
+                # byte-identical and ignores the *_refs kwargs.
+                _src_k_refs = [
+                    device_pool.get_key_buffer(_i) for _i in range(self.layer_num)
+                ]
+                _src_v_refs = [
+                    device_pool.get_value_buffer(_i) for _i in range(self.layer_num)
+                ]
                 transfer_kv_all_layer(
                     src_k_layers=device_pool.k_data_ptrs,
                     dst_k_layers=self.k_data_ptrs,
@@ -384,6 +395,10 @@ class MHATokenToKVPoolHost(HostKVCache):
                     dst_indices=host_indices,
                     item_size=self.token_stride_size,
                     num_layers=self.layer_num,
+                    src_k_layer_refs=_src_k_refs,
+                    dst_k_layer_refs=self.k_data_refs,
+                    src_v_layer_refs=_src_v_refs,
+                    dst_v_layer_refs=self.v_data_refs,
                 )
             elif self.layout == "page_first":
                 transfer_kv_all_layer_lf_pf(
--- a/tokenspeed-kernel/python/tokenspeed_kernel/ops/kvcache/triton.py
+++ b/tokenspeed-kernel/python/tokenspeed_kernel/ops/kvcache/triton.py
@@ -326,6 +326,10 @@ def transfer_kv_all_layer(
     item_size: int,
     num_layers: int,
     block_quota: int | None = None,
+    src_k_layer_refs: list | None = None,
+    dst_k_layer_refs: list | None = None,
+    src_v_layer_refs: list | None = None,
+    dst_v_layer_refs: list | None = None,
 ) -> None:
     """
     Transfer KV cache entries for all layers based on src/dst indices.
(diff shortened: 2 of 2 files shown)


python/tokenspeed/runtime/cache/kv_cache_host.py
```diff
@@ -375,16 +375,32 @@ def backup_from_device_all_layer(
...
-                    dst_k_layers=self.k_data_ptrs,
-                    src_v_layers=device_pool.v_data_ptrs,
-                    dst_v_layers=self.v_data_ptrs,
-                    src_indices=device_indices,
-                    dst_indices=host_indices,
-                    item_size=self.token_stride_size,
-                    num_layers=self.layer_num,
-                )
+                if _is_nvidia:
+                    transfer_kv_all_layer(
```
borontion: can we fix this in `transfer_kv_all_layer` kernel inside of runtime?

andyluo7:
@borontion thanks for the steer — agreed the dispatch belongs in `transfer_kv_all_layer`, not the runtime. Just force-pushed a rewrite that does this:

- `transfer_kv_all_layer` gains four optional kwargs (`{src,dst}_{k,v}_layer_refs`) — lists of per-layer tensors. When on AMD AND refs are provided, dispatches via the existing `transfer_kv_per_layer` in a Python loop. NVIDIA path falls through unchanged.
- `MHATokenToKVPoolHost.backup_from_device_all_layer` just supplies the refs alongside the existing pointer arrays. No platform check at the runtime layer anymore.

Re-validated on MI355X (mia1-p01-g06) with default flags (KVStore + prefix caching ON):
- c=32: 1432 tok/s, 0 failed
- c=64: 2290 tok/s, 0 failed (total 11.4k tok/s)
- c=128: 2055 tok/s, 0 failed (total 10.3k tok/s)
- Single-stream output bit-identical to `mha` backend

Slight perf regression vs the previous call-site version (c=32: 1818→1432, c=128: 2568→2055) — the per-call overhead of building the Python ref lists costs more than the in-line branch did. Open to options: cache the lists at host-pool init time, hoist the `[device_pool.get_*_buffer(i) for i in ...]` to once per allocation, etc. — happy to follow up if you have a preferred shape.

The longer-term fix is still to address the underlying Triton AMD codegen issue for stores to host-pinned memory via pointer-of-pointer indirection — that needs Triton-AMD-backend coordination and isn't a one-line change. This PR keeps AMD users unblocked in the meantime.

borontion:
I can reproduce this on mi355. I have opened a fix: https://github.com/lightseekorg/tokenspeed/pull/36

PR #27, opened 2026-05-08
attention/triton: fix decode + prefill API drift; integrate cudagraphs

andyluo7:
## Summary

`--attention-backend triton` was 100% broken at HEAD: both `decode_attention_fwd` and `prefill_attention_fwd` had been refactored to a 2D paged-table KV layout, but the call sites in this backend were never updated. Both decode and prefill threw `TypeError` immediately.

This PR brings the backend in line with the new kernel signatures and wires it into the cuda-graph machinery. Validated end-to-end on both AMD MI355X and NVIDIA B200.

## Symptoms (before this PR)

```
decode:  TypeError: decode_attention_fwd() missing 2 required positional
                    arguments: 'window_left' and 'sm_scale'
prefill: TypeError: prefill_attention_fwd() got multiple values for
                    argument 'sliding_window_size'
```

NVIDIA users typically don't notice because their default attention backend is `flashinfer`/`trtllm`. AMD users hit it as soon as they explicitly select the cross-platform `triton` backend.

### 1. ForwardMetadata gains `page_table` + `seq_lens` (default `None`)

Backward-compatible — the existing positional-arg constructor calls keep working.

### 2. `init_forward_metadata` builds the page_table

Both the eager and cuda-graph-capture init paths now call the existing `build_page_table(req_pool_indices[:bs], _req_to_token, self.page_size, self.max_context_len)` from `tokenspeed.runtime.layers.attention.utils` (same utility the working `mha` backend uses).

### 3. `forward_decode` / `forward_extend` call sites updated

Decode now passes `(page_table, cache_seqlens, ..., page_table_stride_b, page_size, window_left, sm_scale, logit_cap=, sinks=)` to match the new kernel signature.

Prefill converts `kv_indptr` → `cache_seqlens` via `(kv_indptr[1:] - kv_indptr[:-1]).to(torch.int32)`, threads in `page_table`, and reorders args to match the prefill kernel's new signature (`cu_seqlens_q, cache_seqlens, custom_mask, is_causal, max_len_extend, ...` plus kwargs for `sm_scale`, `logit_cap`, `sliding_window_size`, `sinks`, `page_table`, `page_table_stride_b`, `page_size`, `has_kv_cache`).

### 4. CUDA graph integration

Naive use of `build_page_table` inside the capture path captures a transient tensor address that the replay path never refreshes — the captured graph then reads stale `page_table` contents at replay and produces **silently incorrect outputs** (verified on B200 before fixing).

Fix:
- `init_cuda_graph_state` allocates static `cuda_graph_page_table[max_bs, max_pages]` and `cuda_graph_seq_lens[max_bs]` buffers
- The capture path `.copy_()`s into the static buffers and stores slices in `ForwardMetadata` (so the captured graph references the static buffer's address)
- `init_forward_metadata_replay_cuda_graph` refreshes the same buffers in-place at every replay step
- Same method also gains `**kwargs` so it accepts `mamba_pool_indices` (and other args the cuda-graph wrapper now passes), fixing a separate `unexpected keyword argument` error

### AMD MI355X (mia1-p01-g06, gfx950 / CDNA4, lmsysorg/sglang-rocm:v0.5.11-rocm720-mi35x-20260505)

| | Output tok/s | TPOT median | Failed |
|---|---:|---:|---:|
| **before** (any config) | crash on first request | — | — |
| triton + eager | 914 | 33.4 ms | 0 |
| **triton + cudagraph** | **1795** | **16.18 ms** | 0 |
| (mha + cudagraph for reference) | 1818 | 16.12 ms | 0 |

`gpt-oss-120b` MXFP4, TP=1, c=32, 512in/256out. Triton with cudagraphs is at parity (~99%) with the `mha` backend.

Correctness check: both backends produce bit-identical output on the same prompt (`"Hello, my name is"` → `" John Doe and I am 30 years old.\"\n\`\`\`\n\nIn this example, ..."`).

### NVIDIA B200 (Shadeform, lightseekorg/tokenspeed-runner:cu130-torch-2.11.0)

- Server boots cleanly with `--attention-backend triton` + cudagraphs ON
- `TinyLlama-1.1B-Chat` produces coherent outputs end-to-end
- `current_platform()` + `_get_default_backend_name()` unchanged (triton is not the default on NVIDIA; this PR only fixes it when explicitly selected)

## Test plan

- [ ] CI passes (lint + tests)
- [ ] On NVIDIA: explicit `--attention-backend triton` selection works (was broken)
- [ ] On AMD: eager + cudagraph both work end-to-end with correct outputs
- [ ] Bench numbers reproduce on both platforms

### DIFF
--- a/python/tokenspeed/runtime/layers/attention/backends/triton.py
+++ b/python/tokenspeed/runtime/layers/attention/backends/triton.py
@@ -41,6 +41,7 @@ from tokenspeed.runtime.layers.attention.backends.base import AttentionBackend
 from tokenspeed.runtime.layers.attention.configs.mha import MHAConfig
 from tokenspeed.runtime.layers.attention.registry import register_backend
 from tokenspeed.runtime.layers.attention.utils import (
+    build_page_table,
     create_flashinfer_kv_indices_triton,
 )
 
@@ -67,6 +68,9 @@ class ForwardMetadata:
     window_kv_indptr: torch.Tensor
     window_kv_indices: torch.Tensor
     window_num_kv_splits: torch.Tensor
+    # Bug 1 fix: 2D paged-table form for new decode_attention_fwd signature
+    page_table: torch.Tensor = None  # [B, max_pages] of page IDs
+    seq_lens: torch.Tensor = None  # [B] cache_seqlens for decode
 
 
 class TritonAttnBackend(AttentionBackend):
@@ -376,6 +380,20 @@ class TritonAttnBackend(AttentionBackend):
             max_extend_len = torch.max(_extend_seq_lens).item()
             num_kv_splits = None
 
+        # Bug 1 fix: build page_table + seq_lens for new decode_attention_fwd kernel.
+        # Only useful in decode_or_idle modes; for extend/prefill modes the values are
+        # carried as None and the prefill kernel doesn't need them.
+        try:
+            _page_table = build_page_table(
+                req_pool_indices[:bs],
+                _req_to_token,
+                self.page_size,
+                self.max_context_len,
+            )
+            _seq_lens = seq_lens[:bs] if seq_lens is not None else None
+        except Exception:
+            _page_table = None
+            _seq_lens = None
         self.forward_metadata = ForwardMetadata(
             attn_logits,
             attn_lse,
@@ -389,6 +407,8 @@ class TritonAttnBackend(AttentionBackend):
             window_kv_indptr,
             window_kv_indices,
             window_num_kv_splits,
+            page_table=_page_table,
+            seq_lens=_seq_lens,
         )
 
     def init_cuda_graph_state(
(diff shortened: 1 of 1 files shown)


borontion:
@andyluo7 we prefer using the unified MHA backend instead of triton backend if need to use triton kernels.

andyluo7:
Closing per @borontion's earlier guidance — the unified `mha` backend is the path forward, not the `triton` backend. Validated `mha` works end-to-end on MI355X with cudagraphs (1818 tok/s output, 0 failed at c=32 for gpt-oss-120b MXFP4) so AMD users can switch with no functional gap.

Leaving the diff here for future reference if anyone needs to keep `triton` running for any reason — the decode/prefill API drift fixes + cudagraph static-buffer integration are independent of the MHA-vs-Triton-backend choice. Happy to revive if useful, otherwise this can stay closed.

If `triton` should be removed from `--attention-backend` choices entirely (since it's broken on `main` today and the direction is to deprecate it), happy to spin a small follow-up PR for that.

PR #30, opened 2026-05-08
feat(deepseek-v4): add mega_moe and compressed KV perf path

dongjiyingdjy:
## Summary
- Add the PR2 DeepSeek V4-Flash performance path on top of the merged baseline.
- Enable DeepGEMM mega_moe experts, fast mHC fused kernels, ratio-aware compressed KV cache layout, and V4 compressed attention/indexer helpers.
- Default V4 block_size to 256 when the user keeps the ServerArgs class default, while respecting explicit non-default values.
- Add DeepSeek V4 serving docs and focused runtime/kernel tests.
- Keep MXFP4 e8m0 scale bits intact while avoiding Triton swizzle padding overflow on CI runners.

## Validation
- pre-commit run --all-files
- pytest test/runtime/test_deepseek_v4_config.py test/runtime/test_deepseek_v4_attention_ops.py test/runtime/test_deepseek_v4_mega_moe.py -q: 54 passed, 6 skipped
- git diff --check
- GSM8K 5-shot limit=50 via local-completions, temperature=0: exact_match 0.96 ± 0.028

## Notes
- This is PR2 of the DeepSeek V4 stack, rebased onto latest main.
- Debug-only docs/scripts are not included.


python/pyproject.toml
```diff
@@ -65,6 +65,7 @@ dependencies = [
     "requests",
     "setproctitle",
     "tiktoken",
+    "tilelang==0.1.9",
```
zhyncs: Let's move the kernel dependency to `tokenspeed-kernel`.
`tokenspeed` should depend only on `tokenspeed-kernel`.

dongjiyingdjy: Done. Moved `tilelang==0.1.9` out of `python/pyproject.toml` and into `tokenspeed-kernel/python/requirements/cuda-thirdparty.txt`, so the top-level `tokenspeed` package no longer directly declares this kernel dependency.

zhyncs:
@codex review

PR #31, opened 2026-05-08
perf(K2.5): fix slow tokenize and add warmup for agentic bench

syuoni:
## Summary
  - **Tokenize chat completions via `apply_chat_template(tokenize=False)` + `tokenizer.encode(text)`** in `serving_chat.py`, replacing the slow `tokenize=True` form. K2.5 ships no Rust fast tokenizer, so `tokenize=True` runs `PythonBackend.tokenize` + `convert_tokens_to_ids` (~360 ms for a ~70k-token multi-turn history); the two-step form is ~28 ms.
  - **Make the request-conversion thread pool size env-tunable** via `TOKENSPEED_REQUEST_CONVERSION_WORKERS` (default 8). Old value was hardcoded; the new env var is wired through `tokenspeed/runtime/utils/env.py` and consumed in `serving_base.py`.
  - **Agentic bench-script polish** for `test/agentic_benchmark/{tokenspeed,trtllm}/agentic_bench.sh`:
    - Added a small per-config warmup (`--number 2 --parallel 2 --dataset-offset 68`) before the measured sweep.


zhyncs (approved):

zhyncs:
@slin1237 @key4ng https://github.com/lightseekorg/smg/tree/main/crates/tokenizer/src/cache We can use this one with SMG.

PR #51, opened 2026-05-10
feat(deepseek-v4): add scheduler-managed sliding-window cache groups

SimonCqk:
## Summary

  - Introduce DeepSeek V4 sliding-window cache groups for short-lived cache surfaces such as SWA, compressor state, and indexer state, while keeping compressed KV cache as full-history storage.
  - Use compact sliding-window block tables with per-request base logical-page offsets as the end-to-end runtime contract, so released front pages can be compacted and reused instead of preserving full-context absolute table width.
  - Optimize the V4 cache memory layout by separating cache families with different lifetimes and page granularities, reducing wasted memory for sliding-window state and enabling earlier page reuse.
  - Improve forward-path throughput by reducing sliding-window metadata width, metadata movement, CUDA graph replay refresh work, and unnecessary compact-to-absolute table expansion.
  - Prefix cache support is intentionally out of scope for this PR and will be added in the next PR as a separate group-aware prefix-cache change.

## Test Plan

  - GSM8K regression, limit 100: score `1`; no regression observed.
  - Added/updated coverage for compact sliding cache groups, base logical-page offsets, CUDA graph compact-table replay, and V4 cache allocation behavior.
  - Static validation:
    - `python3 -m py_compile ...`
    - `python3 -m ruff check ...`
    - `python3 -m ruff format --check ...`
    - `git diff --check`

### DIFF
--- a/python/tokenspeed/runtime/configs/paged_cache_spec.py
+++ b/python/tokenspeed/runtime/configs/paged_cache_spec.py
@@ -1,0 +1,108 @@
+# Copyright (c) 2026 LightSeek Foundation
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Dict, Literal, Optional, Sequence
+
+Retention = Literal["full_history", "sliding_window"]
+
+
+@dataclass(frozen=True)
+class PagedCacheGroupSpec:
+    group_id: str
+    retention: Retention
+    rows_per_page: int
+    entry_stride_tokens: int
+    sliding_window_tokens: Optional[int]
+
+
+_PAGED_CACHE_GROUP_DUMMY_PAGES = 1
+
+
+def _ceil_div(numer: int, denom: int) -> int:
+    if denom <= 0:
+        raise ValueError(f"_ceil_div denom must be > 0, got {denom}")
+    return (numer + denom - 1) // denom
+
+
+def compute_paged_cache_group_page_counts(
+    specs: Sequence[PagedCacheGroupSpec],
+    *,
+    max_live_requests: int,
+    max_scheduled_tokens: int,
+    max_total_tokens: int,
+    max_context_len: int,
+    safety_margin: int = 0,
+) -> Dict[str, int]:
+    if max_live_requests < 0:
+        raise ValueError(f"max_live_requests must be >= 0, got {max_live_requests}")
+    if max_scheduled_tokens < 0:
+        raise ValueError(
+            f"max_scheduled_tokens must be >= 0, got {max_scheduled_tokens}"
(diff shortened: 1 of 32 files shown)


LorrinWWW:
Do we need mixed batch in this PR? It looks like the implementation isn't complete yet, so I'd lean toward splitting it into a separate PR.

dongjiyingdjy:
> Do we need mixed batch in this PR? It looks like the implementation isn't complete yet, so I'd lean toward splitting it into a separate PR.

I will add mixed batch in the following PR. Thx!

python/tokenspeed/runtime/configs/paged_cache_spec.py
```diff
@@ -0,0 +1,108 @@
...
+    retention: Retention
+    rows_per_page: int
+    entry_stride_tokens: int
+    sliding_window_tokens: Optional[int]
+
+
+_PAGED_CACHE_GROUP_DUMMY_PAGES = 1
+
+
+def _ceil_div(numer: int, denom: int) -> int:
```
dongjiyingdjy: Could we use ceil_div in `common.py` instead of a new fuction?

SimonCqk:
> Do we need mixed batch in this PR? It looks like the implementation isn't complete yet, so I'd lean toward splitting it into a separate PR.

Thanks for reviewing, `mixed batch` will be submitted in a separate PR, I trimmed it out from the current one.

dongjiyingdjy:
Naming suggestion for the V4 paged-cache group ids.

The current ids work mechanically, but some of them are hard to read once they cross the scheduler/runtime boundary:

- `compressed.4` / `compressed.128` are too generic and not V4-namespaced.
- `4` / `128` expose only the compression ratio, while the model-level semantics are really `c4a` / `c128a`.
- `v4.indexer_state.4` does not make it obvious that this is the rolling compressor state used by the C4 indexer path, not the persisted indexer KV cache itself.

Could we rename these before landing? Suggested mapping:

| Current | Suggested |
|---|---|
| `v4.swa` | `v4.swa_kv` |
| `v4.compressor_state.4` | `v4.c4a.compressor_state` |
| `v4.compressor_state.128` | `v4.c128a.compressor_state` |
| `compressed.4` | `v4.c4a.compressed_kv` |
| `compressed.128` | `v4.c128a.compressed_kv` |
| `v4.indexer_state.4` | `v4.c4a.indexer_compressor_state` |

This is closer to the DeepSeek V4 terminology used elsewhere (`c4a`, `c128a`, SWA KV, compressed/main KV, compressor state, indexer cache), and avoids leaking a generic `compressed.*` namespace into scheduler metadata. The change should be mostly mechanical: the spec builder/constants, runtime lookup prefixes, scheduler/tests, and the scheduler comment that currently mentions `compressed.4` / `v4.swa`.

SimonCqk:
> Naming suggestion for the V4 paged-cache group ids.
> 
> The current ids work mechanically, but some of them are hard to read once they cross the scheduler/runtime boundary:
> 
> * `compressed.4` / `compressed.128` are too generic and not V4-namespaced.
> * `4` / `128` expose only the compression ratio, while the model-level semantics are really `c4a` / `c128a`.
> * `v4.indexer_state.4` does not make it obvious that this is the rolling compressor state used by the C4 indexer path, not the persisted indexer KV cache itself.
> 
> Could we rename these before landing? Suggested mapping:
> 
> Current	Suggested
> `v4.swa`	`v4.swa_kv`
> `v4.compressor_state.4`	`v4.c4a.compressor_state`
> `v4.compressor_state.128`	`v4.c128a.compressor_state`
> `compressed.4`	`v4.c4a.compressed_kv`
> `compressed.128`	`v4.c128a.compressed_kv`
> `v4.indexer_state.4`	`v4.c4a.indexer_compressor_state`
> This is closer to the DeepSeek V4 terminology used elsewhere (`c4a`, `c128a`, SWA KV, compressed/main KV, compressor state, indexer cache), and avoids leaking a generic `compressed.*` namespace into scheduler metadata. The change should be mostly mechanical: the spec builder/constants, runtime lookup prefixes, scheduler/tests, and the scheduler comment that currently mentions `compressed.4` / `v4.swa`.

done with naming improvement

dongjiyingdjy (approved):

python/tokenspeed/runtime/configs/paged_cache_spec.py
SimonCqk: done

PR #76, opened 2026-05-11
fix: support disabling KV prefix cache while keeping Mamba state cache functional

tuanzhangCS:
## Summary
This PR introduces the ability to disable KV prefix caching (`disable_prefix_cache`) .

### Changes

1. **KVPrefixCache support for disabled prefix caching**
   - Add `disable_prefix_cache` flag to `KVPrefixCache`
   - When disabled, `Match()` returns root node directly without actual prefix matching
   - Introduce `RawMatch()` for internal prefix cache operations that still need raw matching logic
   - Skip creating `HybridPrefixCache` when prefix cache is disabled

2. **Fix Mamba state cache independence from KV prefix cache**
   - Decouple `enable_mamba` from `enable_prefix_caching` on Python side
   - Always create `MambaChunkAllocator` and `HybridPrefixCache` when model has Mamba, regardless of prefix cache setting
   - Use `RawMatch()` in retract/finish paths to avoid bypassing Mamba state tracking
   - Remove deprecated `num_mamba_slots` config, unify on `mamba_pool_total_chunks`

3. **Tests**
   - Add `DisablePrefixCacheTestSuite` to verify no device prefix reuse and no prefetch when prefix cache is off
   - Add `DisablePrefixCacheMambaRetractTest` to verify Mamba state is preserved through retract/recover cycles with prefix cache disabled

## Test Plan
- [x] Qwen3.5 with prefix cache enable
- [x] Qwen3.5 with prefix cache disable


XucSh:
The overall direction looks reasonable to me: `disable_prefix_cache` should disable cross-request prefix reuse, while still allowing internal state recovery paths to find their own radix-tree nodes. The current `Match()` vs `RawMatch()` split captures that intent, and it avoids scattering `if disable_prefix_cache` checks across the scheduler.

That said, I think the API could be made cleaner and safer. `RawMatch()` is a bit too broad: it means “ignore the prefix-reuse policy and match the real tree”, but that is only valid for internal lifecycle/state-recovery paths such as finish, retract, and recover. A future caller could easily use `RawMatch()` in a normal scheduling path and accidentally re-enable prefix reuse.

Consider making the intent explicit, for example:

```cpp
enum class MatchIntent {
    PrefixReuse,
    StateRecovery,
};

MatchResult KVPrefixCache::Match(tokens, MatchIntent intent);
```

Then disable_prefix_cache can apply only to the PrefixReuse path, while StateRecovery still walks the real radix tree. This would make it much clearer that disabling prefix cache disables cross-request reuse for both KV and Mamba state, but does not disable per-request Mamba state tracking needed for retract/recover/decode correctness.

I also think the naming around HybridPrefixCache is becoming slightly misleading after this change. Even when prefix reuse is disabled, this object still exists because it also tracks Mamba state/checkpoints. A short comment or a future rename/split would make the ownership clearer: prefix reuse policy and Mamba state tracking are related implementation-wise, but they are not the same feature.

tuanzhangCS:
> Please add an end-to-end unit test for this.

OK

python/tokenspeed/runtime/engine/event_loop.py
XucSh, line 1: During retract, the new code moves the request-local Mamba checkpoint/working slot into the radix-tree terminal node. This is not releasing the Mamba resource; it transfers ownership from the request to the tree and tracks it for Mamba eviction. However, the retracted request still carries its LocalMambaAllocator, and the recover path only moves that local allocator back into Decoding; it never restores or COWs the tree-owned Mamba slot back into the request. If the intent is same-request recovery, keeping the Mamba slot local seems simpler and safer. If the intent is tree-owned recovery, recover must explicitly restore/copy it from the matched node, and the eviction semantics for retracted requests need to be defined.

tuanzhangCS: I believe that when retracting a request, the Mamba state still needs to remain attached to the prefix tree, because its Mamba cache needs to be preemptible. For this scenario, I have made some refinements based on your suggestions.

PR #81, opened 2026-05-11
perf(qwen3): cut H100 decode kernel time -8% with fused stride-aware kernels

qywu:
## Summary
- **Decode-kernel time on H100 falls 972 → 895 ms (-8 %)** on the eager profile (Qwen3-8B, 4 × 512 in × 128 out, conc 4).
- **End-to-end serving rises 18,183 → 20,078 tok/s (+10.4 %)** on the standard random 4096/256, 200-prompt, conc-32 bench (cuda graphs ON, `--disable-kvstore`).
- Four kernel-/runtime-level changes: qk-rmsnorm fusion, FA3 scheduler-metadata cache, fused KV scatter, stride-aware kernels.
- Full bench data + per-step bucket diffs in the comment below.

## Motivation
Profiling the H100 decode hot path identified four mostly-unrelated overhead sources:

1. **QK-norm** ran as two separate flashinfer RMSNorm calls (`q_norm`, `k_norm`) with reshape round-trips, inducing `.contiguous()` copies per layer per step.
2. **FA3 `prepare_varlen_num_blocks`** kernel fired inside every layer's attention call (4,680 launches per decode bench), dominating the "Attention prepare" bucket.
3. **KV-cache writes** went through `tensor[loc] = src` index_put_ on each of k and v separately (~32 ms / 10 k launches in the trace).
4. **Stride mismatches**: the new kernels above forced inputs to `.contiguous()`, producing 14,040 hidden direct-copy kernels per decode bench (~57 ms total).

## Changes

| Change | File(s) | Bucket Δ vs baseline |
|---|---|---:|
| Fused QK-norm (single triton launch, stride-aware, fresh contig outputs) | `tokenspeed-kernel/.../ops/layernorm/triton.py` (+`qk_rmsnorm`); `python/.../models/qwen3.py` | RMSNorm-plain: 14.06 → 9.59 ms |
| Pre-compute FA3 scheduler metadata once per scheduler step | `python/.../layers/attention/backends/mha.py`; `tokenspeed-kernel/.../ops/attention/{__init__.py, flash_attn/__init__.py}` | Attention-prepare: 14.25 → 0.60 ms (4,680 → 200 launches) |
| Fused KV-cache scatter triton kernel | `tokenspeed-kernel/.../ops/kvcache/triton.py` (+`store_kv_cache`); `python/.../layers/attention/kv_cache/mha.py` | KV-write: ~32 ms index_put_ → ~6 ms fused |
| Stride-aware kernels (no `.contiguous()` on hot path) | both new triton kernels above | Elementwise/copy: 73.26 → 6.77 ms |

### Eager decode kernel time (4 × 512 in × 128 out, conc 4)

| Variant | Total | Δ vs baseline |
|---|---:|---:|
| baseline | 972.21 ms | — |
| step 1 (fused QK norm) | 970.27 ms | −2 ms |
| step 2 (cached scheduler metadata) | 956.82 ms | −15 ms |
| step 3 (fused KV scatter) | 925.72 ms | −47 ms |
| **step 4 (stride-aware kernels)** | **894.75 ms** | **−77 ms (−8 %)** |

### End-to-end serving (cuda graphs ON, 4096/256, 200 prompts, conc 32)

| Config | Total tok/s | TPOT med | TTFT med | E2E mean |
|---|---:|---:|---:|---:|
| baseline | 18,183 | 19.93 ms | 2,040 ms | 7,183 ms |
| this PR (steps 1+2+3+4) | 18,871 (+3.8 %) | 19.20 ms | 2,051 ms | 6,972 ms |
| this PR + `--disable-kvstore` | **20,078 (+10.4 %)** | **19.16 ms** | **1,758 ms** | **6,584 ms** |

The KVStore overhead (`_kv_transfer_all_layer_cs32_kernel` running ~28 ms × 16 calls per prefill batch) dominates the e2e gain on this random-prompt benchmark. KVStore stays useful for shared-prefix workloads — see "Caveats" for the demand-driven follow-up.

## Caveats / not addressed in this PR
- All measurements are on the SM 9.0 fallback path of a B200-targeted build (built with `TOKENSPEED_CUDA_ARCH=90`). Step 1's triton kernel may not be the right choice on Blackwell where a tcgen05/CuteDSL fused variant likely exists; would need a sanity check on B200.
- Step 2 (FA3 scheduler-metadata cache) is gated to eager mode (`use_cuda_graph=False`) to keep cuda-graph capture stable. Making it graph-safe would require a fixed-shape pre-allocated metadata buffer; modest additional e2e win expected because graphs already amortize launch overhead.
- After KVStore-off, residual TTFT is dominated by the triton `qk_rmsnorm` being launch-bound on prefill batches. A warp-per-head rewrite is the next high-value target if prefill TTFT matters for the workload.
- KVStore's transfer kernel fires speculatively every prefill batch. A demand-driven version (only fire when GPU pool eviction is imminent) would let it stay enabled by default in production without taxing prefix-free workloads.

## Test plan
- [x] Correctness: standalone unit tests for `qk_rmsnorm` and `store_kv_cache` against a torch reference produce 0 max-abs-diff (q/k bf16 numerical diff < 5e-3 due to triton vs torch reduction order).
- [x] Server boots cleanly with the changes (`tokenspeed serve --model Qwen/Qwen3-8B`).
- [x] Decode profile (eager): all four bucket targets met; total decode kernel time -8 % vs baseline.
- [x] E2e serving bench (cuda graphs ON, `--disable-kvstore`): +10.4 % throughput vs baseline.
- [x] `pre-commit run --all-files` clean (isort, black-jupyter, etc.).
- [ ] B200 sanity check (no access during this work) — should run before merging if possible.
- [ ] Confirm no regression in TP>1 / spec-decode / draft-extend paths (only normal-decode path was touched).

## Reproducing
On H100, build with `TOKENSPEED_CUDA_ARCH=90` (this build's prebuilt fatbins are sm_100a; the editable install will recompile for the requested arch). Then:

```bash
tokenspeed serve --model Qwen/Qwen3-8B --port 30000 \
  --max-model-len 8192 --disable-kvstore
```

End-to-end serving numbers measured with an OpenAI-compatible bench client (random-token dataset, 4096 in / 256 out, 200 prompts, max-concurrency 32, request-rate inf).


qywu:
# Qwen3-8B · H100 · Decode-path optimization

Branch: `qywu/tokenspeed/perf-h100-gap`. Target: reduce decode kernel time
on the H100 SM 9.0 fallback path (this build's prebuilt fatbins target
SM 100a; we rebuild for SM 90 via `TOKENSPEED_CUDA_ARCH=90`).

## Setup

- 1× H100 80 GB, TP=1, bf16, max-model-len 8192.
- Tokenspeed launched with `--enforce-eager` for kernel profiles, with
  cuda graphs ON for end-to-end serving numbers.
- Profiles via `/start_profile` (torch profiler, GPU+CPU activities).
- End-to-end driver is an OpenAI-compatible bench client (random-token
  dataset, configurable input/output lengths and concurrency).

## Headline numbers

| Workload | Baseline kernel time | After this branch | Δ |
|---|---:|---:|---:|
| Prefill (1 × 4096 in × 1 out) | 15.26 ms | 14.29 ms | **−1.0 ms (−6 %)** |
| Decode (4 × 512 in × 128 out, conc 4) | 972.21 ms | 894.75 ms | **−77.5 ms (−8 %)** |

End-to-end serving (random 4096/256, 200 prompts, conc 32, cuda graphs
ON, `--disable-kvstore`):

| Variant | Total tok/s | TPOT med | TTFT med | E2E mean |
|---|---:|---:|---:|---:|
| baseline (`main`) | 18,183 | 19.93 ms | 2,040 ms | 7,183 ms |
| this branch | **20,078 (+10.4 %)** | 19.16 ms | 1,758 ms | 6,584 ms |

## What was costing time on the decode hot path

Per-bucket breakdown of the baseline decode profile (4 × 512 in × 128
out, conc 4):

| Bucket | Baseline (ms) | n_calls | Cause |
|---|---:|---:|---|
| GEMM (cuBLAS) | 726.82 | 18,850 | Hopper kernels — already optimal |
| Elementwise / copy / index | **73.26** | **32,346** | Many small unfused glue ops between layer steps |
| GEMM (CUTLASS / splitK) | 80.95 | 18,576 | Attention compute — already optimal |
| RMSNorm (fused add) | 27.28 | 9,360 | Layer-norm + residual — already fused |
| SiLU/SwiGLU | 15.92 | 4,680 | Already fused |
| **Attention prepare** (`flash::prepare_varlen_num_blocks`) | **14.25** | **4,680** | Re-runs per layer per step; can be batched/cached |
| **RMSNorm (plain)** — q_norm, k_norm | **14.06** | **9,490** | Two separate launches per layer + reshape round-trips |
| RoPE | 9.11 | 4,680 | Already fused |
| Other | 10.57 | 1,985 | Misc |
| **TOTAL** | **972.21** | | |

Three outliers:
1. **Plain RMSNorm** at 14 ms / 9 k launches — Qwen3's `_apply_qk_norm`
   ran two sequential RMSNorm calls (`q_norm`, `k_norm`) through reshape
   round-trips; each round-trip plus per-head RMSNorm produced a
   `.contiguous()` copy.
2. **`prepare_varlen_num_blocks`** at 14 ms / 4,680 launches — the FA3
   kernel's internal scheduling pre-compute fired once per layer per
   decode step (36 layers × ~130 steps).
3. **Elementwise / copy / index** at 73 ms / 32 k launches — dominated
   by `aten::contiguous` (14 k calls) and `aten::index_put_` for KV
   cache writes (10 k calls). Most of these were avoidable.

## Optimization plan and results

### 1. Fuse QK-norm into a single triton launch

`_apply_qk_norm` (`qwen3.py:178-187`) ran two sequential RMSNorm calls
plus reshape round-trips. New kernel
`tokenspeed_kernel.ops.layernorm.triton.qk_rmsnorm` does both q and k in
one launch.

- **Bucket effect:** RMSNorm-plain 14.06 → 9.59 ms (−4.47 ms, −32 %),
  call count 9,490 → 4,810 (one launch per layer per step).
- **Risk:** none observed; bf16 numerical diff vs torch reference
  < 5e-3 (different reduction order).

### 2. Pre-compute FA3 scheduler metadata once per scheduler step

`MHAAttnBackend.init_forward_metadata` now calls
`flash_attn_interface.get_scheduler_metadata` once per scheduler step
for the normal-decode path and threads the result through
`mha_decode_with_kvcache` and the FA3 binding. The FA3 kernel skips its
internal `prepare_varlen_num_blocks_kernel` launch when the metadata is
provided.

- **Bucket effect:** Attention-prepare 14.25 → 0.60 ms (−13.6 ms, 4,680
  → 200 launches — one per scheduler step rather than one per layer per
  step).
- **Gating:** eager only. Cuda-graph capture keeps the prep inside the
  graph for shape stability.
- **Risk:** correctness preserved; the FA3 kernel produces identical
  output whether scheduling is precomputed or done inline.

### 3. Fused KV-cache scatter

`MHATokenToKVPool.set_kv_buffer` previously did two separate
`tensor[loc] = source` index_put_ assignments for k and v. New kernel
`tokenspeed_kernel.ops.kvcache.triton.store_kv_cache` does both in a
single triton scatter.

- **Bucket effect:** KV write ~32 ms (index_put_) → ~6 ms (fused).
- **Risk:** none; single kernel, simpler code path.

### 4. Stride-aware kernels — eliminate `.contiguous()` copies

The first cuts of the kernels in steps 1 and 3 forced `.contiguous()` on
their inputs. Because q, k, v come from `qkv.split(...)` they are
gap-strided (last-dim contiguous, leading-dim has the qkv-merged stride).
The forced contig produced 14,040 hidden direct-copy kernels per decode
bench (~57 ms). The fix: both kernels accept an explicit leading-axis
stride argument and read from non-contig inputs directly. `qk_rmsnorm`
writes a fresh contiguous output (so RoPE and attention see contig q, k);
`store_kv_cache` writes directly into the cache buffer.

- **Bucket effect:** Elementwise / copy / index 73.26 → 6.77 ms (−66.5
  ms, 32,346 → 4,269 launches). `aten::contiguous` cpu_op count on the
  hot path: 14,040 → 0.

### Cumulative

| Variant | Total decode | Δ vs baseline |
|---|---:|---:|
| baseline | 972.21 ms | — |
| step 1 only | 970.27 ms | −2 ms |
| steps 1+2 | 956.82 ms | −15 ms |
| steps 1+2+3 | 925.72 ms | −47 ms |
| **steps 1+2+3+4** | **894.75 ms** | **−77 ms (−8 %)** |

The `Other` bucket (+8.4 ms residual) and CUTLASS splitK (+0.5 ms) are
launch-overhead-bound and not chased here.

## End-to-end finding: KVStore prefill overhead

While verifying the e2e impact of steps 1–4 we found a separate large
overhead in the *prefill* path that was unrelated to any kernel-quality
issue: `_kv_transfer_all_layer_cs32_kernel` — KVStore's host-offload
transfer — fires 16× per prefill batch at ~28 ms each, ≈ 450 ms of D2H
bandwidth-bound work per batch. KVStore mirrors the GPU KV pool to host
memory for prefix caching across requests; it is useful when prefixes
are shared across requests, but on a random-prompt benchmark with no
shared prefix it's pure overhead.

Workaround for benchmarks: launch with `--disable-kvstore`. That is
already supported in `tokenspeed serve`. With it disabled, e2e
throughput rises from 18,871 to 20,078 tok/s (+6.4 % on top of the
kernel-level wins).

For production with shared-prefix workloads (RAG, system-prompted
chat) KVStore stays useful, but the transfer pattern (firing
speculatively every prefill batch even when no eviction is imminent) is
itself a worthwhile optimization opportunity. A demand-driven version —
fire only when GPU pool pressure is high or a previously-evicted
prefix is requested — would let KVStore stay enabled by default
without taxing prefix-free workloads. Tracked separately from this
branch.

## Files changed

| File | Purpose |
|---|---|
| `tokenspeed-kernel/.../ops/layernorm/triton.py` | New `qk_rmsnorm` (fused, stride-aware) |
| `tokenspeed-kernel/.../ops/kvcache/triton.py` | New `store_kv_cache` (fused k/v scatter, stride-aware) |
| `tokenspeed-kernel/.../ops/attention/__init__.py` | Wrapper threads optional `scheduler_metadata` |
| `tokenspeed-kernel/.../ops/attention/flash_attn/__init__.py` | FA3 binding honors `scheduler_metadata` |
| `python/tokenspeed/runtime/models/qwen3.py` | `_apply_qk_norm` calls `qk_rmsnorm` |
| `python/tokenspeed/runtime/layers/attention/backends/mha.py` | Pre-computes scheduler metadata once per step |
| `python/tokenspeed/runtime/layers/attention/kv_cache/mha.py` | Uses `store_kv_cache` |

## Caveats

- All measurements are on the SM 9.0 fallback path of a B200-targeted
  build. Step 1's triton kernel may not be the right choice on Blackwell,
  where a tcgen05/CuteDSL fused variant likely exists; sanity check on
  B200 recommended before assuming the win generalizes.
- Step 2 (FA3 scheduler-metadata cache) is gated to eager mode
  (`use_cuda_graph=False`) to keep cuda-graph capture stable. Making it
  graph-safe would need a fixed-shape pre-allocated metadata buffer —
  modest additional e2e win expected because graphs already amortize
  launch overhead.
- After the e2e wins above, residual TTFT is dominated by the triton
  `qk_rmsnorm` being launch-bound: per-call cost is competitive on
  decode (small batches) but ~3× behind a warp-per-head implementation
  on prefill batches. A warp-per-head rewrite is the next high-value
  target if prefill TTFT matters for the workload.
- Only the normal-decode path of the MHA backend was touched. Spec
  decode (target_verify, draft_extend) and other backends are
  unchanged.

python/tokenspeed/runtime/layers/attention/backends/mha.py
```diff
@@ -158,11 +170,47 @@ def init_forward_metadata(
...
+                from flash_attn_interface import (  # noqa: WPS433
+                    get_scheduler_metadata,
+                )
+
+                self._get_scheduler_metadata = get_scheduler_metadata
+            except ImportError:
+                self._get_scheduler_metadata = False
+        if not self._get_scheduler_metadata:
+            return None
+        return self._get_scheduler_metadata(
```
borontion: is it possible to create the scheduler metadata in the FA3 wrapper: `tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/flash_attn/__init__.py` instead of passing it through the mha backend?

PR #87, opened 2026-05-12
Fix KV cache event parent hash performance

### LINKED ISSUE #86: perf(dynamo): BuildBlockEventsForNode rehashes full prefix per insert (O(N^2 * page_size))
Follow-up from #8 ([review comment](https://github.com/lightseekorg/tokenspeed/pull/8#issuecomment-4426233722)).

`BuildBlockEventsForNode` in `tokenspeed-scheduler/csrc/resource/kv_prefix_cache/kv_prefix_cache.cpp` walks `RootToLeaf(target)` and re-hashes every ancestor page just to derive `parent_hash`. `Insert` calls it once per newly-stored node, so inserting N new pages re-hashes the full prefix N times — O(N^2 * page_size) FNV mixes.

The disabled path is well gated, so this only bites when KV events are enabled (`SetKvEventSink` set).

### Suggested fix
Cache `block_hash` on `TreeNode` so `BuildBlockEventsForNode` can read the parent's stored hash directly instead of recomputing the entire prefix chain. This drops the cost to O(N * page_size) per insert.

### Acceptance
- `BuildBlockEventsForNode` no longer iterates ancestors purely to derive `parent_hash`.
- Existing tests in `tokenspeed-scheduler/tests/cpp/test_kv_cache_events.cpp` and `test/runtime/test_kv_events.py` still pass.
- A microbench (or test) confirms insert-with-events scales linearly in number of new pages.

Aphoh:
Fixes #86.

## Summary
- Cache rolling KV block hashes on prefix-cache tree nodes and preserve them across radix splits.
- Build KV stored/removed events from cached node hashes instead of recomputing the full ancestor chain for each node.
- Publish `BlockStored` events when host-cache recovery materializes device pages, so downstream consumers receive parent blocks before children.
- Add coverage for split parent hashes, host-to-device recovery events, removals after recovery, and a small measured performance comparison.

## Performance
Added `KVPrefixCacheEventBenchTest.OptimizedInsertIsFasterThanLegacyAncestorRehashing`, which compares simulated legacy ancestor rehash work against the optimized event-enabled device insert path on a 512-page chain.

Local result:
```text
[kv-event-bench] pages=512 legacy_ms=13.1622 optimized_ms=0.275621 speedup=47.7547
```

## Testing
- `cmake --build /tmp/tokenspeed-scheduler-build --target tokenspeed_scheduler_tests -j 4`
- `/tmp/tokenspeed-scheduler-build/tokenspeed_scheduler_tests --gtest_filter='KVPrefixCacheEventTestSuite.*:KVPrefixCacheEventBenchTest.*:SchedulerKvCacheEventTestSuite.*'` (9 passed)
- `/tmp/tokenspeed-scheduler-build/tokenspeed_scheduler_tests` (135 passed)
- `uvx pre-commit run --all-files`


LorrinWWW (approved):
lgtm

PR #105, opened 2026-05-13
perf: optimize flashinfer sampling backend

syuoni:

### DIFF
--- a/python/tokenspeed/runtime/sampling/backends/flashinfer.py
+++ b/python/tokenspeed/runtime/sampling/backends/flashinfer.py
@@ -348,7 +348,7 @@ class FlashInferSamplingBackend(SamplingBackend):
                 uniform_samples=self._coins_buf[:bs, :n],
                 uniform_samples_for_final_sampling=self._final_coins_buf[:bs],
                 target_probs=target_probs,
-                draft_probs=torch.zeros_like(target_probs),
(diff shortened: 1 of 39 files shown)


python/tokenspeed/runtime/sampling/backends/flashinfer.py
```diff
@@ -324,25 +312,31 @@ def verify(
...
+                temperature=temperatures,
+                enable_pdl=pdl_enabled(),
             )
-            target_probs = top_k_top_p_renorm_torch(
-                target_probs,
-                torch.repeat_interleave(top_ks, n),
-                torch.repeat_interleave(top_ps, n),
+            target_probs = top_k_renorm_prob(target_probs, top_ks)
+            target_probs = top_p_renorm_prob(
+                target_probs, top_ps, is_deterministic=True
```
yubofredwang, line 339: Just curious, why does this have to be deterministic?

syuoni: Otherwise, the sampling results may diverge between TP ranks. Currently `top_k_renorm_prob` is still a source of non-deterministic, so we need `maybe_broadcast` to sync the sampling results.

In the future, we should introduce a fully deterministic implementation, so the broadcast op can be removed.

PR #121, opened 2026-05-13
fix: increase engine startup timeout and use /readiness for server readiness query

syuoni:


test/runtime/models/test_mm25_perf.py
```diff
@@ -550,6 +550,7 @@ def measure(port):
         )
 
     # xgrammar poem: stream decode TPS + JSON validity.
+    @unittest.skip("MIN_XGRAMMAR_GEN_TOKENS floor needs recalibration")
```
syuoni, line 553: @LorrinWWW can you take a look at this failure? I can repro it on ToT main.

LorrinWWW: we can skip it for now.

PR #126, opened 2026-05-13
perf(K2.5): Optimize lm_head

yweng0828:
## Summary

Optimize `lm_head` for low_latency.

- Before: 94.072 us
image

aime25 accuracy
```
+-----------------+-----------+----------+----------+-------+---------+---------+
| Model           | Dataset   | Metric   | Subset   |   Num |   Score | Cat.0   |
+=================+===========+==========+==========+=======+=========+=========+
| Kimi-K2.5-NVFP4 | aime25    | mean_acc | default  |    30 |  0.9667 | default |
+-----------------+-----------+----------+----------+-------+---------+---------+ 
```

- After: 87.008 us (1.08x)
image

aime25 accuracy
```
+-----------------+-----------+----------+----------+-------+---------+---------+
| Model           | Dataset   | Metric   | Subset   |   Num |   Score | Cat.0   |
+=================+===========+==========+==========+=======+=========+=========+
| Kimi-K2.5-NVFP4 | aime25    | mean_acc | default  |    30 |       1 | default |
+-----------------+-----------+----------+----------+-------+---------+---------+ 
```

### DIFF
--- a/python/tokenspeed/runtime/layers/logits_processor.py
+++ b/python/tokenspeed/runtime/layers/logits_processor.py
@@ -122,6 +122,51 @@ class LogitsMetadata:
         )
 
 
+_FUSED_LM_HEAD_GEMM = None
+
+
+def _get_fused_lm_head_gemm():
+    """Lazily import the fused lm_head GEMM kernel.
+
+    The kernel is only present when tokenspeed-kernel was built with a
+    compatible nvcc. Cache a sentinel when unavailable so we fall back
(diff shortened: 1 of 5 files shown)


LorrinWWW:
is there a reason to gate for kimi only?

yweng0828:
> is there a reason to gate for kimi only?

Hi, because we only verify the performance on kimi. It would be preferable if the selection could be performed heuristically, focusing solely on the GEMM shape, independent of the specific model.

PR #142, opened 2026-05-14
perf(K2.5): optimize small kernels in EAGEL3 drafter loop

syuoni:
## Summary

  Per-call overhead reductions in the EAGLE3 drafter loop and surrounding metadata prep.

  ### Changes

  - **`compute_out_cache_loc` uniform variant** — added `compute_out_cache_loc_uniform`
    for the drafter's multi-step decode where every request has `input_length=1`. Skips
    the per-call `torch.cumsum` + `torch.full` host-side work and the kernel's GMEM
    reads of `input_lengths_ptr`/`cumsum_lengths_ptr` (Triton specializes on
    None-pointer at JIT time).
  - **`req_pool_indices_buf` is now int64** — eliminates ~12 implicit int32→int64
    `unrolled_elementwise` cast kernels (~1.6 µs each) along the per-iteration
    metadata prep path. Pairs with switching the `valid_cache_lengths[idx]` fancy
    index to `valid_cache_lengths.index_select(0, idx)` so int32 indices are still
    accepted natively where they remain (e.g. in `index_add_`).
  - **Persistent drafter buffers** — `draft_seq_lens_buf`, `draft_out_cache_loc_buf`,
    `draft_input_lengths_buf`, and `last_index_offsets_buf` (= `arange(max_bs) *
    spec_num_tokens - 1`) are hoisted to `Eagle.__init__` to avoid per-call alloc +
    init. `last_index_offsets` is now plumbed via `ForwardContext` to
    `LogitsProcessor` for the padded-static-len last-token-select path.
  - **`Eagle.draft()` cleanup** — fused `cache_start + 1` into
    `torch.add(..., out=draft_seq_lens)`; replaced the post-`draft()` `torch.cat`
    with direct writes into a shared `next_tokens[bs, spec_num_steps+1]` buffer;
    skip the last-iter `positions.add_(1)` / `draft_seq_lens.add_(1)` since they're
    not consumed; removed the dead `logits.shape[0] != bs` fallback branch.
- **Persistent drafter buffers** — `draft_seq_lens_buf`, `draft_out_cache_loc_buf`,
    `draft_input_lengths_buf`, and `last_index_offsets_buf` are hoisted to
    `Eagle.__init__` to avoid per-call alloc + init. The last one
    (`= torch.arange(max_bs) * spec_num_tokens - 1`, int64) replaces **two**
    per-call `torch.arange(bs, ...) * spec_num_tokens` patterns — one in the
    drafter's last-verified-id selection, one in `LogitsProcessor`'s padded-static-len
    last-token-select. The precomputed buffer is sliced to `[:bs]` and plumbed via
    `ForwardContext.last_index_offsets` so `LogitsProcessor` can skip the
    `arange + mul + sub` triplet. Net: pre-step-0 last-token-select drops from
    6 kernels (arange + mul + sub + 2 gathers) to 3 (1 add + 2 gathers).

### DIFF
--- a/docs/serving/deepseek-v4.md
+++ b/docs/serving/deepseek-v4.md
@@ -68,7 +68,7 @@ GSM8K 5-shot, 50 samples is the standard quick-validation harness for V4:
 ```bash
 HF_DATASETS_TRUST_REMOTE_CODE=1 lm_eval run \
     --model local-completions \
-    --model_args "model=deepseek-ai/DeepSeek-V4-Flash,base_url=http://127.0.0.1:30100/v1/completions,tokenized_requests=False,tokenizer_backend=None,num_concurrent=4,max_retries=1,timeout=600,max_gen_toks=256" \
+    --model_args "model=deepseek-ai/DeepSeek-V4-Flash,base_url=http://127.0.0.1:8000/v1/completions,tokenized_requests=False,tokenizer_backend=None,num_concurrent=4,max_retries=1,timeout=600,max_gen_toks=256" \
     --tasks gsm8k --num_fewshot 5 --limit 50 --batch_size 1 \
     --gen_kwargs temperature=0
 ```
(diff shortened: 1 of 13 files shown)


LorrinWWW (approved):
lgtm

PR #144, opened 2026-05-14
deps: switch SMG gateway packages to `tokenspeed-smg`

zhyncs:
## Summary

- Pull the SMG gateway into tokenspeed's hard runtime dependencies under the tokenspeed-namespaced names now published to PyPI:
  - `tokenspeed-smg==1.4.1.post20260514`
  - `tokenspeed-smg-grpc-proto==0.4.7.post20260514`
  - `tokenspeed-smg-grpc-servicer==0.5.2.post20260514`
- Drop the standalone `pip install smg smg-grpc-* --extra-index-url https://lightseek.org/whl/...` step from `install_deps.sh` and `install_deps_rocm.sh`. pip now resolves the same set straight from PyPI when it installs tokenspeed (`pip install -e ./python[...]`).
- The runtime safety net in `_check_serve_extra_installed` still verifies that `smg` and `smg_grpc_servicer.tokenspeed.server` are importable — the upstream Python module names are unchanged — but its install hint is refreshed to point at the new PyPI names.

## Test plan

- [x] `pre-commit run --files python/pyproject.toml python/tokenspeed/cli/serve_smg.py test/ci_system/install_deps.sh test/ci_system/install_deps_rocm.sh`
- [x] `pytest test/cli/test_serve_smg_unit.py -v` — 18 passed locally; the install-hint rewrite doesn't touch any behavior the suite exercises.
- [ ] `ut-runtime-1gpu` per-commit task will re-run `install_deps.sh` end-to-end and execute the CLI + runtime suites, exercising the new dependency resolution path.


zhyncs:
### Why repackage instead of consuming upstream directly

Same playbook as the other `tokenspeed-` third-party recipes (`tokenspeed-deepgemm`, `tokenspeed-fa3`, `tokenspeed-flashmla`, `tokenspeed-mla`, …): we don't fork `smg`, we **repackage** it under the `tokenspeed-` PyPI namespace pinned to a specific upstream commit.

Why:

- **Move fast.** Most of these projects don't tag releases frequently enough for our needs. Pinning by commit + post-versioning (`<upstream_version>.post<YYYYMMDD>`) lets us pull in a fix the same day it lands upstream, without waiting on someone else's release cadence.
- **Stay close to upstream.** The patch we apply at build time is intentionally minimal — just the `pyproject.toml` `name = …` rewrite plus, for proto, the runtime `importlib.metadata.version(...)` lookup. Python module names, Rust crate names, console scripts, every import path are identical to upstream. `import smg` and `smg launch ...` keep working unchanged. The intent is for our `tokenspeed-smg` to be a drop-in for whichever upstream `smg` commit we pinned — and for the codebase to track `lightseekorg/smg` `main` (or its latest stable tag) as closely as possible.
- **Reproducible installs.** A `tokenspeed` install pulling `tokenspeed-smg==1.4.1.post20260514` is bit-for-bit the same gateway every time, regardless of what upstream's `main` looks like next week. Bumping is a one-PR motion: regenerate the patch dir against the new SHA, bump three pins here, done.
- **Single install surface.** Before this PR, `ts serve` needed a separate `pip install smg smg-grpc-* --extra-index-url https://lightseek.org/whl/…` step (also baked into `test/ci_system/install_deps.sh`). Moving those into `pyproject.toml` means `pip install tokenspeed` resolves the entire serve stack from PyPI in one shot — no out-of-band index, no two-step instructions.

Bumping flow when we want to follow upstream `main`:

1. Pick the upstream commit, push it to the protected `lightseekorg/smg/tokenspeed-smg` branch (or just record the full SHA).
2. Add a `tokenspeed-smg/patches/<short_sha>/` directory in the repackage repo.
3. Fire the three publish workflows (`tokenspeed-smg-grpc-proto` → `tokenspeed-smg-grpc-servicer` with pin → `tokenspeed-smg`) with `publish_to_pypi=true`.
4. Open a follow-up PR here bumping the three pins in `python/pyproject.toml`.

PR #154, opened 2026-05-15
[AMD]Support a-fp8-w-mxfp4 gpt-oss-120b model

knwng:
- Support https://huggingface.co/amd/gpt-oss-120b-w-mxfp4-a-fp8 model
- Make `score_threshold` support different thresholds on different runners since they're using different weights.


knwng:
@lightseek-bot @borontion can you take a look?

PR #159, opened 2026-05-15
Perf[Qwen3.5]: eliminate Mamba intermediate state memcpy in MTP target-verify

tuanzhangCS:
## Summary

- Replace the intermediate state buffer + post-verify scatter approach with inline output-state-indices: during MTP target-verify, conv1d and SSM kernels now write states directly to per-draft-token slots, eliminating the large intermediate buffers (intermediate_ssm_state_cache, intermediate_conv_window_cache) and the fused_mamba_state_scatter_with_mask memcpy step after verification.
- Introduce current_input_indices bookkeeping in SimpleMambaPool to track which cache slot each request should read from at the start of target-verify, with proper reset on extend/retract to stay in sync with the C++ scheduler.
- Reduce GPU memory footprint by removing per-layer (pool_size, draft_tokens, state_shape) intermediate buffers, replaced by a flat draft-slot region appended to the existing conv/SSM state tensors.

## Test Plan
perf:
input=1024, output=4096, concurrency=64, num-requests=200

|   | before | optimize |
---|---|---
| output tokens/s | 4143.97  | 4301.36 | 
| nsys | image |  image |

accuracy：
gsm8k
image

### DIFF
--- a/python/tokenspeed/runtime/engine/event_loop.py
+++ b/python/tokenspeed/runtime/engine/event_loop.py
@@ -640,6 +640,16 @@ class EventLoop:
         if host_exec is not None:
             self.model_executor.execution_stream.wait_stream(host_exec.write_stream)
 
+    def _flush_mamba_retract_states(self, forward_op) -> None:
+        """Copy draft->working mamba states when retract occurred (no forward scheduled)."""
+        if forward_op is not None:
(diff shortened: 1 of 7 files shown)


zhyncs:
@codex review

PR #164, opened 2026-05-16
refactor: forward mode consolidation

syuoni:
## Summary

  This is a prerequisite for a large-scale enabledment of mixed prefill&decode batching.

  Collapse `ForwardMode.TARGET_VERIFY` and `ForwardMode.DRAFT_EXTEND` into `DECODE`. Backends now dispatch on `config.is_draft` + `spec_num_tokens = num_tokens // bs` instead of separate enum values — `TARGET_VERIFY` is `not is_draft and spec_num_tokens > 1`, `DRAFT_EXTEND` is `is_draft and spec_num_tokens > 1`. The enum keeps only `EXTEND`, `DECODE`, `MIXED`, `IDLE`.

  ### Why

  The old enum split a single concept (decode with variable q_len) across three values and required the controller to know about drafter state to pick the right one. The new split is local to each backend, and drafter vs. target is already a static property of the backend's config.

  ### Changes

  - `ForwardMode`: remove `TARGET_VERIFY` / `DRAFT_EXTEND`; drop `has_drafter` from `from_num_extends`; add `is_extend_or_mixed()` helper.
  - `BaseAttnConfig.is_draft: bool` (defaults to `False`); `AttentionBackend.__init__` reads it.
  - `init_forward_metadata_replay_cuda_graph` signature: `num_tokens` becomes the 2nd positional, so backends can compute `spec_num_tokens` without mode-specific signals.
  - `CudaGraphWrapper`: capture/replay always uses `ForwardMode.DECODE`; passes `padded_bs * max_tokens_per_req` as `num_tokens` for target and drafter alike.
  - All 10 attention backends migrated (flash_attention, flashmla, trtllm, trtllm_mla, mha, triton, tokenspeed_mla, deepseek_v4, hybrid_linear_attn, base) using consistent local-predicate renames. `trtllm_mla` / `tokenspeed_mla`: drop `forward_extend` — old verify/draft-extend paths now route through `forward_decode` based on `spec_num_tokens > 1`.
  - `Eagle` drafter, `LogitsProcessor`, `EventLoop`, `ModelExecutor`: drop mode-specific branches in favor of `is_decode()` / `is_extend_or_mixed()`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,3 +1,4 @@
 # CLAUDE.md
 
 @AGENTS.md
+@AGENTS.local.md
--- a/python/tokenspeed/runtime/engine/event_loop.py
+++ b/python/tokenspeed/runtime/engine/event_loop.py
@@ -792,7 +792,6 @@ class EventLoop:
         forward_mode = ForwardMode.from_num_extends(
             forward_op.num_extends(),
             len(forward_op.request_ids),
-            has_drafter=self.server_args.speculative_algorithm is not None,
         )
         self.request_handler._profile_batch_predicate(forward_mode)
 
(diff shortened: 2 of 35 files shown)


zhyncs:
NB

PR #165, opened 2026-05-16
perf: add gluon fp16 prefill kernel

borontion:
## Summary

Add an optimized prefill kernel for amd mi355x.

## Test Plan

For gpt-oss shape 64 q heads / 8 kv heads / head dim 64, tested on mi355x:

| Batch | SeqLen | Current Triton Kernel | Gluon |
|---:|---:|---:|---:|
| `2` | `8192` | `5087.541 us` | `1462.470 us` |
| `16` | `1024` | `871.163 us` | `316.721 us` |

### DIFF
--- a/tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/flashinfer/__init__.py
+++ b/tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/flashinfer/__init__.py
@@ -92,6 +92,7 @@ if platform.is_nvidia and platform.is_blackwell:
         dtypes={torch.float16, torch.bfloat16},
         priority=Priority.SPECIALIZED + 2,
         traits={
+            "head_dim": frozenset({128}),
             "sliding_window": frozenset({False, True}),
             "support_sinks": frozenset({False, True}),
             "support_logit_cap": frozenset({False}),
--- a/tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/flashinfer/__init__.py
+++ b/tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/flashinfer/__init__.py
@@ -87,7 +87,7 @@ def _get_ragged_prefill_wrapper(
     wrapper = _ragged_prefill_wrappers.get(device)
     if wrapper is None:
         workspace = torch.empty(
-            128 * 1024 * 1024,
+            256 * 1024 * 1024,
             dtype=torch.uint8,
             device=device,
         )
@@ -111,7 +111,7 @@ if platform.is_nvidia and platform.is_blackwell:
         dtypes={torch.float16, torch.bfloat16},
         priority=Priority.SPECIALIZED + 1,
         traits={
-            "head_dim": frozenset({64, 128}),
+            "head_dim": frozenset({64, 128, 256}),
             "sliding_window": frozenset({False, True}),
             "support_sinks": frozenset({False}),
             "support_logit_cap": frozenset({False, True}),
--- a/tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/flashinfer/__init__.py
+++ b/tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/flashinfer/__init__.py
@@ -40,7 +40,6 @@ trtllm_batch_context_with_kv_cache = error_fn
 trtllm_batch_decode_with_kv_cache = error_fn
 trtllm_batch_decode_with_kv_cache_mla = error_fn
 trtllm_ragged_attention_deepseek = error_fn
-flashinfer_mha_prefill_ragged = error_fn
 
 if platform.is_nvidia:
     try:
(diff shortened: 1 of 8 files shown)


tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/gluon/mha_prefill_fp16_gfx950.py
```diff
@@ -0,0 +1,1071 @@
...
+
+"""MHA prefill Gluon kernel optimized for AMD GFX950."""
+
+from __future__ import annotations
+
+import math
+from typing import NamedTuple
+
+import tokenspeed_triton as triton
+import tokenspeed_triton.experimental.gluon.language as gl
```
antiagainst: We should `from tokenspeed_kernel._triton import ..` to have single point of triton control.

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/gluon/mha_prefill_fp16_gfx950.py
```diff
@@ -0,0 +1,1071 @@
...
+
+
+@gluon.aggregate
+class InputLayout:
+    stride_t: gl.constexpr
+    stride_h: gl.constexpr
+    stride_d: gl.constexpr
+
+    @gluon.constexpr_function
+    def __init__(self, stride_t, stride_h, stride_d):
```
antiagainst: This should not be needed after https://github.com/triton-lang/triton/commit/0a15086d7dde9c8b4415fcc81ebf188fafd3b7f6?

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/gluon/mha_prefill_fp16_gfx950.py
```diff
@@ -0,0 +1,1071 @@
...
+    p_layout: gl.constexpr
+    v_layout: gl.constexpr
+    load_layout: gl.constexpr
+    store_layout: gl.constexpr
+    k_smem_layout: gl.constexpr
+    v_smem_layout: gl.constexpr
+
+    @gluon.constexpr_function
+    def __init__(
+        self,
```
antiagainst, line 104: Nit: I'd typically group multiple ones into one line to be a bit condensed. (Similarly for other long lists.)

antiagainst (commented):
Nice!

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/flashinfer/__init__.py
```diff
@@ -92,6 +92,7 @@
         dtypes={torch.float16, torch.bfloat16},
         priority=Priority.SPECIALIZED + 2,
         traits={
+            "head_dim": frozenset({128}),
```
raikonenfnu: QQ, is the flashinfer path somehow related to this gluon kernel addition/change?

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/gluon/mha_prefill_fp16_gfx950.py
raikonenfnu, line 1: Thoughts on having a `gluon/cdna4/mha_prefill_fp16.py` to make it a bit more organized in case new kernels from different variations + generations come along? or do we think it is too soon?

borontion: @antiagainst what do you think

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/flashinfer/__init__.py
borontion: there is a change in the mha backend to switch using standard prefill instead of first writing kv to cache then prefill read from kv cache. standard prefill routes to this flashinfer kernel and it does not support gpt-oss shape.

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/gluon/mha_prefill_fp16_gfx950.py
borontion: looks like our formatter is not happy with that. didn't see a quick workaround for it. may need to tweak the formatter config later.

antiagainst: Trick is to drop the last `,` I think, like [here](https://github.com/triton-lang/triton/blob/4768da5e8228dfbda8e0b7a61101f87d953341bd/python/tutorials/03-matrix-multiplication.py#L234-L247). Not a big deal though--feel free to leave it as is

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/gluon/mha_prefill_fp16_gfx950.py
antiagainst: I think this is fine for now--I do see that we many need to support more variants but we can refactor later given it depends on what/how we can share different pieces.

antiagainst (approved):
Great!

PR #169, opened 2026-05-17
fix: detect GPU arch automatically for kernel building

Dogacel:
## Summary

Detect GPU architecture automatically while building `tokenspeed-kernel`.

## Test Plan

1. (Old flow) Built kernel on GH200, however sm100a is not comptible by default, so I get `[ts] RuntimeError: Check failed: (status == cudaSuccess) is false: BatchQKApplyRotaryPosIdsCosSinCacheEnhanced failed with error code no kernel image is available for execution on the device` while running gpt-oss-20b.
2. (New flow) Clear packages, cache and re-run `pip install -e tokenspeed-kernel/python/ --no-build-isolation`. The GPU architecture is automatically detected without setting `FLASHINFER_CUDA_ARCH_LIST` or `TOKENSPEED_CUDA_ARCH` (which also fixes the issue).

### DIFF
--- a/tokenspeed-kernel/python/setup.py
+++ b/tokenspeed-kernel/python/setup.py
@@ -453,6 +453,30 @@ class CudaKernelBuilder:
             archs.add(self._normalize_cuda_arch(direct))
             return archs
 
+        # Try detecting from the NVIDIA driver.
+        if not archs:
+            try:
+                caps = (
+                    subprocess.check_output(
+                        [
+                            "nvidia-smi",
+                            "--query-gpu=compute_cap",
+                            "--format=csv,noheader",
+                        ],
+                        text=True,
+                        stderr=subprocess.DEVNULL,
+                    )
+                    .strip()
+                    .splitlines()
+                )
+                for cap in caps:
+                    cap = cap.strip()
+                    if cap:
+                        archs.add(self._normalize_cuda_arch(cap + "a"))
+            except (OSError, subprocess.CalledProcessError):
+                pass
+
+        # Fallback: Blackwell
         if not archs:
             archs.add("100a")
         return archs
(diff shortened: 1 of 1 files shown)


tokenspeed-kernel/python/setup.py
```diff
@@ -453,6 +453,30 @@ def _detect_cuda_archs(self):
...
+                        text=True,
+                        stderr=subprocess.DEVNULL,
+                    )
+                    .strip()
+                    .splitlines()
+                )
+                for cap in caps:
+                    cap = cap.strip()
+                    if cap:
+                        archs.add(self._normalize_cuda_arch(cap + "a"))
```
borontion: why there is a suffix "a" here - `cap + "a"`. shouldn't it be handled by `_normalize_cuda_arch`?

Dogacel: I am a bit confused, it seems like we can remove.

The goal was to get sm100a, sm90a, sm120a etc. The normalize does,

```
        suffix = "a" if has_suffix or major >= 9 else ""
        return f"{major}{minor}{suffix}"
```

So it will always has "a", BUT when I tried to compile tokenspeed on sm80, I never succeeded so I couldn't test.

So just removing it should result in identical code for supported architectures, I'll do that.

Dogacel:
Hi @borontion just wondering why this is closed. The issue is not only about sm80.

borontion:
> Hi @borontion just wondering why this is closed. The issue is not only about sm80. 

Sorry I didn't follow this afterwards. What's the device and cuda version you are using? We will only cover devices listed in CI.

Dogacel:
> > Hi @borontion just wondering why this is closed. The issue is not only about sm80.
> 
> Sorry I didn't follow this afterwards. What's the device and cuda version you are using? We will only cover devices listed in CI.

I was using GB200 / RTX PRO 6000 for doing development.

PR #171, opened 2026-05-17
perf(moe): triton biased grouped topk for deepseek-v3 routing

roycho96:
## Summary

DeepSeek-V3/V4 expert routing (`biased_grouped_topk_gpu` with `num_expert_group=8, topk_group=4, topk=8, num_fused_shared_experts ∈ {0,1}`) runs as a `@torch.compile` reference: 8-10 launches per layer per decode step. The MiniMax sibling kernel at `tokenspeed-kernel/ops/moe/triton.py:227` is trait-gated to `num_expert_group=1, topk_group=1` and falls through for DSv3 shapes.

This PR adds a sibling Triton kernel `_grouped_topk_biased_kernel` plus `triton_biased_grouped_topk` wrapper registered on the DSv3 trait. The kernel folds sigmoid, bias, group-top-2-sum, group selection, per-token top-K, renorm, and the shared-expert weight into one program per token. Dispatch is handled by the existing trait filter; no runtime changes.

Microbench (fp32 gating, 30 warmup + 400 timed iters, cudaEvent + cudaSync):

| shared | batch | ref µs | trt µs | speedup |
|--------|-------|--------|--------|---------|
|      0 |     1 | 100.51 |  18.41 |   5.46x |
|      0 |    32 | 103.53 |  17.94 |   5.77x |
|      0 |   256 | 100.52 |  17.76 |   5.66x |
|      1 |     1 | 130.49 |  17.74 |   7.36x |
|      1 |    32 | 129.69 |  18.12 |   7.16x |
|      1 |   256 | 129.54 |  17.74 |   7.30x |

bf16 gating falls back to reference (bitwise-identical, verified by test).

## Test Plan

```
pytest tokenspeed-kernel/test/ops/test_moe_routing.py
```


LorrinWWW (approved):

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/triton.py
```diff
@@ -430,6 +431,258 @@ def minimax_biased_grouped_topk(
...
 
 
+@triton.jit
+def _grouped_topk_biased_kernel(
+    gating_output_ptr,
+    correction_bias_ptr,
+    static_logical_to_physical_map_ptr,
+    topk_weights_ptr,
+    topk_ids_ptr,
+    stride_gm,
```
borontion, line 440: should stride be constexpr?

roycho96: output strides here are always (TOPK, 1) since we allocate the outputs inside the wrapper, so I just dropped them instead of constexpr. left the input strides.

PR #182, opened 2026-05-19
fix scheduler req pool move assignment leak

curious-whq:
## Summary

Fix `ReqPoolIndex` move assignment to release the slot currently owned by the destination before taking ownership from the source.

`ReqPoolIndex` is an RAII handle: its destructor returns the held slot to `ReqPoolAllocator`. The previous move-assignment operator overwrote `slot_` and `allocator_` directly, so assigning into an already-valid `ReqPoolIndex` leaked the destination's old slot.

## Tests

- `./tokenspeed-scheduler/build-test/tokenspeed_scheduler_tests --gtest_filter='ReqPoolAllocatorTest.*'`


zhyncs (approved):
To be honest, this is not a real bug. We only use `ReqPoolAllocator` with unique ptr. And it's fine, we can merge this.

PR #194, opened 2026-05-20
fix: fix the bug that kv cache pool size is too small.

tuanzhangCS:
## Summary

- Fix KV cache pool under-allocation for hybrid Mamba-attention models. When computing the KV cache budget for models with both Mamba and attention layers (e.g., Jamba), the profiler was using num_attention_layers from _profile_kwargs, which defaulted to the total layer count (num_layers). This caused the per-page memory estimate to be inflated by counting Mamba layers as if they needed KV cache, resulting in fewer pages being allocated than necessary. The fix computes the true number of attention-only layers — either from full_attention_layer_ids in the model config or by subtracting the Mamba layer count — and passes it explicitly to profile_max_num_pages / profile_cache_budget.
- Fix KV cache budget for MTP NextN draft models. For draft workers using MTP NextN models, num_hidden_layers in the text config could reflect the full target model rather than the draft model. This caused the cache budget to be computed for the wrong number of layers. The fix overrides num_hidden_layers with mtp_num_hidden_layers when running as a draft worker.

### DIFF
--- a/.github/workflows/pr-test.yml
+++ b/.github/workflows/pr-test.yml
@@ -134,6 +134,7 @@ jobs:
 
       - name: Install CI task
         run: |
+          rm -rf /home/runner/.cache/flashinfer || true
           cd "${{ env.WORK_DIR }}"
           mkdir -p .ci-artifacts
           python3 test/ci_system/pipeline.py execute \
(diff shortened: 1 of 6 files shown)


minedec (approved):

tuanzhangCS:
> @tuanzhangCS https://github.com/lightseekorg/tokenspeed/actions/runs/26161367390/job/76953845254?pr=194 score 0

It's strange—the benchmark scores are normal when I run it locally. let me re-run this CI.

PR #207, opened 2026-05-21
fix(deepseek-v4): close MTP acceptance gap

Xiangyi1996:
## Summary
++This PR is a bit large because it contains the V4/R1 MTP stack plus the acceptance-gap fix and follow-up CI/review fixes.

This PR closes the DeepSeek V4 MTP acceptance gap between TokenSpeed and TRTLLM.

Root cause:
- The remaining gap was not from compressed KV / CSA indexer cache.
- It came from MTP draft decode using stale/incorrect V4 paged KV cache metadata.
- V4 has multiple cache tables; the SWA compact table could be observed with the wrong request context during draft/target-verify transitions.

Fix:
- Make target-verify/draft-extend forward modes explicit.
- Refresh paged cache group metadata for MTP draft/target paths.
- Carry V4 SWA/compressed KV/CSA metadata consistently through draft decode.
- Keep target-verify logits/hidden states correctly for speculative decoding.
- Add tests for V4 SWA slot sanitization / paged metadata behavior.

## Validation

- `py_compile` on touched runtime/test files: passed
- Acceptance rerun after rebase:
  - `Decoded Tok/Iter = 2.8447`
  - `Spec Accept Rate = 0.6485`
  - In TRTLLM 2.8-2.9 range


python/tokenspeed/runtime/models/deepseek_v4.py
```diff
@@ -152,6 +154,37 @@
...
+    if ctx.forward_mode == ForwardMode.EXTEND:
+        return getattr(ctx.attn_backend, "forward_prefill_metadata", None) or metadata
+    if ctx.forward_mode is not None and ctx.forward_mode.is_draft_extend():
+        prefill_metadata = getattr(ctx.attn_backend, "forward_prefill_metadata", None)
+        if _deepseek_v4_metadata_matches_tokens(
+            prefill_metadata,
+            ctx.input_num_tokens,
+        ):
+            return prefill_metadata
+        return prefill_metadata or metadata
```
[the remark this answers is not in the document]
Xiangyi1996: Fixed in current head e4223a0. Draft graph replay now forwards paged cache-group metadata to V4 draft backends, and non-V4 draft backends keep the draft_seq_lens_buf alias.

python/tokenspeed/runtime/engine/scheduler_utils.py
```diff
@@ -163,6 +163,37 @@ def pool_to_prefix_cache_adjunct_spec(
...
+    speculative_algorithm: Any | None,
+    paged_cache_groups: Sequence["PagedCacheGroupConfig"] | None = None,
+) -> bool:
+    """Return whether the runtime can use the overlapped scheduler loop."""
+
+    if disable_overlap_schedule:
+        return False
+    if disaggregation_mode == "prefill":
+        return False
+    if speculative_algorithm is not None and paged_cache_groups:
```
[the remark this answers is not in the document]
Xiangyi1996: The current PR is focused on closing the V4 MTP acceptance gap and fixing the paged-cache metadata propagation for target-verify / draft-extend paths. Overlap changes the scheduling and metadata lifetime further, so I would prefer not to enable it in the same PR before we have separate validation.

We may first land the correctness/acceptance fix, then validate V4 MTP + overlap in a follow-up perf/serving PR.

python/tokenspeed/runtime/models/deepseek_v4_mtp.py
```diff
@@ -0,0 +1,524 @@
...
+        positions: torch.Tensor,
+        previous_hidden_states: torch.Tensor,
+        ctx: ForwardContext,
+        out_cache_loc: torch.Tensor,
+        input_embeds: Optional[torch.Tensor] = None,
+        spec_step_idx: int = 0,
+    ) -> torch.Tensor:
+        if input_embeds is None:
+            input_embeds = self.embed_tokens(input_ids)
+        current_step_idx = spec_step_idx % self.num_mtp_layers
```
SimonCqk, line 245: its caller `ModelRunner.forward` seems doesn't pass `spec_step_idx` in both explicit arguments and kwargs, so it is always 0?

Xiangyi1996: You are right. `spec_step_idx` is currently not propagated through `ModelRunner.forward`, so the V4 MTP draft model would always use the default step 0.

I’ll patch this by threading an optional `spec_step_idx` through `ModelRunner.forward`, guarded so it is only forwarded to models whose `forward()` supports it. Non-V4 draft models should keep the existing behavior.

Xiangyi1996: Fixed in [1d90c5c](https://github.com/lightseekorg/tokenspeed/pull/207/commits/1d90c5cf399748910beb7ba6b7ef4afddb8770e6).

`spec_step_idx` is now threaded through `ModelRunner.forward`, and only forwarded when the model `forward()` signature supports it, so non-V4 draft models keep the existing behavior.

Xiangyi1996:
This PR is a bit large because it contains the V4/R1 MTP stack plus the acceptance-gap fix and follow-up CI/review fixes.

The main risk areas are:
1. ForwardMode changes: target_verify / draft_extend should only affect V4 paged-cache MTP paths.
2. V4 paged cache metadata propagation: SWA / compressed KV / CSA indexer group tables and base offsets.
3. Draft decode step handling: spec_step_idx and accepted-prefix advancement.
4. Guards: prefix-cache and overlap scheduling are disabled for speculative V4 paged-cache groups.
5. Non-V4 compatibility: existing draft/spec paths should continue using their previous metadata behavior.

Validation so far:
- Acceptance recovered from ~2.10 to ~2.84 decoded tok/iter, back to TRTLLM 2.8-2.9 range.
- Local b200 CI passed: ut-runtime-minimax-m2 / b200-2gpu and ut-runtime-1gpu / b200-1gpu.
- Kimi b200-4gpu eval is being rerun once GPU/cache environment is available.

dongjiyingdjy (commented):
## Code Review

23 files, +3096/-180. Enables MTP speculative decoding for V4, fixes acceptance gap vs TRTLLM (2.8447 tok/iter match). Core approach — separate `TARGET_VERIFY`/`DRAFT_EXTEND` modes with V4-specific metadata lifecycle — is sound.

### Issues to address

**1. `_forward_flashmla_sparse` per-token Python loop (perf)**

The new sparse attention method iterates per-token in Python with GPU ops inside each iteration. If this is on the MTP draft hot path, it will dominate latency. Please confirm this is only a correctness fallback, or add a guard/warning.

**2. Mixed batch disabled for ALL speculative models, not just V4 (regression risk)**

```python
enable_mixed_prefill_decode = (
    server_args.enable_mixed_batch and server_args.speculative_algorithm is None
)
```

This disables mixed batch whenever any speculative algorithm is active, including EAGLE3 on non-V4 models. If mixed batch previously worked with EAGLE3, this is a regression. Should this be V4-specific?

**3. `_select_decode_metadata` fallback chain is fragile for CUDA graph replay**

The method searches through `forward_metadata`, `forward_decode_metadata`, `forward_prefill_metadata` by matching `num_tokens`. If a stale graph replay happens to match the wrong metadata's token count, it silently reads wrong data. The RuntimeError fallback is fine for eager but catastrophic inside a captured graph.

**4. `_refresh_cuda_graph_packed_metadata` allocates temporaries every replay**

```python
self._cuda_graph_query_start_loc[:bs+1].copy_(torch.arange(bs+1, ...) * tokens_per_req)
self._cuda_graph_token_to_req[:total_tokens].copy_(torch.arange(bs, ...).repeat_interleave(tokens_per_req))
```

`torch.arange` + `repeat_interleave` on every CUDA graph replay step. These could be precomputed once during capture and reused.

### Suggestions

**5. Repeated `is_target_verify() or is_draft_extend()` pattern (~15 occurrences)**

Consider adding a `ForwardMode.is_speculative()` helper to reduce repetition and ensure consistent handling.

**6. `num_tokens` fallback in `init_forward_metadata` uses stale spec config**

When neither `positions` nor explicit `num_tokens` is provided for speculative modes, the code estimates from `speculative_num_draft_tokens` or `speculative_num_steps` set at init time. After partial accept the actual count could differ. Callers in `CudaGraphWrapper` do pass `num_tokens` explicitly, so this is a latent edge case, not a current bug.

### Looks good

- Test coverage is thorough (metadata shape transitions, non-V4 backend contracts, SWA slot sanitization)
- `cache_start.clone()` correctly prevents drafter from modifying shared `seq_lens_buf`
- SWA slot sanitization + CUDA kernel guard is defense-in-depth
- `_init_capture_metadata` ordering (before warmup + before capture) is correct, though a comment explaining why it runs twice would help

dongjiyingdjy:
Supplement: 6 new functions in this PR are never called — dead code. Suggest removing them:

- `_forward_flashmla_sparse` and its 3 helpers (`_slots_from_local_indices`, `_swa_slots_for_token`, `_compressed_slots_for_token`)
- `_deep_gemm_fp8_linear`
- `_prefill_chunk_token_offsets`

If these are reserved for follow-up work, please don't include them in this PR — add them when they're actually wired in.

Xiangyi1996:
Thanks for the detailed review. I addressed the items one by one:

1. `_forward_flashmla_sparse` per-token Python loop

Removed `_forward_flashmla_sparse` and its helper methods. The decode/speculative decode now always goes through `forward_deepseek_v4_decode`.

2. Mixed batch disabled for all speculative models

Fixed. I replaced the global `speculative_algorithm is None` guard with `should_enable_mixed_prefill_decode()`. Mixed prefill/decode is now disabled only for speculative runs with V4 paged-cache groups; non-V4 speculative backends such as EAGLE3 keep the previous mixed-batch behavior. 

3. `_select_decode_metadata` fallback chain

 `_select_decode_metadata()` no longer falls back to `forward_prefill_metadata`. It only returns matching decode metadata, otherwise `None`. 

4. CUDA graph replay temporary allocations

Fixed. `_refresh_cuda_graph_packed_metadata()` now reuses precomputed `query_start` and `token_to_req` , instead of creating `torch.arange()` / `repeat_interleave()` temporaries during replay.

5. Repeated `is_target_verify() or is_draft_extend()`

Added `ForwardMode.is_speculative()` relpace is_target_verify() or is_draft_extend()

6. `num_tokens` fallback using stale speculative config

Fixed. V4 speculative metadata setup now requires explicit `num_tokens` or `positions`

dead code:

Removed the unused `_forward_flashmla_sparse` helpers, `_deep_gemm_fp8_linear`, and `_prefill_chunk_token_offsets`. Chunked prefill now directly uses `metadata.query_start_loc` for token offsets.

Xiangyi1996:
## Update on follow-ups since last review

Addressed chatgpt-codex-connector's SWA cache padded-token concern,
and expanded the fix to all 5 V4 cache insert paths:

- SWA KV cache (`_insert_swa_cache`)
- Compressed KV cache (`compressed_slot_mapping`)
- Compressor state (`DeepseekV4Compressor`)
- Indexer state (`DeepseekV4Indexer`)
- CSA indexer cache (via compressed_slot_mapping plumbing)

New tests:
- `test_slot_mapping_guard_masks_invalid_graph_tokens`
- `test_swa_slot_mapping_guard_masks_invalid_graph_tokens`
- `compressed_slot_mapping` w/ `is_valid_token` masking test

Local validation:
- Focused pytest: SWA / V4 ops / V4 config / logits processor all pass
- DP=4 E2E sanity: 4/4 OK, accept_rate ~0.61, avg_accept_len ~2.82
- TP=2 E2E sanity: 2/2 OK, no NCCL hang, avg accept ~2.52
  - TP=2 caught and fixed a per-token req-index broadcasting issue in
    `_group_slot_mapping_from_raw` for packed-token draft path

Ready for re-review.

python/tokenspeed/runtime/execution/cuda_graph_wrapper.py
```diff
@@ -616,41 +671,71 @@ def _init_forward_metadata(
...
-                bs=padded_bs,
-                num_extends=num_extends,
-                req_pool_indices=req_pool_indices,
-                seq_lens=draft_seq_lens,
-                req_to_page=self.drafter.req_to_page,
-                forward_mode=forward_mode,
-                **kwargs,
-            )
+            draft_seq_lens.copy_(seq_lens[:padded_bs])
+            if forward_mode == ForwardMode.EXTEND or forward_mode.is_mixed():
```
dongjiyingdjy: Can we use consistent wording?

python/tokenspeed/runtime/execution/model_executor.py
```diff
@@ -153,6 +154,10 @@ def from_server_args(
             spec_algo=server_args.speculative_algorithm,
             spec_num_steps=server_args.speculative_num_steps,
             spec_num_tokens=server_args.speculative_num_draft_tokens,
+            use_target_verify_forward_mode=(
+                server_args.speculative_algorithm is not None
+                and is_deepseek_v4(model_config.hf_config)
```
dongjiyingdjy: could we use `model_config.use_target_verify_forward_mode` instead？

python/tokenspeed/runtime/engine/scheduler_utils.py
```diff
@@ -163,6 +163,54 @@ def pool_to_prefix_cache_adjunct_spec(
     return spec
 
 
+def should_disable_prefix_cache(
```
dongjiyingdjy: I suggest that we not directly disable these features in the code, but instead claim that MTP and prefix cache/mixed batch cannot be enabled together for now in doc.

yweng0828 (approved):

python/tokenspeed/runtime/execution/cuda_graph_wrapper.py
Xiangyi1996: Fixed in 01942d5. changed this to `forward_mode.is_extend_or_mixed()`

python/tokenspeed/runtime/execution/model_executor.py
Xiangyi1996: Fixed in 01942d5. moved the target-verify gating into `ModelConfig.use_target_verify_forward_mode` and `ModelExecutorConfig` now reads that field instead of duplicating the inline check.

python/tokenspeed/runtime/engine/scheduler_utils.py
Xiangyi1996: Fixed in 01942d5

python/tokenspeed/runtime/layers/attention/kv_cache/deepseek_v4.py
```diff
@@ -515,17 +561,27 @@ def _update_decode_compressed_slot_mapping(
...
+                        device=page_indices.device,
+                        dtype=torch.int64,
+                    )[req_idx]
                 )
+            page_ids = _safe_page_ids(block_table, req_idx, page_indices)
+            slot_mapping = torch.where(
+                page_ids >= 0,
+                page_ids * kv_cache_block_size + offsets,
+                torch.full_like(page_ids, -1),
             )
```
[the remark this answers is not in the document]
Xiangyi1996: Addressed in 01942d5. I added a compression-boundary validity mask for paged compressed-cache slot mapping, so non-boundary tokens now map to `-1` instead of writing into compressed/indexer cache slots.

python/tokenspeed/runtime/configs/qwen3_5_config.py
```diff
@@ -136,3 +151,8 @@ class Qwen3_5MoeConfig(Qwen3_5Config):
         "vision_config": Qwen3_5MoeVisionConfig,
         "text_config": Qwen3_5MoeTextConfig,
     }
+
+    def __init__(self, **kwargs):
+        # Explicit __init__ prevents transformers from auto-generating one
+        # that skips Qwen3_5Config.__init__ (text/vision config setup).
+        super().__init__(**kwargs)
```
Xiangyi1996, line 158: **Why this Qwen3.5 fix is in a V4 MTP PR**

This PR adds new attribute-access paths in `ModelConfig.__init__` (via
`_derive_num_attention_layers` / `is_deepseek_v4_nextn`) to support V4 MTP.
Those new paths happen to exercise the Qwen3.5 config in a way that triggers
a **latent bug on main** introduced by transformers 5.6:

- `Qwen3_5MoeConfig` had no explicit `__init__`
- transformers 5.6 auto-generates one for subclasses, **bypassing**
  `Qwen3_5Config.__init__`
- `text_config` is therefore left in an uninitialized/wrapped form
- Subsequent `__getattr__` forwards loop on `text_config` itself → RecursionError

Reproduced on `origin/main` (without this PR) using a minimal
self-referencing `text_config` setup; the same trace matches the CI
failure at `qwen3_5_config.py:114-115`.

@yweng0828 confirmed his other PR doesn't hit this because it doesn't go
through the new `ModelConfig` path.

Discussed with @dongjiyingdjy and @yweng0828: keeping the fix in this PR
since it's needed to unbreak CI here. Happy to cherry-pick to a standalone
follow-up if preferred.

CI evidence (pre-fix): https://github.com/lightseekorg/tokenspeed/actions/runs/26626283458/job/78483017501

python/tokenspeed/runtime/layers/attention/backends/deepseek_v4.py
```diff
@@ -1616,18 +2029,53 @@ def init_forward_metadata_replay_cuda_graph(
...
+            or self._draft_decode_metadata is None
+        ):
+            raise RuntimeError("DeepSeek V4 draft metadata was not initialized")
+        self._draft_decode_step += 1
+        metadata = self._draft_decode_metadata
+        if seq_lens is None:
+            metadata.seq_lens.add_(1)
+        else:
+            metadata.seq_lens.copy_(seq_lens[: metadata.seq_lens.numel()])
+        metadata.forward_mode = ForwardMode.DECODE
```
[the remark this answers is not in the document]
Xiangyi1996: I think this is already handled in the current commit. After refreshing the draft decode metadata, `advance_draft_forward_metadata()` publishes it back to both `forward_decode_metadata` and `forward_metadata`:

```python
_refresh_decode_indexer_schedule_metadata(metadata)
self.forward_decode_metadata = metadata
self.forward_metadata = metadata
self._decode_tile_metadata = {}

PR #216, opened 2026-05-22
feat(kernel): add quantization kernel api

borontion:
## Summary

Add shared quantization api for fp8 | fp8 with scale | mxfp8 | mxfp4 | nvfp4, backed by trtllm/flashinfer/triton kernels. Replace the torch impl of quant in runtime.

## Test Plan

Test on H100/B200/MI355.

```sh
pytest tokenspeed-kernel/test/ops/test_quantization.py"
```

### DIFF
--- a/tokenspeed-kernel/python/tokenspeed_kernel/ops/quantization/__init__.py
+++ b/tokenspeed-kernel/python/tokenspeed_kernel/ops/quantization/__init__.py
@@ -95,13 +95,7 @@ def quantize_fp8_with_scale(
     # quantization options
     granularity: Literal["tensor", "token", "token_group"] = "tensor",
     group_size: int | None = None,
-    num_token_padding: int | None = None,
-    scale_dtype: torch.dtype = torch.float32,
-    scale_layout: Literal[
-        "row_major", "column_major", "tma_aligned", "linear"
-    ] = "row_major",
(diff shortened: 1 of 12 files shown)


tokenspeed-kernel/python/tokenspeed_kernel/ops/quantization/__init__.py
```diff
@@ -14,12 +14,409 @@
...
+
+    MXFP8 uses FP8 data plus encoded vector scales, commonly one scale per 32
+    values along the last dimension.  Backends may pad the last dimension to
+    alignment and may return a flat encoded scale buffer whose layout is
+    specified by scale_layout.
+
+    Args:
+        x: Input tensor.
+        alignment: Last-dimension alignment requested from the backend.
+        scale_layout: Scale-factor layout, such as linear, 128x4, or 8x4.
```
antiagainst: What does `scale_layout` 128x4 and 8x4 mean? How are they used?

antiagainst (approved):
Nice. I just have one question.

tokenspeed-kernel/python/tokenspeed_kernel/ops/quantization/__init__.py
borontion: It is some leftover. Will clean it up soon.

PR #230, opened 2026-05-23
Use operand format signatures for kernel selection

antiagainst:
Replace KernelSpec.dtypes with format_signatures so registrations can describe role-specific operand tensor formats instead of a flat dtype set. Model each tensor role with TensorFormat and optional ScaleFormat, using storage_dtype consistently for payload and scale sidecar storage.

Convert in-tree GEMM, attention, embedding, quantization, MoE, reference, and plugin registrations to the new FormatSignature API. Move quantized GEMM and MoE fused format choices out of selection traits, with MoE fused call sites passing weight_format explicitly.

Update selection, numerics, benchmark, docs, and tests to resolve by full FormatSignature while preserving dtype-oriented convenience filters through primary_storage_dtype helpers.


tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/flashinfer.py
```diff
@@ -21,11 +21,106 @@
...
+)
+_NVFP4_SCALE = ScaleFormat(
+    storage_dtype=torch.float32,
+    granularity="block",
+)
+_MXFP4_SCALE = ScaleFormat(
+    storage_dtype=torch.uint8,
+    granularity="block",
+    block_shape=(32,),
+)
```
borontion, line 52: why do we only specify the block shape for mxfp4 but not nvfp4 and fp8 here?

tokenspeed-kernel/python/tokenspeed_kernel/signature.py
```diff
@@ -0,0 +1,247 @@
...
+        return "scale(" + ", ".join(parts) + ")"
+
+
+@dataclass(frozen=True)
+class TensorFormat:
+    """Metadata representation for one logical tensor.
+
+    Args:
+        storage_dtype: Physical dtype used by the main tensor payload.
+        format: Logical representation format, such as "dense", "fp8",
```
borontion: should fp8 be considered as dense? I saw somewhere like 

```python
    signatures=format_signatures(
        "x", "dense", {torch.float16, torch.bfloat16, torch.float8_e4m3fn}
    ),
```

but other places like:

```python
weight=tensor_format("fp8", torch.float8_e4m3fn, scale=_FP8_SCALE)
```

antiagainst: It's a bit subtle here. "dense" basically means non-scaled. for fp8, we can have cases where we have no scale values (the former), and where we have custom scale schemes (the latter).

antiagainst: It's due to MoE; there are some docs [here](https://github.com/lightseekorg/tokenspeed/pull/230/changes#diff-713dc35252c64224fba31dd0e08720d3427171e4487bfb333db9d36b53ff751d) which I think worth moving to the common place. I find it a bit struggling to find better names to reflect these differences. Any better suggestions are welcome!

borontion (approved):
Thanks for the help! It is really useful.

tokenspeed-kernel/python/tokenspeed_kernel/signature.py
borontion: > It's a bit subtle here. "dense" basically means non-scaled. for fp8, we can have cases where we have no scale values (the former), and where we have custom scale schemes (the latter).

I see this makes sense. maybe we should document this "dense fp8"  and "fp8" case somewhere or it looks a little confusing.

antiagainst: Renamed to be "scaled-fp8" to be clear in https://github.com/lightseekorg/tokenspeed/pull/230/commits/d7972e9c1a9f8cd4ae6e906c019d73c44b3d433f

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/flashinfer.py
antiagainst: tightened it up with https://github.com/lightseekorg/tokenspeed/pull/230/commits/e2dd4cb77a801442014300423777e413e8c8bb43

tokenspeed-kernel/python/tokenspeed_kernel/numerics/gemm.py
```diff
@@ -119,45 +120,97 @@ def _generate_scales(self, shape: tuple[int, ...], dtype) -> torch.Tensor:
...
+                raise ValueError(
+                    "mxfp8 block scale format requires concrete block_shape"
+                )
+            block_n, block_k = block_size
+            k_tiles = math.ceil(K / block_k)
+            if role == "a":
+                return self._generate_scales((M, k_tiles), scale.storage_dtype)
+            if role == "b":
+                n_tiles = math.ceil(N / block_n)
+                return self._generate_scales((n_tiles, k_tiles), scale.storage_dtype)
```
knwng: For mxfp8 block scale, we should need to know the block size along K. Here's a little strange to see non-symmetric block_size.

PR #232, opened 2026-05-24
Support DP sampling for spec decode

yubofredwang:
## Summary

Enable data parallele sampling. This has a few benefits:
1. No need to use deterministic kernel during larger batch. Potentially better perf.
2. Reduce the all gather communication overhead by calculating needed part and use all to all single operation.
3. Avoid duplicate work from ranks in TP mode. No need to sync verify results between ranks, but rather, gather the results.

Implemented:
- swap vocab and batch dimension
- onesided all2all kernel
- padding for cuda graph and even split
- Support logprobs
- Support xgrammar

## Design

image

## Test Plan
Ran Kimi-K2.5 TP8 FP8 KV, batch 40, EAGLE3 spec 5/1/6, ctx=5000, max_tokens=4096, comparing TP sampling baseline vs DP sampling onesided.

```
mode | round 1 tok/s | round 2 tok/s | round 3 tok/s | avg rounds 2-3
-- | -- | -- | -- | --
TP sampling | 4475.0 | 5227.5 | 4956.8 | 5092.2
DP sampling | 4333.1 | 5543.3 | 5238.8 | 5391.1
```
DP sampling was about +5.9% on warm rounds 2-3. 

```
Run | Eval batch | DP gate | Route coverage | AIME score
-- | -- | -- | -- | --
Initial AIME | 16 | default min_bs=16 | bs=16 used DP sampling; smaller tail batches fell back | 0.9667
Large batch AIME | 30 | default min_bs=16 | bs=30 through bs=16 used DP sampling; bs<16 fell back | 0.9000
Full DP AIME | 30 | min_bs=1 | all decode batch sizes from bs=30 down to bs=1 used DP sampling | 0.9333
```

## Follow up
- Support full flashinfer backend including min_p, frequency penalty etc...
- Support dp sampling in case like EP8, DP2, TP4


python/tokenspeed/runtime/execution/context.py
```diff
@@ -58,3 +58,8 @@ class ForwardContext:
 
     # --- logits processor ---
     gather_ids: torch.Tensor | None = None
+    keep_full_logits: bool = False
+    last_index_offsets: torch.Tensor | None = None
+    padded_static_len: int = -1
```
LorrinWWW: Do we need `keep_full_logits` and `padded_static_len`?

python/tokenspeed/runtime/execution/cuda_graph_wrapper.py
LorrinWWW, line 1: Do we need `cuda_graph_wrapper.py` and `model_executor.py` to be aware of DP sampling? I feel it may be cleaner to keep the DP sampling decision internal to `logits_processor.py` or the sampling wrappers, rather than exposing it to the higher-level execution path (I'm good to pass flags via controllers but feel like threshold decision etc can be internalized).

yubofredwang: hmm I think we can't determine the graph bucket inside logits processor, because we need 1. forward path padding for other ops like MoE layer + 2. padding due to the bsz % tp_size

python/tokenspeed/runtime/execution/context.py
yubofredwang: removed

python/tokenspeed/runtime/execution/cuda_graph_wrapper.py
```diff
@@ -55,6 +56,38 @@
 logger = get_colorful_logger(__name__)
 
 _is_capture_mode = False
+_DP_SAMPLING_MIN_BS_ENV = "TOKENSPEED_DP_SAMPLING_MIN_BS"
+
+
+def resolve_dp_sampling_min_bs(tp_size: int, configured_min_bs: int | None) -> int:
+    env_value = os.environ.get(_DP_SAMPLING_MIN_BS_ENV)
```
syuoni: Can we manage this env var in the unified class？

https://github.com/lightseekorg/tokenspeed/blob/33ec6174157ec1030602c338418bacc64434db9a/python/tokenspeed/runtime/utils/env.py#L246-L259


In addition, I'm a bit concerned with enabling a feature with either env vars or CLI args. Can we use a single source of truth?

python/tokenspeed/runtime/execution/model_executor.py
```diff
@@ -282,6 +290,73 @@ def __init__(
...
+            and backend_supports_dp
+            and lm_head_supports_dp
+            and processor.tp_size > 1
+            and processor.tp_group is not None
+        )
+
+        dp_mode = os.environ.get(
+            "TOKENSPEED_DP_SAMPLING",
+            "on" if self.config.dp_sampling else "off",
+        ).lower()
```
syuoni: Same here.

python/tokenspeed/runtime/layers/logits_processor.py
```diff
@@ -199,6 +213,40 @@ def __init__(
         # Gate the fused lm_head GEMM to Kimi only. See ``_lm_head_matmul``.
         self._use_fused_lm_head = getattr(self.config, "model_type", None) == "kimi_k2"
 
+    @staticmethod
+    def supports_dp_sampling_lm_head(lm_head) -> bool:
+        return hasattr(lm_head, "weight")
```
syuoni: Can we inline these short wrappers? Also, I don't think it's necessary to check if lm_head has weight attribute.

python/tokenspeed/runtime/layers/logits_processor.py
```diff
@@ -199,6 +213,40 @@ def __init__(
...
+            )
+        return int(weight.shape[0]) * tp_size
+
+    @staticmethod
+    def validate_dp_sampling_lm_head(lm_head) -> None:
+        if not LogitsProcessor.supports_dp_sampling_lm_head(lm_head):
+            raise RuntimeError(
+                "dp_sampling requires a standard LM head with .weight; "
+                "GGUF linear_method is not supported on this path"
+            )
```
syuoni: Same here. Is there any strong reasons for these wrappers?

python/tokenspeed/runtime/execution/cuda_graph_wrapper.py
syuoni: I'm confused:
1. What is "forward path padding for other ops like MoE layer"?
2. Could "padding due to the bsz % tp_size" be checked inside logits processor?

syuoni (commented):
This PR currently introduces a wide range of changes across the whole runtime. I would expected this feature being scoped inside the sampling module only.

tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/dp_sampling.py
```diff
@@ -0,0 +1,447 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+"""One-sided NVLink collectives for Batch-DP speculative verify."""
```
syuoni: Seems according to the convention of tokenspeed_kernel, we should place the new kernel in `tokenspeed_kernel/ops/communication/triton.py`?  cc @borontion

yubofredwang:
@syuoni I think the bigger issue is because LogitsProcessor needs to know whether to run all_gather or let sampling module run the all-to-all depending on dp sampling is enabled or not for this batch. Any suggestion?

tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/dp_sampling.py
borontion: yes. our convention is `<family>/<solution>`.

yubofredwang: Fixed

python/tokenspeed/runtime/execution/cuda_graph_wrapper.py
yubofredwang: The blocker to moving this entirely into LogitsProcessor is that LogitsProcessor runs too late. By then we already needed to know:

DP sampling decision changes the logits/hidden-state communication contract. In current code, DP mode makes LogitsProcessor either slice hidden states before LM head when skip_all_gather is true, or use DP batch-vocab swap instead of normal TP logits all-gather. Non-DP uses the normal TP logits all-gather path.

But the deeper issue is that the decision must be made before logits processing so graph replay, logits layout, and sampling verify all agree. 
@LorrinWWW please take a look at the new design to use a logitslayout to represent the decision

python/tokenspeed/runtime/sampling/logits_layout.py
```diff
@@ -0,0 +1,251 @@
...
+        tp_size: int,
+        num_tokens_per_req: int,
+    ) -> "LogitsLayoutPlan":
+        return cls(
+            mode="dp_all_to_all",
+            real_bs=real_bs,
+            bucket_bs=bucket_bs,
+            tp_size=tp_size,
+            num_tokens_per_req=num_tokens_per_req,
+        )
```
syuoni: Why do we need these two ctor? It seems to be same to directly using `LogitsLayoutPlan(...)`

python/tokenspeed/runtime/models/minimax_m2.py
syuoni: Two branches can be merged: 

```python
        return LogitsProcessor(
            config,
            skip_all_gather=self.mapping.attn.has_dp,
            tp_rank=self.mapping.attn.tp_rank,
            tp_size=self.mapping.attn.tp_size,
            tp_group=self.mapping.attn.tp_group,
        )
```

This applies to other modeling files.

python/tokenspeed/runtime/sampling/backends/flashinfer.py
```diff
@@ -372,28 +517,64 @@ def verify(
...
+                ]
         # TP-rank sync: rank 0 wins on the full verify-output triple.
         # Load-bearing: flashinfer top_k_renorm_prob has no is_deterministic
         # knob and produces non-bit-identical results across ranks (sub-ulp
         # FP accumulation order).
         # For fused top-k + top-p, the results are bit-identical across ranks.
         # So we don't need to broadcast the results.
-        if not _FUSED_TOPK_TOPP_AVAILABLE:
+        elif pdl_enabled():
+            self.maybe_broadcast(predict, accept_index, accept_length)
```
syuoni, line 592: Do we still hit the out-of-sync issue for `_FUSED_TOPK_TOPP_AVAILABLE = True` and `pdl_enabled()`?

syuoni (approved):

python/tokenspeed/runtime/sampling/backends/flashinfer.py
yubofredwang: I think so. @yweng0828 need confirmation from you

yubofredwang: removed

PR #243, opened 2026-05-25
refactor(scheduler): unify hybrid cache model boundaries.

SimonCqk:
## Motivation

  The scheduler path previously mixed scheduling logic with Mamba and paged-cache-specific lifecycle details. This
  refactor makes the scheduler depend on a smaller hybrid-cache lifecycle boundary while keeping concrete family
  behavior inside `HybridPrefixCache`.

  This is a concrete module split, not a broad framework, plugin system, or universal allocator. Fuller family registry / policy abstractions are left as follow-ups once the common shape across more cache families is clearer.

## Summary

  - Make `HybridPrefixCache` the concrete lifecycle boundary for `MatchPrefix`, typed admission ops, `StepCommit`, `FinishRequest`.
  - Replace phase-specific request bags/enums with typed cache ops for admission, publication, materialization,
  request-local state, and worker metadata.
  - Split hybrid-cache implementation by responsibility:
    - `hybrid_prefix_cache_types.h` for typed lifecycle request/result structures and recovery-plan plumbing.
    - `mamba_family_ops.cpp` for Mamba-specific lifecycle behavior.
    - `paged_cache_family_ops.cpp` for paged-cache group admission, match, ownership, and cleanup behavior.
  - Move Mamba / paged-cache admission, match augmentation, publication, metadata, and cleanup logic out of scheduler/FSM code.
  - Remove unused `CacheCoordinator` scaffold and the unused scheduler prefetch operation implementation from the build.
  - Add/adjust tests around durable lifecycle behavior and core cache invariants instead of intermediate phase
  helpers.

## Notes for Reviewers

  A few follow-up-shaped pieces are deliberately not introduced in this PR to keep the size and review surface controlled:

  - `HybridPrefixCache` remains the concrete coordinator. The family-specific code is split into responsibility-
  focused files, but not yet abstracted behind a generic family registry or polymorphic interface.
  - `*_family_ops.cpp` favors direct readable family logic over premature framework shape. A fuller registry/policy
  layer will be introduced later if additional cache families prove a stable common lifecycle.
  - TreeNode attachment generalization is out of scope. This PR keeps the current concrete Mamba / paged-cache
  attachment model and focuses on scheduler boundary cleanup, will be a follow up to do the generalization after this merged.
  - Existing prefetch state/event types remain; this PR only removes the unused scheduler prefetch operation  implementation.

## Test Plan & Regression Status

  - Focused hybrid/Mamba/PagedCache tests: `69 passed`
  - Full scheduler C++ tests: `210 passed`
  - Regression test report **Ongoing** :
    - Accuracy (evalscope eval)
      - DeepSeek v4 Flash: gsm8k limit 200, accuracy `0.98`.
      - Kimi-K2.5-NVFP4: gsm8k limit 200, accuracy `0.965`.
      - Qwen-397B-NVFP4: gsm8k limit 200, accuracy `0.97`.
    - Multi-turn cache hit ratio (evalscope perf multi-turn)
      - Kimi-K2.5-NVFP4: 90.1%
      - Qwen-397B-NVFP4: 86.5%
      - DeepSeek v4 Flash:
    - Overlap scheduling regression: not observed.


tokenspeed-scheduler/csrc/scheduler/outside_event_handler.cpp
```diff
@@ -149,13 +148,9 @@ void Scheduler::handleEvent(const cache::WriteBackDone& event) {
     auto spec = std::move(it->second);
     cache_op_tracker_.erase(it);
 
-    auto now = std::chrono::steady_clock::now();
-    for (TreeNode* n : spec.nodes) n->Touch(now);
-
     if (!spec.request_id.empty()) {
         if (auto* req = find_request(spec.request_id)) {
```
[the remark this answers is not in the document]
SimonCqk: re-review this comment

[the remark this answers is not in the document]
SimonCqk: done

tokenspeed-scheduler/csrc/resource/hybrid_prefix_cache/mamba_family_ops.cpp
```diff
@@ -0,0 +1,368 @@
...
+        !local_mamba_allocator->AllocateCheckpoint(checkpoint_raw_position.value_or(-1))) {
+        return nullptr;
+    }
+    return local_mamba_allocator;
+}
+
+void HybridPrefixCache::PublishFinishMambaState(const std::vector<std::span<const std::int32_t>>& full_paged_tokens,
+                                                LocalMambaAllocator* local_mamba_allocator) {
+    if (!HasMambaAdjunct() || local_mamba_allocator == nullptr || !local_mamba_allocator->HasCheckpoint()) {
+        return;
```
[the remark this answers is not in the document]
SimonCqk: done

tokenspeed-scheduler/csrc/resource/hybrid_prefix_cache/mamba_family_ops.cpp
```diff
@@ -0,0 +1,368 @@
...
+                throw std::logic_error("ScheduleDecodeFromRetractedEvent: failed to allocate Mamba recovery slots");
+            }
+            return result;
+        case RequestLocalMambaKind::kNextCheckpoint:
+            if (!HasMambaAdjunct() || request.local_mamba_allocator == nullptr) return result;
+            if (!should_materialize_checkpoint()) {
+                (void)request.local_mamba_allocator->DetachCheckpoint();
+                return result;
+            }
+            (void)request.local_mamba_allocator->AllocateCheckpoint(request.checkpoint_raw_position.value_or(-1));
```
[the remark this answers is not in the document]
SimonCqk: done

tokenspeed-scheduler/csrc/scheduler/scheduler.h
```diff
@@ -136,8 +135,8 @@ class Scheduler {
     std::optional<MambaChunkAllocator> mamba_allocator_{};
     std::optional<MambaHostAllocator> mamba_host_allocator_{};
     KVPrefixCache kv_prefix_cache_;
+    HybridPrefixCache hybrid_prefix_cache_;
```
tuanzhangCS, line 138: Will kv_prefix_cache_ and hybrid_prefix_cache_ be merged into a single variable in the future?

tokenspeed-scheduler/csrc/fsm/forward_events.cpp
```diff
@@ -213,17 +147,22 @@ std::variant<PrefillDone, Prefilling> SchedulePrefillEvent::operator()(Prefillin
...
+            .local_mamba_allocator = local_mamba_allocator.get(),
+            .chunk_begin = state.window.begin,
+        },
+        cache::state::AcquireRequestLocalKV{
+            .allocator = *local_kv_allocator,
+            .tokens = tokens_this_round_,
+        },
+        cache::state::RefreshMambaCheckpoint{
+            .allocator = local_mamba_allocator.get(),
+            .checkpoint_raw_position = state.window.begin + state.window.size + tokens_this_round_,
```
tuanzhangCS, line 142: As a follow-up: the FSM still explicitly composes family-specific ops (e.g. RefreshMambaCheckpoint, AcquireRequestLocalKV) at each StepCommit call site. Would it be possible to encapsulate these per-phase op bundles inside HybridPrefixCache (e.g. a single StepCommit(Phase::PrefillChunk, context)) so the scheduler stays agnostic when new cache families are added?

tokenspeed-scheduler/csrc/scheduler/scheduler.h
SimonCqk: yes, this will be addressed in the next follow-up PR. The current PR is mainly intended to keep the LOC manageable and merge the mamba & ds v4 cache models first, and avoiding introducing the KV ownership rewriting in step one, and now `HybridPrefixCache` already holds a reference to `KVPrefixCache`, so the follow-up change should be a relatively small one.

tokenspeed-scheduler/csrc/fsm/forward_events.cpp
SimonCqk: can be further folded into HybridPrefixCache as well with manageable size, I've pushed a new commit with the implementation.

tuanzhangCS (approved):

tuanzhangCS (approved):

tokenspeed-scheduler/csrc/scheduler/scheduler.cpp
```diff
@@ -209,14 +264,15 @@ std::size_t Scheduler::RetractedSize() const {
 }
 
 std::size_t Scheduler::AvailableKvPages() const {
-    return device_allocator_.AvailablePages();
+    return hybrid_prefix_cache_.Stats().available_device_pages;
```
XucSh: Stats() always reserves and string-copies every paged_cache_group_ids and, when no paged_cache_group_ids filter is supplied, iterates every group filling three std::maps (hybrid_prefix_cache.cpp:435-450). AvailableKvPages() is called from NextExecutionPlan on every
  scheduling iteration just to read device_allocator_.AvailablePages().

tokenspeed-scheduler/csrc/resource/hybrid_prefix_cache/hybrid_prefix_cache_types.h
```diff
@@ -0,0 +1,340 @@
...
+    const std::string& request_id;
+    std::int32_t device_pages_needed{0};
+    std::int32_t first_raw_position_of_op{0};
+    std::int32_t target_raw_tokens_exclusive{0};
+    bool refresh_local_cache_state{false};
+
+    using Result = AdmissionVerdict;
+};
+
+struct DecodeFromRetracted {
```
XucSh, line 127: Every admit op holds request_id by reference. Today's call sites in forward.cpp work around it with a local const std::string request_id = request->Id(); (e.g. lines 90, 134, 167, 198), but Request::Id() returns by value, so the obvious-looking .request_id = 
  request->Id() in any future caller will bind the reference to a temporary and dangle.

  Suggest changing the field to std::string_view (cheapest, lifetime ties to caller's local) or std::string by value. Same applies to Stats({.request_id = ...}) — that one already uses std::optional<std::string> (line 324), so the convention is inconsistent within the
  same header.

tokenspeed-scheduler/csrc/resource/hybrid_prefix_cache/hybrid_prefix_cache.h
```diff
@@ -20,54 +20,152 @@
...
+        const cache::worker::CommitPrefillFirstChunkMetadata& op);
+    [[nodiscard]] cache::worker::CommitPrefillMetadata::Result Apply(const cache::worker::CommitPrefillMetadata& op);
+    [[nodiscard]] cache::worker::CommitDecodeAfterPrefillMetadata::Result Apply(
+        const cache::worker::CommitDecodeAfterPrefillMetadata& op);
+    [[nodiscard]] cache::worker::CommitDecodeMetadata::Result Apply(const cache::worker::CommitDecodeMetadata& op);
+    [[nodiscard]] cache::worker::CommitDecodeRecoveryMetadata::Result Apply(
+        const cache::worker::CommitDecodeRecoveryMetadata& op);
+
+    template <typename... Ops>
+    StepCommitResult StepCommit(Ops&&... ops) {
```
XucSh, line 94: A type is "step-committable" iff it has both an Apply(op) overload (no simulated_free) and an AccumulateStepResult(result, op_result) overload (lines 154-163). Admit ops have neither shape, so passing one yields a long template error instead of a readable diagnostic.
  The contract is invisible at the call site.

Suggest either:
  - a concept StepCommittable = requires { ... } and constraining Ops..., or
  - a static_assert((StepCommittable<std::decay_t<Ops>> && ...)) inside the function with a message.
  
Either makes "this op cannot be chained — call Apply directly" a compile-time message rather than an SFINAE wall.

tokenspeed-scheduler/csrc/scheduler/operations/forward.cpp
```diff
@@ -101,70 +80,36 @@ std::optional<fsm::SchedulePrefillFirstChunkEvent> Scheduler::schedulePrefillFir
...
         return {};
     }
-    if (needs_mamba_loadback) {
-        hybrid_prefix_cache_->PrepareMambaDeviceLoadBack(mamba_loadback_nodes);
-        TreeNode* mamba_node = hybrid_prefix_cache_->FindLastMambaNode(match_result.host.last_node);
-        if (mamba_node != nullptr) {
-            match_result.mamba_cow_src_index = mamba_node->MambaSlotIndex();
-        }
+    if (admission.mamba_branching_seqlen.has_value()) {
+        match_result.mamba_branching_seqlen = *admission.mamba_branching_seqlen;
```
XucSh: The typed-op interface returns Mamba state in AdmissionVerdict, but the scheduler then writes it back into match_result.mamba_* for the FSM event. So the state lives in two places and the scheduler still has to know Mamba exists. This is a half-step toward
  encapsulation: the registry was deliberately deferred (and that's fine), but having AdmissionVerdict and MatchResult both carry Mamba indices is a regression-friendly setup.

Suggest folding mamba_branching_seqlen / mamba_cow_src_index into the FSM event (so the scheduler never reads them) or, conversely, dropping them from AdmissionVerdict and letting Apply mutate match_result directly. Either way one field should be the source of truth.

tokenspeed-scheduler/csrc/resource/hybrid_prefix_cache/hybrid_prefix_cache.cpp
```diff
@@ -34,256 +39,267 @@
...
+
+    result.device_insert_page_count = std::max(0, alloc_count);
+    result.match_result = kv_prefix_cache_.Match(op.full_paged_tokens);
+    return result;
 }
 
-bool HybridPrefixCache::EnsureMambaCapacityByEvict(std::int32_t num_slots, TreeNode* protected_node) {
-    if (mamba_allocator_ == nullptr) return num_slots <= 0;
-    return mamba_eviction_manager_.EnsureCapacity(num_slots, protected_node);
+cache::publish::RetractPrefixPlan::Result HybridPrefixCache::Apply(const cache::publish::RetractPrefixPlan& op) {
```
XucSh: scheduleRetract (forward.cpp:236-249) calls Plan only to get device_insert_page_count so the caller can TakeFirstLocalKVPages(alloc_count) and pass them into Commit. Both ops re-walk DevicePagesFromRoot(current_device_node) and recompute alloc_count.

Suggest collapsing into a single op that takes LocalKVAllocator& (or RequestCacheMutation&) and does the take internally — the caller no longer needs to know alloc_count, and the prefix walk happens once.

tokenspeed-scheduler/csrc/scheduler/scheduler.cpp
```diff
@@ -44,22 +44,128 @@
...
+
 Scheduler::Scheduler(SchedulerConfig config)
     : config_{std::move(config)},
       device_allocator_{config_.page_size, config_.device_allocator.total_pages},
       host_allocator_{config_.page_size, config_.host_allocator.total_pages},
-      mamba_allocator_{},
+      mamba_allocator_{MakeMambaAllocator(config_)},
+      mamba_host_allocator_{MakeMambaHostAllocator(config_)},
       kv_prefix_cache_{&device_allocator_, &host_allocator_, config_.enable_l3_storage, config_.disable_prefix_cache},
+      hybrid_prefix_cache_{kv_prefix_cache_, device_allocator_, MambaAdjunctAllocator(mamba_allocator_, config_),
```
XucSh, line 167: Previously the facade was optional and only created when an adjunct/paged-cache feature was on. Now it's always constructed and unconditionally installs eviction callbacks (hybrid_prefix_cache.cpp:99-101), so a pure-KV config also pays a virtual hop on every KV evict
  (OnKVEvict checks mamba_allocator_ != nullptr then no-ops, and may call DetachPagedCacheSnapshotFromNode even when no group is registered).

Probably negligible at typical eviction rates, but worth a microbench on the no-Mamba/no-paged-cache config path before merging — the regression bench in the PR description doesn't seem to cover that profile.

tokenspeed-scheduler/csrc/scheduler/scheduler.cpp
```diff
@@ -345,73 +387,27 @@ ExecutionPlan Scheduler::NextExecutionPlan() {
...
-                ok = false;
-            }
-        }
-    }
-    for (auto& [p, cnt] : tree_device_pages) {
-        if (p <= 0 || p > total_device) {
-            spdlog::error("[check_mem] INVALID DEVICE PAGE id={} in radix tree (valid range [1,{}])", p, total_device);
-            ok = false;
-        }
+    auto stats_snapshot = hybrid_prefix_cache_.Stats({.include_device_memory_diagnostics = true});
```
XucSh, line 401: Stats() populates device_memory_diagnostics whenever request.include_device_memory_diagnostics is true (hybrid_prefix_cache.cpp:453-459), so this branch can never fire today. Either drop the check, or document the contract on Stats (e.g. an assert/[[nodiscard]] invariant) so the relationship is explicit instead of implicit.

tokenspeed-scheduler/csrc/scheduler/scheduler.cpp
SimonCqk: Fixed. Scheduler::AvailableKvPages() now goes through `HybridPrefixCache::AvailableDevicePages()`, which directly reads `device_allocator_.AvailablePages()` instead of materializing a full CacheStatsSnapshot. This avoids building paged-cache maps/vectors from the scheduler hot path.

tokenspeed-scheduler/csrc/resource/hybrid_prefix_cache/hybrid_prefix_cache_types.h
SimonCqk: Fixed. Admit op request IDs now use `std::string_view`, and `Request::Id()` returns `const std::string&` to avoid repeated string copies in scheduler paths. The string_view is only consumed synchronously inside Apply(...); it is not stored.

tokenspeed-scheduler/csrc/resource/hybrid_prefix_cache/hybrid_prefix_cache.h
SimonCqk: If an admit op is accidentally passed to `StepCommit`, the error now explicitly says that admit ops must call Apply(op, simulated_free). This keeps the typed op model zero-dispatch while making the API contract clearer.

tokenspeed-scheduler/csrc/scheduler/operations/forward.cpp
SimonCqk: Fixed. AdmissionVerdict now only carries the admission result and transfer pairs. Mamba-specific fields were removed from the verdict, and the relevant admit ops take a mutable MatchResult& so Apply(...) updates the final match result directly. Scheduler no longer reads Mamba fields from the verdict and writes them back separately.

tokenspeed-scheduler/csrc/resource/hybrid_prefix_cache/hybrid_prefix_cache.cpp
SimonCqk: `RetractPrefixPlan` was removed. `RetractPrefixCommit` now takes `RequestLocalCacheState&` and internally computes the missing prefix page count, calls `TakeFirstKVPages(...)`, inserts the pages, and returns the state-recovery `MatchResult`. Scheduler no longer knows alloc_count or performs the local page extraction itself.

tokenspeed-scheduler/csrc/scheduler/scheduler.cpp
SimonCqk: The unreachable optional-empty runtime branch in check_device_mem() was replaced by an invariant assert after requesting diagnostics explicitly. DEBUG_MEM behavior is preserved, but the contract is now clearer.

PR #245, opened 2026-05-25
ci(perf-qwen3.5-agentic): route to gb200-4gpu-perf runner

zhyncs:
Switches the qwen3.5 agentic perf task from the shared `gb200-4gpu` pool to the dedicated `gb200-4gpu-perf` pool so the multi-turn sweep doesn't contend with other gb200-4gpu jobs and runs on a node tuned for perf measurements.


zhyncs:
https://github.com/lightseekorg/tokenspeed/actions/runs/26382239666/job/77653660637?pr=245

minedec (approved):

PR #261, opened 2026-05-26
fix(deepseek): guard missing quant weight_block_size

mesaleh:
## Summary

- use `getattr(..., "weight_block_size", None)` before reading `quant_config.weight_block_size`
- keep the existing default `quant_block_size = 1` behavior when the quant config does not expose that attribute

## Motivation

Some quantization config implementations do not define `weight_block_size`. The current DeepSeek/Kimi loader path assumes the attribute exists before indexing it, which can fail during NVFP4 model loading.

## Validation

- `python3 -m compileall -q python/tokenspeed/runtime/models/deepseek_v3.py`
- Hit and validated locally while bringing up Kimi K2.6 NVFP4 under TokenSpeed

### DIFF
--- a/python/tokenspeed/runtime/models/deepseek_v3.py
+++ b/python/tokenspeed/runtime/models/deepseek_v3.py
@@ -1460,11 +1460,11 @@ class DeepseekV3ForCausalLM(BaseCausalLM):
                     "q_a_proj" in name or "kv_a_proj_with_mqa" in name
                 ):
                     quant_block_size = 1
-                    if (
-                        self.quant_config is not None
-                        and self.quant_config.weight_block_size is not None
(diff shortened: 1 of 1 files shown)


mesaleh:
I checked this after the stale auto-close. No need to reopen this PR now: the same loader guard is already present on current `main` via #444 (`feat(moe): add trtllm mxint4...`), merged 2026-06-14.

Current `python/tokenspeed/runtime/models/deepseek_v3.py` now uses `getattr(self.quant_config, "weight_block_size", None)` at this fused q/kv_a loader path, so the missing-attribute failure this PR targeted is covered upstream.

I am leaving this closed to avoid duplicating an already-merged fix.

PR #264, opened 2026-05-26
ci(perf): add 1m perf bench for qwen3.5

minedec:
## Summary
Add a 1M-context perf bench for Qwen3.5-397B-A17B-NVFP4 on TokenSpeed. Manual-trigger only on `b200-8gpu`.

- New CI task `qwen3.5-397b-a17b-nvfp4-evalscope-longctx`:
  - Drives the server with `evalscope perf` against four prompt lengths (128k / 256k / 512k / 1024k), each replayed for multiple runs, generating 1024 output tokens with greedy decoding.
  - Results are aggregated after the sweep, and surfaced via the GitHub step summary.
- New `test/long_context_benchmark/`:
  - `tokenspeed/collect_outputs.py`: walks the per-run / per-length output tree and produces a per-prompt-len table of `TPS/user`, `Avg AR`, and `Total Throughput`. 

## Test Plan
- Full multi-run B200 execution will be triggered through the manual workflow.


minedec:
> @minedec Let's use https://huggingface.co/datasets/lightseekorg/long-context/tree/main instead

ok

tuanzhangCS (approved):

PR #269, opened 2026-05-26
feat(engine): add pause_generation / continue_generation API

qywu:
## Summary
- Adds `POST /pause_generation` (modes: `abort`, `retract`, `in_place`) and `POST /continue_generation` endpoints, mirroring SGLang's RL training API
- `abort`: aborts all in-flight requests, then halts scheduling
- `in_place`: halts scheduling without touching running requests (KV cache stays valid; for in-place weight updates)
- `retract`: halts scheduling (full running→waiting retraction requires C++ scheduler `bulk_retract()` — tracked as TODO)
- Exposed as Python API on `Engine` and as HTTP routes on the PD mini load balancer

## Does it need scheduler modification?
`abort` and `in_place` modes are pure Python changes. `retract` mode needs a new C++ scheduler API (`bulk_retract_running()` or similar) to atomically move all running requests back to the waiting queue with their KV cache pages retained. That is left as a follow-up.

## Test plan
- [ ] `engine.pause_generation(mode="abort")` aborts all requests and blocks new scheduling
- [ ] `engine.continue_generation()` resumes scheduling
- [ ] `engine.pause_generation(mode="in_place")` halts new batches without aborting running ones
- [ ] PD mini-LB routes forward to both prefill and decode servers

### DIFF
--- a/python/tokenspeed/runtime/engine/event_loop.py
+++ b/python/tokenspeed/runtime/engine/event_loop.py
@@ -935,6 +935,8 @@ class EventLoop:
         while True:
             self._process_new_requests()
             self._commit_cache_results()
+            if self.request_handler.is_paused:
+                continue
             execution_plan = self.scheduler.next_execution_plan()
             self._publish_scheduler_kv_events()
             self._submit_cache_ops(execution_plan)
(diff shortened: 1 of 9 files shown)


qywu:
Superseded by cleaner branch — recreating to drop accidental FP8 MoE commits

PR #282, opened 2026-05-27
fix: MiniMax-M2.7 FP8 MoE compatibility on H100 (SM90)

qywu:
## Summary

Fixes several kernel-level incompatibilities that prevented MiniMax-M2.7 (FP8 block-quant, TP4) from running on H100 (SM90) with the triton FP8 MoE backend. Discovered and diagnosed via head-to-head benchmarking against tgl.

- **FP8 activation quantization crashes for M ≤ 32**: Both the TRT-LLM `fp8_quantize_1x128` kernel and the Triton `_per_token_group_quant_8bit` fallback fire async CUDA errors/illegal memory accesses for small decode/extend batches on SM90. Fixed with a pure-PyTorch path for M ≤ 32 (no performance impact since decode is memory-BW-bound).
- **FP8 scale normalization wrong for M%4==0**: `_normalize_fp8_group_scale_layout` misinterpreted the 1D scale buffer as [K, M] and called `.T.contiguous()` which triggered an async CUDA error. Fixed by detecting `m == aligned_m` and returning `view(m, K)` directly.
- **flashinfer silu_and_mul crashes on SM90**: The AOT-compiled flashinfer silu_and_mul kernel fails for certain batch sizes. Replaced with a pure-torch `F.silu + mul + copy_` equivalent.
- **TRTLLM FMHA extend kernel not compiled for SM90**: `trtllm_batch_context_with_kv_cache` uses TRT-LLM FMHA which has no SM90 CUBIN. Raise min arch to SM100+; the standard flashinfer paged-prefill kernel handles SM90. Also removes the unsupported `causal=` kwarg and adds `return_lse=False` support to the flashinfer extend kernel so it serves as the SM90 fallback.
- **bs=0 in CUDA graph capture list**: When `max_num_seqs=1`, `max_bs-1=0` was added to the capture list, causing crashes throughout the forward pass. Filtered to `0 < bs`.
- **OOM from flashinfer workspace fragmentation**: After CUDA graph capture fragments the memory pool, the 256 MB flashinfer prefill workspace allocation fails on first extend call. Fixed by setting `PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` in the engine subprocess.

## Test plan

- [ ] MiniMax-M2.7 serves decode requests at BS=1,4,8,16,32 on H100 without crashing
- [ ] No regression for MiniMax-M2.5 (NVFP4, B200) — triton FP8 path not used there
- [ ] No regression for DeepSeek V3, GPT-OSS — changes gated behind small-M path and SM90-specific kernels

## Follow-up (not in this PR)

- Add tuned triton MoE kernel config for H100 + MiniMax M2.7 FP8 (`E=256,inter_size=384,...,fp8_w8a8.json`)
- Fix `flashinfer_cutlass` SM90 CUBIN registration (`setup_cubin_loader()` missing after JIT compile)

### DIFF
--- a/python/tokenspeed/cli/_proc.py
+++ b/python/tokenspeed/cli/_proc.py
@@ -53,11 +53,17 @@ async def spawn_engine(
         str(port),
         *args,
     ]
+    # expandable_segments avoids OOM from memory fragmentation when a large lazy
+    # allocation (e.g., 256 MB flashinfer workspace) is requested after CUDA graphs
+    # have carved up the memory pool into non-contiguous chunks.
+    env = os.environ.copy()
+    env.setdefault("PYTORCH_CUDA_ALLOC_CONF", "expandable_segments:True")
     logger.info("spawn engine: %s", " ".join(cmd))
     return await asyncio.create_subprocess_exec(
         *cmd,
         stdout=asyncio.subprocess.PIPE,
         stderr=asyncio.subprocess.PIPE,
+        env=env,
     )
 
 
--- a/python/tokenspeed/cli/_proc.py
+++ b/python/tokenspeed/cli/_proc.py
@@ -53,17 +53,11 @@ async def spawn_engine(
         str(port),
         *args,
     ]
-    # expandable_segments avoids OOM from memory fragmentation when a large lazy
-    # allocation (e.g., 256 MB flashinfer workspace) is requested after CUDA graphs
-    # have carved up the memory pool into non-contiguous chunks.
-    env = os.environ.copy()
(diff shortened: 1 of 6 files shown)


qywu:
Closing — all changes were workarounds for a flashinfer-jit-cache version mismatch (0.6.9+cu130 vs flashinfer 0.6.11). The correct fix is upgrading to `flashinfer-jit-cache==0.6.11+cu130`. With the right jit-cache installed, pure main runs MiniMax-M2.7 on H100 (SM90) correctly across all batch sizes without any code changes. Benchmark results: BS1=107, BS4=395, BS16=1332, BS32=2237 tok/s (identical with and without the PR changes).

PR #286, opened 2026-05-28
fix(PD): fix PD speculative bootstrap input seeding

XucSh:
## Test Plan
Test with
1. Qwen3.5-397B-A17B-NVFP4 with MTP
2. Kimi-K2.5-NVFP4  with EAGLE3


python/tokenspeed/runtime/execution/input_buffer.py
```diff
@@ -168,6 +168,40 @@ def fill_input_buffers(
...
+            ).to(req_pool_indices_device.device, non_blocking=True)
+            mask = (decode_input_ids_tensor != -1).unsqueeze(1)
+            ids = decode_input_ids_tensor.unsqueeze(1)
+
+            # Only seed col 0 (the verified token). The draft slots
+            # (cols 1..spec_num_draft_tokens-1) are owned by either the local
+            # drafter or write_remote_spec_candidate_ids; for retract-recovery
+            # there is no draft source, so leaving the tail untouched lets the
+            # verifier reject those slots and the step naturally falls back to
+            # a 1-token decode. Broadcasting the seed across the tail would
```
tuanzhangCS: "naturally falls back to a 1-token decode" isn't strictly guaranteed. `future_input_map` is `torch.empty`-initialized (runtime_states.py:52) and never reset on slot reuse, so cols 1..N-1 hold stale ids from a previous request (or uninitialized memory). If the target happens to predict one of those ids, the verifier accepts a phantom draft — same failure shape as the bug this PR fixes, just lower probability.

The PD remote-prefill path is safe because `write_remote_spec_candidate_ids` fills real candidates. The retract-recovery path (`num_extends == 0` with `decode_input_ids` set) has no candidate source, so leaving the tail untouched is only probabilistically safe.

Consider seeding the tail with -1 (sentinel the verifier always rejects), or forcing this row through a 1-token path.

python/tokenspeed/runtime/execution/model_executor.py
```diff
@@ -914,6 +916,43 @@ def update_block_table(self, forward_op) -> ModelExecutionResult:
...
+        req_pool_tensor = torch.tensor(
+            [req_pool_idx for req_pool_idx, _ in pairs],
+            dtype=torch.int64,
+            device=self.device,
+        )
+        mamba_tensor = torch.tensor(
+            [mamba_idx for _, mamba_idx in pairs],
+            dtype=torch.int64,
+            device=self.device,
+        )
```
tuanzhangCS: Two synchronous H2D allocations in this PR:

1. **Here (L944-953)**: both `torch.tensor(..., device=self.device)` calls block. This function also runs on the default stream (no `execution_stream` context), so the stall hits the main loop directly.

2. **`runtime_states.py:85`** in `write_remote_spec_candidate_ids` — same pattern. The outer stream wrapper at `model_executor.py:1369` only affects kernel scheduling; `torch.tensor(..., device=cuda)` itself still blocks regardless of stream context.

Compare to `reset_valid_cache_length` (L994) and the new `write_decode_input_ids` helper, which both use the project-standard pattern:

```python
torch.tensor(..., dtype=..., device="cpu", pin_memory=True).to(self.device, non_blocking=True)
```

Suggest matching that pattern in both spots.

python/tokenspeed/runtime/execution/input_buffer.py
XucSh: need a refactor to fix this issue

tuanzhangCS:
LGTM, but need resolve conflicts before merge.

tuanzhangCS (approved):

PR #307, opened 2026-05-29
refactor: drop legacy attn backends

borontion:
## Summary

The triton and flash_attention backend (fa3) should be subsumed under the unified mha backend, which routes to triton/gluon/fa3/fa4/flashinfer kernels.


antiagainst (approved):
Nice reduction in code!

zhyncs (approved):
NB

PR #305, opened 2026-05-28
feat(entrypoints): add control-plane HTTP server sidecar (--control-port)

qywu:
## Summary

Adds an HTTP server sidecar that starts automatically alongside \`tokenspeed serve\` on \`main_port + 1\` (override with \`--control-port\`).

**Architecture:**
\`\`\`
Client  ──►  http_server  :8001
                ├─ /health, /get_server_info, /get_model_info,
                │  /health_check, /abort  ──►  gRPC engine (direct)
                └─ /generate, /v1/*, /flush_cache
                     ──►  smg gateway :8000 ──►  gRPC engine
\`\`\`

## Endpoints

| Method | Path | Backend |
|--------|------|---------|
| GET | \`/health\` | local (always 200) |
| GET | \`/get_server_info\` | gRPC direct |
| GET | \`/get_model_info\` | gRPC direct |
| GET | \`/health_check\` | gRPC direct |
| POST | \`/abort\` | gRPC direct |
| GET/POST | \`/generate\` | smg passthrough |
| POST | \`/v1/completions\` | smg passthrough (streaming supported) |
| POST | \`/v1/chat/completions\` | smg passthrough (streaming supported) |
| GET | \`/v1/models\` | smg passthrough |
| POST | \`/v1/messages\` | smg passthrough (Anthropic API) |
| POST | \`/v1/responses\` | smg passthrough (OpenAI Responses API) |
| POST | \`/flush_cache\` | smg passthrough |

## Usage

\`\`\`bash
# Auto-starts on port+1 (no flag needed)
tokenspeed serve --model <path> --port 8000
# → smg gateway on :8000, HTTP server on :8001

# Override control port
tokenspeed serve --model <path> --port 8000 --control-port 9000
\`\`\`

## Changes

- \`python/tokenspeed/runtime/entrypoints/http_server.py\` — FastAPI server with gRPC direct + smg passthrough
- \`python/tokenspeed/cli/_argsplit.py\` — \`--control-port\` orchestrator flag
- \`python/tokenspeed/cli/serve_smg.py\` — auto-start sidecar after smg ready, passing both \`gateway_url\` and \`engine_grpc_addr\`
- \`test/runtime/test_http_server.py\` — unit tests for \`/health\` and \`--control-port\` parsing

## Notes

- \`start_profile\` / \`stop_profile\` are not yet exposed (no gRPC method in the proto; sent over ZMQ internally)
- \`pause_generation\` / \`continue_generation\` will be added once PR #270 merges

PR #308, opened 2026-05-29
feat(entrypoints): add HTTP server sidecar alongside smg gateway

qywu:
## Summary

Adds an HTTP server that starts automatically with `tokenspeed serve` on `main_port + 1` (override with `--control-port`). It runs in a daemon thread alongside the smg gateway and gives engine control + generation a single HTTP surface.

Squashed replacement for #305 (single commit + comprehensive regression tests).

**Routing:**
```
Client ──► http_server :8001
             ├─ /get_server_info, /get_model_info, /health_check,
             │  /abort                        ──► gRPC engine (direct)
             └─ /health, /generate, /v1/completions, /v1/chat/completions,
                /v1/models, /v1/messages, /v1/responses, /flush_cache,
                /start_profile, /stop_profile ──► smg gateway ──► engine
```

## Usage

```bash
# Auto-starts on port+1, no flag needed
tokenspeed serve --model <path> --port 8000
# → smg gateway on :8000, HTTP sidecar on :8001

# Override the sidecar port
tokenspeed serve --model <path> --port 8000 --control-port 9000
```

## Endpoints

| Method | Path | Backend |
|--------|------|---------|
| GET | `/health` | smg passthrough (real engine health) |
| GET | `/get_server_info` | gRPC direct |
| GET | `/get_model_info` | gRPC direct |
| GET | `/health_check` | gRPC direct |
| POST | `/abort` | gRPC direct |
| GET/POST | `/generate` | smg passthrough |
| POST | `/v1/completions` | smg passthrough (streaming) |
| POST | `/v1/chat/completions` | smg passthrough (streaming) |
| GET | `/v1/models` | smg passthrough |
| POST | `/v1/messages` | smg passthrough (Anthropic API) |
| POST | `/v1/responses` | smg passthrough (OpenAI Responses API) |
| POST | `/flush_cache` | smg passthrough |
| GET/POST | `/start_profile`, `/stop_profile` | smg passthrough |

## Implementation notes

- **gRPC channel reuse** — direct calls share one lazily-created channel/stub; channels are expensive and must not be created per request (fd/socket leak).
- **gRPC error mapping** — engine RPC failures return a clean `503 {"error": "engine unavailable"}` instead of an unhandled 500.
- **Streaming session lifetime** — the upstream aiohttp session outlives the `StreamingResponse` and is closed in the stream generator's `finally`; closing it eagerly (e.g. `async with`) truncates the stream mid-flight.
- **Faithful non-streaming body** — relays the raw upstream bytes via `Response`; `JSONResponse(str)` would double-encode the already-serialized JSON.

## Tests

`test/runtime/test_http_server.py` — runs entirely on mocks (no engine/smg/GPU), 12 tests covering each of the above as regression guards:

- **Streaming** uses a deliberately *slow* mock (sleeps between chunks). The original fast mock buffered the whole body before the session closed and hid the bug — this one fails against the eager-close version and passes with the fix (verified by temporarily reverting).
- Non-streaming asserts the body is byte-faithful JSON (not a quoted string).
- gRPC stub identity (reuse), gRPC→503 mapping, plain-text `/health` passthrough, status-code relay, and `--control-port` parsing.

## Manual verification

Verified end-to-end against a live `tokenspeed serve` (Qwen2.5-0.5B):
- `:8001/health` → `OK` (proxied from smg)
- streaming `/v1/completions` → SSE tokens stream incrementally with `data: [DONE]`
- non-streaming `/v1/completions` → correct JSON, 200
- `/get_server_info`, `/get_model_info` (gRPC direct) → full JSON
- gRPC failure → clean 503


qywu:
Addressed in `b40dd34` (force-pushed after the reviewed commit `22a594c`): `_start_control_server` is now async and polls `uvicorn.Server.started` — it only reports `ts control server ready` once the socket is bound, and prints a WARNING (serving continues) if the port is in use or it times out. Added regression tests for both the ready-after-bind and port-in-use paths.

@codex review

qywu:
Addressed the streaming-cutoff P2: replaced `ClientTimeout(total=600)` with `ClientTimeout(total=None, sock_connect=30, sock_read=600)`. Dropping the wall-clock `total` cap means a long but actively-streaming generation is never aborted mid-flight; `sock_read` still bounds inactivity so a genuinely hung upstream is detected. Added a regression test asserting `total is None` with the connect/read bounds set.

python/tokenspeed/cli/serve_smg.py
```diff
@@ -393,6 +433,28 @@ async def run_smg(
...
+                f"ts control server ready on http://{user_host}:{control_port}\n"
+            )
+        else:
+            sys.stderr.write(
+                f"WARNING: ts control server failed to bind on "
+                f"http://{user_host}:{control_port} (port in use?); "
+                f"serving continues without it\n"
+            )
+        sys.stdout.flush()
+        sys.stderr.flush()
```
HJSang: are those two flush necessary?

python/tokenspeed/cli/serve_smg.py
```diff
@@ -281,6 +281,46 @@ def _gateway_args_with_defaults(gateway_args: list[str]) -> list[str]:
...
+        engine_grpc_addr=engine_grpc_addr,
+        host=host,
+        port=port,
+    )
+    thread = threading.Thread(target=server.run, daemon=True, name="ts-http-server")
+    thread.start()
+
+    # uvicorn sets `started = True` only after the socket is bound and serving.
+    loop = asyncio.get_running_loop()
+    deadline = loop.time() + timeout
```
HJSang: Add a logging for hong long until the server is ready?

python/tokenspeed/cli/serve_smg.py
qywu: stdout one yes — it's block-buffered when piped/redirected (normal deploy), and we block on asyncio.wait() right after, so without it the readiness line never hits the logs. Mirrors the `ts serve ready` flush above. stderr flush is redundant (stderr is line-buffered) — dropping it.

HJSang (approved):
LGTM with minor comments.

python/tokenspeed/cli/serve_smg.py
qywu: Good call — added `logger.info("control server bound in %.2fs", ...)` on the success path. Gives us signal when the bind is slow or creeping toward the 30s timeout (port contention / loaded host).

qywu:
> LGTM with minor comments.

thanks! I have addressed the comments and will merge after checks are passed

PR #314, opened 2026-05-29
perf: add Gluon MoE kernels for GPT-OSS

knwng:
Added support for Gluon MoE kernels and applied optimizations including:
- sliceMN for prefill shapes
- plain pipelining for decode shapes for now
- weight preshuffling on host


python/tokenspeed/runtime/layers/moe/backends/mxfp4/triton_kernel.py
borontion, line 1: 1. I am thinking of whether we can wrap routing + group gemms into a single `moe_fused` call, basically put multiple kernels together. that's how flashinfer moe works in my understanding, internally it also calls multiple kernels. that way we can avoid the intermediate branches - like support quant or not.
2. Should we consider to rename this `Mxfp4TritonKernelBackend` to something else, given it is using gluon kernels. another reason is I think we should try to hide the kernel detail specially triton_kernerls detail from the runtime.

what do you think?

knwng: 1. Do you mean fusing permute, activation into gemm1, and combine into gemm2? If so, we've already done that. If you mean merging the entire routing into gemms or merging everything, I didn't find the implementation in flashinfer so far. Can you give me a pointer? We can discuss it in follow-up anyway.
2. Make sense. We should have a gluon backend or something.

borontion: > Do you mean fusing permute, activation into gemm1, and combine into gemm2? If so, we've already done that. If you mean merging the entire routing into gemms or merging everything, I didn't find the implementation in flashinfer so far. Can you give me a pointer? We can discuss it in follow-up anyway.

no. it is not about kernel fusion. we can still keep the current implementation. I mean we can register a gluon version of `moe_fused`, and inside it call multiple kernels. 

```python
@register_kernel(...)
def gluon_moe_fused(...):
  # call route kernel
  # call group gemm1 kernel + fused activation
  # call group gemm2 kernel + combine
```

based on our original design, this should be `ExecutionPlan`. but given we don't have it now.

knwng: oh I see what you mean. SGTM.

borontion:
@knwng do you have any perf number vs. the triton kernel? it is fine if perf is lower than triton_kernels - we can lower its priority and do not auto-select it for now.

knwng:
> @knwng do you have any perf number vs. the triton kernel? it is fine if perf is lower than triton_kernels - we can lower its priority and do not auto-select it for now.

It's largely close. But to be safe I set it as secondary.

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/gluon.py
```diff
@@ -0,0 +1,4794 @@
...
+    capability=CapabilityRequirement(
+        vendors=frozenset({"amd"}),
+        min_arch_version=ArchVersion(9, 5),
+        max_arch_version=ArchVersion(9, 5),
+    ),
+    priority=_kernel_priority(),
+    tags={"throughput", "latency"},
+)
+
+register_kernel(
```
antiagainst: This is duplicated with the following.

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/gluon.py
```diff
@@ -0,0 +1,4794 @@
...
+        max_arch_version=ArchVersion(9, 5),
+    ),
+    signatures=_GLUON_FUSED_SIGNATURES,
+    traits={"activation_dtype": frozenset({"fp8"})},
+    priority=_kernel_priority(),
+    tags={"throughput", "latency"},
+)(_gluon_mxfp_fused_moe)
+
+
+__all__ = [
```
antiagainst: Move this to the beginng to be more discoverable?

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/gluon.py
```diff
@@ -0,0 +1,4794 @@
...
+
+register_kernel(
+    "moe",
+    "experts",
+    name="gluon_dispatch_gemm",
+    features={"ragged_metadata", "dispatch_gemm"},
+    **_common,
+)(_gluon_mxfp_ragged_matmul)
+
+register_kernel(
```
antiagainst: We need `@register_kernel`?

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/gluon.py
```diff
@@ -0,0 +1,4794 @@
...
+)
+from tokenspeed_kernel.platform import ArchVersion, CapabilityRequirement
+from tokenspeed_kernel.registry import Priority, register_kernel
+from tokenspeed_kernel.signature import (
+    ScaleFormat,
+    dense_tensor_format,
+    format_signature,
+    tensor_format,
+)
+from tokenspeed_triton.experimental import gluon
```
antiagainst: Some of these imports are duplicated from L28. For the ones missing, let's put inside `_triton.py` and then re-import here for single central place.

antiagainst (approved):
Nice! Quite clean baseline. Overall lgtm; just a few nits.

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/gluon.py
knwng: These are for different features, i.e. dispatch gemm and combine gemm

antiagainst: Oh yeah makes sense.

antiagainst (approved):

PR #318, opened 2026-05-30
fix(eagle): avoid mutating drafter sequence lengths

XucSh:

### DIFF
--- a/python/tokenspeed/runtime/execution/drafter/eagle.py
+++ b/python/tokenspeed/runtime/execution/drafter/eagle.py
@@ -222,14 +222,6 @@ class Eagle(BaseDrafter):
 
         draft_first_step_reduce = forward_mode.is_decode()
 
-        if draft_first_step_reduce and self.attn_backend.support_kv_cache_prewrite:
-            # Trim seq_lens by rejected-draft count so the sliced decode
-            # query does not attend to dead positions.
-            correction = (self.spec_num_tokens - draft_input.accept_lengths).to(
-                self.draft_seq_lens_buf.dtype
-            )
-            self.draft_seq_lens_buf[:bs].sub_(correction)
-
         ctx = ForwardContext(
             attn_backend=self.attn_backend,
             token_to_kv_pool=self.token_to_kv_pool,


LorrinWWW (approved):

rjzhb:
Thank you for the fix. I overlooked this case before. The correction was originally added for Llama Eagle3's decode-kernel direct dispatch optimization (Q-slice + decode kernel on the prewrite-decode catch-up step): since the decode kernel has no causal mask, the seq_lens correction is required there to prevent the catch-up query from attending to the previous multi-step's rejected positions. Qwen3.5 NextN and DeepSeek NextN don't use this optimization, but they currently enter the same branch, which is where the bug originates.  I think this PR change alone can currently fix Qwen nextN, but it may cause minor errors in Llama Eagle3.

LorrinWWW:
Maybe a tentative walkaround is to remove this code block (what this PR did), and also change llama_eagle to post-attn-trim to make sure there is no correctness issue in main. And we can bring it back when we have a clean solution to localize the trim. Wdyt? @XucSh @rjzhb

rjzhb:
> Maybe a tentative walkaround is to remove this code block (what this PR did), and also change llama_eagle to post-attn-trim to make sure there is no correctness issue in main. And we can bring it back when we have a clean solution to localize the trim. Wdyt? @XucSh @rjzhb

Yeah, i think it does make sense.  

Then can delete the decode-kernel direct dispatch branch in `python/tokenspeed/runtime/models/llama_eagle3.py:189-212`, keeping only the `else` (standard PagedAttention call):

```python
if ctx.draft_first_step_reduce:
    # KV already written via fused_set_kv_buffer_arg above; slice Q
    # to one query per request and route attn as decode.
    q_rope = q_rope.index_select(0, ctx.gather_ids)
    attn_output = ctx.attn_backend.forward(
        q_rope,
        None,
        None,
        self.attn,
        out_cache_loc,
        ctx.token_to_kv_pool,
        ForwardMode.DECODE,
        ctx.bs,
        save_kv_cache=False,
    )
else:
    attn_output = self.attn(
        q_rope,
        None,
        None,
        save_kv_cache=False,
        ctx=ctx,
        out_cache_loc=out_cache_loc,
    )
```
image

And move this part to outside of if-else for post attn trim.

XucSh:
> Maybe a tentative walkaround is to remove this code block (what this PR did), and also change llama_eagle to post-attn-trim to make sure there is no correctness issue in main. And we can bring it back when we have a clean solution to localize the trim. Wdyt? @XucSh @rjzhb

Investigating the root cause, let's hold this PR for now.

LorrinWWW:
@XucSh will do a quick fix in a separate PR. And then work on refactoring.

LorrinWWW:
#335 as a temporary fix

PR #319, opened 2026-05-30
deps(kernel): try testpypi triton 3.7.10.post20260530

antiagainst:


antiagainst:
Looks there are some correctness regressions. Need to bisect..

zhyncs (approved):
Looks good.

Let’s bump the version on the production PyPI.

By the way, could we make the test index optional and disable it by default? We can keep it for debugging purposes.

antiagainst:
This is meant for testing. I'm gonna push this to pypi next. So won't land this one.

zhyncs:
> This is meant for testing. I'm gonna push this to pypi next. So won't land this one.

Cool. I think we can keep this PR since we’ll need TestPyPI for testing next time as well. We can disable it by default and enable it via an environment variable when needed. wdyt

antiagainst:
Yup. Next time bumping triton I'll resurrect it for testing.

PR #326, opened 2026-05-31
fix(chore): sampling vocab bounds for padded model vocab

XucSh:


LorrinWWW:
Do we still need it? I feel we probably still need something to avoid generating OOB token @XucSh 
(like if the model somehow generates an OOB token, it shouldn't crash)

PR #335, opened 2026-06-01
Fix(spec decode): catch up trim bug

rjzhb:
## Summary

fix #217 that broke spec-decode catch-up on Qwen NextN and DeepSeek V3 NextN:
- catch-up `seq_lens` trim in `drafter/eagle.py`
- Llama Eagle3 pre-attention q-slice (dispatch B) in `llama_eagle3.py`

The trim assumed q was sliced to 1 row before the kernel. Llama Eagle3 had the matching q-slice; Qwen NextN and DeepSeek V3 NextN didn't — trim alone left `q_len > cache_seqlens`, giving negative kernel positions, all-masked softmax, NaN, and `accept_rate=0`. 

Solution is to add a temp flag: pre_attention_trim. 

NOTICE: this flag would be removed in the future refactor PR once adding pre_attention_trim feature to both Qwen NextN and DeepSeek V3 NextN.

## Test Plan

- Qwen3.5-35B-A3B-NVFP4 + MTP local serve: no NaN in catch-up attention output, `accept_rate > 0` restored.
- Llama Eagle3 (MiniMax-M2.5-NVFP4 target + thoughtworks/MiniMax-M2.5-Eagle3 draft): serves without regression.

### DIFF
--- a/python/tokenspeed/runtime/execution/drafter/eagle.py
+++ b/python/tokenspeed/runtime/execution/drafter/eagle.py
@@ -222,14 +222,6 @@ class Eagle(BaseDrafter):
 
         draft_first_step_reduce = forward_mode.is_decode()
 
-        if draft_first_step_reduce and self.attn_backend.support_kv_cache_prewrite:
-            # Trim seq_lens by rejected-draft count so the sliced decode
-            # query does not attend to dead positions.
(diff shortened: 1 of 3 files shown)


LorrinWWW (approved):
This is a correctness fix for spec. And we will have a follow-up refactoring to reduce / remove the extra if-else branching.

PR #337, opened 2026-06-02
feat(logprobs): output token logprobs in vLLM and SGLang formats

HJSang:
## Summary

Adds **output (generated-token) logprobs in two request dialects — vLLM and SGLang — served from one compute path**, chosen per request. The scheduler/sampler emit format-neutral wire arrays; the frontend splits into two pure renderers and dispatches on the request's dialect. Scheduler/execution/PD untouched.

## API

| Dialect | Request | Response (`meta_info`) |
|---|---|---|
| **vLLM** | `SamplingParams.logprobs` (`None`=off, `0`=sampled token) | `logprobs`: `list[dict[token_id, Logprob]]` + `cumulative_logprob` |
| **SGLang** | `GenerateReqInput.return_logprob` (+ `top_logprobs_num` / `logprob_start_len` / `token_ids_logprob` / `return_text_in_logprobs`) | `output_token_logprobs` (+ top-k / token-id variants): `(logprob, token_id, text\|None)` tuple-lists |

`GenerateReqInput.logprob_format` overrides the dialect (`vllm` \| `sglang` \| `both`); default = **match the request** (vLLM if `sampling_params.logprobs` is set, else SGLang). The engine entrypoints expose both surfaces.

**One compute path.** `LogprobsProcessor` splits into `render_vllm` / `render_sglang`, both reading the unchanged neutral wire arrays. `collector._APPEND_META_KEYS` is the union of both dialects' list keys (only one is present per request). `Logprob` carries `logprob` + `rank` (slot index, 0 = sampled — not the full-vocab rank) and a `to_dict()` for serving boundaries.

**Capabilities gated loudly, not silently clamped.** Output top-k (`logprobs>0` / `top_logprobs_num>0`), full-vocab (`-1`), and prompt logprobs (`logprob_start_len>=0` / `token_ids_logprob`) raise in both dialects. Output logprobs are gated by the static `enable_output_logprobs` server arg: requesting them when it's off raises a clear error (no silent empty result).

## Speculative decoding

Output logprobs work under speculative decoding (the old guard that nulled logprob requests under a spec algorithm is removed; the sampler `verify()` path gathers correct next-token logprobs). Validated below.

## Validation (nv2, B300)

Each cell issues **both dialects × 7 prompts** (factual / code / math / multilingual / science / narrative / symbol). "Cross-format" = vLLM and SGLang report **byte-identical** token ids + logprob values from the same run (the core dual-format guarantee). "HF Δ" = mean |engine − HF| (teacher-forced on the engine's tokens) where transformers can serve as a reference.

| Model | Arch | Quant | Backend | Cross-format | HF mean Δ | Result |
|---|---|---|---|---|---|---|
| Qwen2-0.5B | dense | bf16 | mha | ✅ identical | 0.012 | ✅ |
| Qwen2-0.5B | dense | fp8 KV | trtllm | ✅ identical | 0.104 | ✅ |
| Qwen3-30B-A3B | MoE | bf16 | mha | ✅ identical | 0.028 | ✅ |
| gpt-oss-20b | MoE | MXFP4 | flashinfer_trtllm | ✅ identical | 0.027 | ✅ |
| Qwen3.5-35B-A3B | MoE | NVFP4 | trtllm + flashinfer_trtllm | ✅ identical | n/a¹ | ✅ |

¹ NVFP4 isn't loadable in transformers, so no HF reference; validated via cross-format identity + a finite/≤0 sanity check.

**Speculative decoding (MTP):** DeepSeek-V4-Pro, greedy, MTP-on vs MTP-off (lossless → tokens + logprobs must match): 16/16 tokens identical, mean Δ 0.082 nats (fp8 verify-vs-decode noise), PASS at fp8 tolerance.

Coverage: dense + MoE; bf16 / fp8-KV / MXFP4 / NVFP4; speculative decoding; both dialects. Both renderers are also unit-tested over `format ∈ {vllm, sglang}`, and the CI runtime suites (`ut-runtime-1gpu`, `ut-runtime-prefix-cache-e2e`) are green.

## Scope

No C++ scheduler, execution-layer, or PD changes — format is a render-time choice over the neutral wire arrays. Implements the dual-format design from the review thread.


HJSang:
### Codex review triage

- **Prompt-only requests return empty** & **prompt top-k hardcoded empty** — already fixed by the Phase B / top-k routing commits; `_maybe_set_input_logprob_ctx` sets `extend_return_logprob` independently and prompt top-k is now routed. Validated on B200 (`prompt_logprobs=0` and `=5` match HF).
- **`logprobs`/`prompt_logprobs = -1` (full vocab)** — was accepted by `verify()` but nothing materializes a vocab-sized result. Now **rejected** loudly until implemented (`04b1532`).
- **OUTPUT top-k (`logprobs=N>0`) / output `logprob_token_ids`** — **deferred** (`5e00c8c`, documented at the hardcoded site). Output logprobs flow through the captured CUDA-graph decode path, so top-k must be computed in the sampler and captured into the graph output buffers — a separate milestone. Until then `logprobs=N>0` returns the sampled token's logprob (correct value, no alternatives).
- **Mixed prefill/decode batches skip prompt logprobs** — known pure-extend limitation (`num_extends==bs`); the input-path per-request slicing assumes no interleaved decode rows. Documented follow-up.

Validated end-to-end on nv2 B200 vs HF `log_softmax`: output `logprobs=0` (2 models), prompt `prompt_logprobs=0` (vLLM convention, mean Δ≈0.008–0.05), prompt top-k `prompt_logprobs=5`, prompt `logprob_token_ids` (24/24 ids match), and prefix-cache correctness (`SkipRead`).

qywu:
**`prompt_logprobs` is only correct on the single-chunk path — worth making the unsupported cases loud (or scoping them out).**

Verified locally: `prompt_logprobs` works for a single-chunk, pure-extend prefill (56-token prompt → 56 positions, `[0]=None`, values line up). But `model_executor._maybe_set_input_logprob_ctx` activates *only* for pure-extend batches and hardcodes `start_lens=[0]` with no cross-chunk offset — so a prompt that gets chunked (`prompt_tokens > chunked_prefill_size`, default 8192) or lands in a mixed extend+decode batch silently falls outside the validated path. `SkipRead` handles prefix-cache reuse but not chunking. The PR already lists this under follow-ups; the concern is it's currently *silent* rather than *loud*.

Two ways to make that safe:

1. **Keep `prompt_logprobs`, add a guard** — raise `NotImplementedError` (or reject in `LogprobParams.verify()`) when `prompt_tokens > chunked_prefill_size` / mixed batch / `start_len > 0`, until the chunked path lands. Preserves the common-case feature + vLLM parity, makes the edges safe.
2. **Scope this PR to output logprobs only** — reject `prompt_logprobs` in `verify()` with a clear "not supported yet" message and defer the prompt path (incl. the `SkipRead` recompute) to a follow-up. Smaller, fully-correct surface.

I'd lean (1), since the single-chunk path is already validated and `prompt_logprobs` matters for vLLM parity — but (2) is reasonable if it isn't needed yet. Either beats silently-wrong logprobs on long/chunked prompts.

_Caveat: I verified single-chunk directly; I couldn't run a clean chunked repro because chunked prefill crashes independently on my older test base (unrelated to this PR), so I can't say whether the real-base chunked behavior is silently-wrong vs. an error._

python/tokenspeed/runtime/engine/logprob_params.py
```diff
@@ -0,0 +1,107 @@
...
+
+    @property
+    def requested(self) -> bool:
+        return (
+            self.logprobs is not None
+            or self.prompt_logprobs is not None
+            or bool(self.logprob_token_ids)
+        )
+
+    def verify(self, vocab_size: int, max_logprobs: int) -> None:
```
qywu: If you decide to scope this PR to **output logprobs only** for now (prompt logprobs is the hard part — correct only for single-chunk, pure-extend prefill, with no guard for chunked/mixed paths), the cleanest place to enforce it is right here in `verify()`: reject `prompt_logprobs` and `logprob_token_ids` (the latter only feeds the prompt path) loudly, instead of silently returning wrong/partial logprobs on long or chunked prompts.

```suggestion
    def verify(self, vocab_size: int, max_logprobs: int) -> None:
        # Output-only scope for now: the prompt-logprob path is validated only
        # for a single-chunk, pure-extend prefill (start_len=0, no cross-chunk
        # offset). Reject prompt_logprobs / logprob_token_ids loudly rather than
        # silently returning wrong/partial logprobs on chunked or mixed batches.
        if self.prompt_logprobs is not None:
            raise ValueError(
                "prompt_logprobs is not supported yet; only output logprobs "
                "are available."
            )
        if self.logprob_token_ids:
            raise ValueError(
                "logprob_token_ids is not supported yet; only output logprobs "
                "are available."
            )
```

Notes:
- This makes the loop's `prompt_logprobs` branch and the `logprob_token_ids` length/vocab checks below unreachable — fine to leave, or simplify the loop to just `("logprobs", self.logprobs)`.
- Back-compat: a legacy caller that sets `logprob_start_len >= 0` (coerced to `prompt_logprobs`) will now get a clear error instead of partial results — intended for the scope-down.
- Separate but adjacent: output top-k (`logprobs > 0`) is also deferred (only the sampled token is produced today), so you may want to clamp/reject `logprobs > 0` here too — left out to keep this suggestion focused on the prompt-vs-output split.

qywu:
there is merge conflict for python/tokenspeed/runtime/layers/logits_processor.py

python/tokenspeed/runtime/entrypoints/engine.py
```diff
@@ -155,10 +156,7 @@ def generate(
         # - List of images (one per request in a batch)
         # - List of lists of images (multiple images per request)
         # See also python/tokenspeed/runtime/utils/common.py:load_image for more details.
-        return_logprob: list[bool] | bool | None = False,
-        logprob_start_len: list[int] | int | None = None,
-        top_logprobs_num: list[int] | int | None = None,
-        token_ids_logprob: list[list[int]] | list[int] | None = None,
+        logprob_params: list[LogprobParams] | LogprobParams | None = None,
```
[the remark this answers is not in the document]
HJSang: Still accepted — `generate()`/`async_generate()` keep the `return_logprob`/`logprob_start_len`/`top_logprobs_num`/`token_ids_logprob` kwargs and translate them via `GenerateReqInput`; they aren't rejected. Separately, the spec-decode guard that *did* null them has been removed in 035f19e5.

python/tokenspeed/runtime/execution/model_executor.py
```diff
@@ -512,7 +512,60 @@ def _forward_step(
...
+            return
+        prompt_lp = [int(x) for x in forward_op.prompt_logprobs[:num_extends]]
+        if not any(p >= 0 for p in prompt_lp):
+            return
+        seq_lens = [
+            int(x)
+            for x in self.input_buffers.extend_seq_lens_cpu[:num_extends].tolist()
+        ]
+        start_lens = [0 for _ in seq_lens]
+        pruned = [el - sl for sl, el in zip(start_lens, seq_lens)]
```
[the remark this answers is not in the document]
HJSang: Prompt logprobs are now scoped out (rejected at `LogprobParams.verify()`), so this pure-extend input-scoring path is inert. The per-row `start_len` handling will be revisited when the prompt path is re-enabled (035f19e5).

python/tokenspeed/runtime/engine/generation_output_processor.py
```diff
@@ -538,12 +599,62 @@ def post_process_forward_op(
...
+            # every i < num_extends (the ctx activates input logprobs for the
+            # whole pure-extend batch), so advance ilp_pt unconditionally here —
+            # before the rid_to_state / prefill-finished guards — to keep the
+            # pointer aligned with the flat array even for delayed/chunked rows.
+            seg_val = None
+            seg_idx = None
+            pl_req = -1
+            if input_logprobs_list is not None and i < num_extends:
+                plen = int(forward_op.input_lengths[i])
+                pl_req = int(forward_op.prompt_logprobs[i])
```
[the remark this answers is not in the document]
HJSang: Moot for now — the prompt-logprob path is scoped out (rejected at `verify()`); cached-prefix padding returns with the prompt path (035f19e5).

python/tokenspeed/runtime/engine/io_struct.py
```diff
@@ -111,7 +112,38 @@ class GenerateReqInput:
...
 
+    def _coerce_legacy_logprob_fields(self):
+        """Translate the deprecated scalar logprob request fields into
+        ``logprob_params`` when the caller did not set the new field directly.
+        Keeps the old request API working (e.g. the SMG gRPC servicer)."""
+        if self.logprob_params is not None:
+            return
+        rl = self.return_logprob
+        if isinstance(rl, list):
+            rl = bool(rl[0]) if rl else False
```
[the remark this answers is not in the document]
HJSang: Fixed in 035f19e5 — `_coerce_legacy_logprob_fields` now builds a per-row `list[LogprobParams|None]` for batched legacy inputs (e.g. `return_logprob=[False, True]`) instead of reading element 0.

python/tokenspeed/runtime/engine/io_struct.py
```diff
@@ -111,7 +115,54 @@ class GenerateReqInput:
...
+        if isinstance(tln, list):
+            tln = tln[0] if tln else 0
+        tln = int(tln or 0)
+        lsl = self.logprob_start_len
+        if isinstance(lsl, list):
+            lsl = lsl[0] if lsl else -1
+        want_prompt = isinstance(lsl, int) and lsl >= 0
+        self.logprob_params = LogprobParams(
+            logprobs=tln if rl else None,
+            prompt_logprobs=tln if (rl and want_prompt) else None,
```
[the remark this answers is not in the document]
HJSang: Prompt logprobs are scoped out: a legacy `return_logprob=True` + `logprob_start_len>=0` now maps to `prompt_logprobs`, which `verify()` rejects loudly (no longer silently scored from 0). The offset will be honored again when the prompt path lands (035f19e5).

python/tokenspeed/runtime/engine/collector.py
```diff
@@ -25,13 +25,14 @@
 from collections.abc import Sequence
 from typing import Any
 
+# Logprob outputs: each is a list[dict[int, Logprob]] that EXTENDS
+# across streamed frames (handled by _extend_sequence, which is dict-safe since
+# _is_prefix compares elements with ==). ``cumulative_logprob`` is intentionally
+# NOT here: it is a scalar that the default overwrite path replaces with the
+# latest frame's value.
```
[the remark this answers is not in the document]
HJSang: Fixed in 035f19e5 — `cumulative_logprob` is now summed across coalesced streamed frames (`_SUM_META_KEYS`), so it stays consistent with the appended per-position `logprobs`.

python/tokenspeed/runtime/engine/logprobs.py
```diff
@@ -56,118 +103,121 @@ def __init__(self, engine: AsyncLLM) -> None:
...
+        ``logprob_token_ids`` entries (``tid_*[p]``) are folded in without
+        clobbering already-present (ranked) entries.
+        """
+        out: list[dict[int, Logprob]] = []
+        for p in range(len(sampled_idx)):
+            tvals = top_val[p] if p < len(top_val) and top_val[p] else []
+            tids = top_idx[p] if p < len(top_idx) and top_idx[p] else []
+            vals = [sampled_val[p]] + list(tvals)
+            ids = [sampled_idx[p]] + list(tids)
+            decoded = self._maybe_decode(ids, return_text)
```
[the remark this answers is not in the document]
HJSang: Moot — the prompt-logprob path (incl. the `return_text` decode of the `-1`-padded trailing prompt target) is scoped out at `verify()`; handled when the prompt path is re-enabled (035f19e5).

python/tokenspeed/runtime/engine/logprob_params.py
HJSang: Done — went with option (2): `verify()` now rejects `prompt_logprobs` and `logprob_token_ids` (plus output top-k `logprobs>0` and full-vocab `-1`), scoping the PR to output `logprobs=0`. Thanks for the suggestion! (035f19e5)

python/tokenspeed/runtime/engine/io_struct.py
```diff
@@ -111,7 +115,54 @@ class GenerateReqInput:
...
+        tln = self.top_logprobs_num
+        if isinstance(tln, list):
+            tln = tln[0] if tln else 0
+        tln = int(tln or 0)
+        lsl = self.logprob_start_len
+        if isinstance(lsl, list):
+            lsl = lsl[0] if lsl else -1
+        want_prompt = isinstance(lsl, int) and lsl >= 0
+        self.logprob_params = LogprobParams(
+            logprobs=tln if rl else None,
```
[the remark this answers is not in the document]
HJSang: Fixed in 035f19e5 — legacy `return_logprob=True, top_logprobs_num=N>0` is now clamped to the sampled-token logprob (`logprobs=0`) instead of raising; output top-k remains unsupported, but legacy callers no longer error.

HJSang:
> there is merge conflict for python/tokenspeed/runtime/layers/logits_processor.py

resolved

qywu (changes requested):
**Requesting changes — keep this PR scoped to the functional surface (output logprobs).**

The entire `tokenspeed-scheduler/` C++ diff exists only to support the prompt-logprob path, which `LogprobParams.verify()` rejects in this PR. On the output-only path `request->PromptLogprobs()` is always `-1`, so `MatchIntent::SkipRead` never fires and the new `prompt_logprobs` / `logprob_token_ids` fields just carry `-1`/empty — it's dead code until the follow-up.

Could we drop the scheduler changes here and land them with the prompt-logprob follow-up, where they're actually exercised and testable? Output logprobs are computed entirely Python-side (`model_executor` / sampler `verify()` / output processor), so they need no scheduler change.

If removed, the coupled Python plumbing has to go too so nothing references the removed pybind fields:
- `scheduler_utils.py::make_spec` — the `prompt_logprobs` / `logprob_token_ids` params + `spec.*` assignments
- `request_handler.py` — the `lp.num_prompt_logprobs()` / `logprob_token_ids` -> `make_spec(...)` wiring
- the C++ side: `python_module.cpp`, `request_spec.h`, `request.{h,cpp}`, `forward.{h,cpp}`, `types.h`, and the prefix-cache `SkipRead` handling

(Keeping them is defensible too — harmless no-op, saves a recompile later. But if the goal is a minimal, fully-exercised PR, splitting the scheduler work into the follow-up matches each diff to what it enables.)

python/tokenspeed/runtime/engine/collector.py
```diff
@@ -25,13 +25,29 @@
...
+# *summed* (not overwritten) to stay consistent with the appended list.
 _APPEND_META_KEYS = {
-    "input_token_logprobs",
-    "output_token_logprobs",
-    "input_top_logprobs",
-    "output_top_logprobs",
-    "input_token_ids_logprobs",
-    "output_token_ids_logprobs",
+    "logprobs",
+    "prompt_logprobs",
```
qywu: `prompt_logprobs` is the only prompt-specific entry in this merge policy — `logprobs` and `cumulative_logprob` are output-side and stay. Drop `prompt_logprobs` from `_APPEND_META_KEYS` when the prompt path is removed (the engine never emits a `prompt_logprobs` meta key once `verify()` rejects it).

python/tokenspeed/runtime/engine/generation_output_processor.py
```diff
@@ -74,6 +74,7 @@ def __init__(
         token_ids_logprob: list[int] | None = None,
         multimodal_inputs=None,
         prompt_input_ids_unpadded: list[int] | None = None,
+        return_prompt_logprob: bool = False,
```
qywu: `return_prompt_logprob` and every accumulator it gates (`input_token_logprobs_val/idx`, `input_top_logprobs_val/idx`, `input_token_ids_logprobs_val/idx`, `sent_prompt_logprob_offset` — lines 112–135) are prompt-path only. Remove them, plus the `return_prompt_logprob` derivation/passing in `from_recv_req` (~189, ~201).

python/tokenspeed/runtime/engine/generation_output_processor.py
```diff
@@ -534,6 +570,31 @@ def post_process_forward_op(
             if model_execution_results.output_logprobs is not None
             else None
         )
+        _input_lp = getattr(model_execution_results, "input_token_logprobs", None)
```
qywu: This whole block through ~line 670 is prompt-path only: it reads `ModelExecutionResult.input_*` fields and `forward_op.prompt_logprobs` (both removed with the scheduler change), the `ilp_pt` accumulator, and the per-request `seg_val`/`seg_idx`/`pl_req` accumulation into `request_state.input_*`. Remove.

python/tokenspeed/runtime/engine/generation_output_processor.py
```diff
@@ -817,6 +934,49 @@ def stream_output(
                 output_token_logprobs_val.append([])
                 output_token_logprobs_idx.append([])
 
+            # Prompt/input-token logprobs: ship the un-shipped tail of the
```
qywu: The prompt-logprob shipping block (937–978) and its wiring into `BatchStrOut` (999–1015) should revert to the prior empties: `input_token_logprobs_val=[]`, `input_top_logprobs_val=[]`, `input_token_ids_logprobs_val=[]` (idx likewise). The `output_*` fields and the output-top-k TODO (1005–1011) stay.

python/tokenspeed/runtime/engine/logprobs.py
```diff
@@ -56,118 +103,121 @@ def __init__(self, engine: AsyncLLM) -> None:
...
-                    _get("output_top_logprobs_val"),
-                    _get("output_top_logprobs_idx"),
-                    return_text_in_logprobs,
-                )
+            logprobs_info.setdefault("logprobs", []).extend(positions)
+            logprobs_info["cumulative_logprob"] = logprobs_info.get(
+                "cumulative_logprob", 0.0
+            ) + self._sum_slot0(out_sampled_val)
+
+        if num_prompt is not None:
```
qywu: `if num_prompt is not None:` … ~167 builds the `prompt_logprobs` meta entry — prompt path only; remove. The output block (`num_out`), `_build_positions`, `build_position_logprobs`, and the `Logprob` dataclass stay. Note: with prompt **and** output-token-id logprobs both deferred, `want_token_ids` is always False, so the `tid_*` args + the fold-in loop in `_build_positions` (196–211) are dead for this PR too.

python/tokenspeed/runtime/engine/logprob_params.py
```diff
@@ -0,0 +1,120 @@
...
+    # top-N logprobs per OUTPUT (sampled) token; None = off, 0 = chosen only.
+    # N>0 (top-N + chosen) and -1 (full vocab) are reserved but NOT yet
+    # supported: the sampler/output path only materializes the sampled token's
+    # logprob, so verify() rejects them rather than silently returning a single
+    # token. Only None / 0 are currently honored.
+    logprobs: int | None = None
+    # top-N logprobs per PROMPT token; same value semantics as ``logprobs``.
+    # SCOPED OUT for now: verify() rejects it (only the single-chunk pure-extend
+    # prompt path is validated; chunked/mixed/prefix-cache paths are unsafe).
+    prompt_logprobs: int | None = None
```
qywu: If prompt support is dropped from this PR, consider removing the `prompt_logprobs` and `logprob_token_ids` fields, `num_prompt_logprobs()`, and their `verify()` reject branches (88–97) outright rather than plumbing a field end-to-end that only ever raises. Alternatively keep them purely as an explicit 'not supported' guard — but then the downstream readers (request_handler, scheduler_utils, generation_output_processor, the prompt branch in logprobs.py) shouldn't exist. Pick one; today it's fully wired but always rejected.

python/tokenspeed/runtime/engine/request_handler.py
```diff
@@ -206,9 +206,18 @@ def handle_generate_request(
         if recv_req.bootstrap_port is None:
             recv_req.bootstrap_port = self.server_args.disaggregation_bootstrap_port
 
+        lp = getattr(recv_req, "logprob_params", None)
```
qywu: This `num_prompt_logprobs()` / `logprob_token_ids` → `make_spec(prompt_logprobs=…, logprob_token_ids=…)` wiring (209–220) feeds the scheduler fields flagged in my other review. Remove together with the `make_spec` params and the C++ change — `make_spec(rid=…, tokens=…)` is all output logprobs need.

python/tokenspeed/runtime/engine/scheduler_utils.py
```diff
@@ -44,10 +44,21 @@
 _TRUTHY_ENV_VALUES = {"1", "true", "yes", "on"}
 
 
-def make_spec(rid: str, tokens: list[int]) -> RequestSpec:
+def make_spec(
```
qywu: `make_spec`'s new `prompt_logprobs` / `logprob_token_ids` params + the `spec.*` assignments (56–61) are prompt-path only and write the scheduler fields flagged in my other review. Revert `make_spec` to `(rid, tokens)`.

python/tokenspeed/runtime/execution/context.py
```diff
@@ -69,3 +69,14 @@ class ForwardContext:
     draft_seq_lens_buf: torch.Tensor | None = None
     # accept_lengths: per-request accepted verify width for cache_seqlens correction.
     accept_lengths: torch.Tensor | None = None
+
+    # --- input/prompt logprobs (off-policy); set by _maybe_set_input_logprob_ctx ---
```
qywu: All nine fields here (74–82) are the input/prompt-logprob context, set only by `_maybe_set_input_logprob_ctx` in model_executor. Nothing on the output path reads them. Remove for an output-only PR.

python/tokenspeed/runtime/execution/types.py
```diff
@@ -57,6 +57,21 @@ class ModelExecutionResult:
     output_logprobs: torch.Tensor | None = None
     # Optional next-round input rows captured for PD prefill data-plane handoff.
     next_input_ids: torch.Tensor | None = None
+    # Flat per-scored-position prompt/input-token logprobs (pure-extend batches
+    # that requested prompt_logprobs), concatenated across requests in
+    # forward-op order. None when no request asked for prompt logprobs.
+    input_token_logprobs: torch.Tensor | None = None
```
qywu: These six `input_*` fields (63–74) on `ModelExecutionResult` are prompt-path only — nothing on the output path reads them (`output_logprobs` is the only logprob field output needs). Remove.

qywu (commented):
Following up on the request-changes review with the specific prompt-logprob code to remove so this PR is scoped to **output logprobs** only. The output-logprobs API refactor stays (the `LogprobParams`/`Logprob` types, `logprob_params` plumbing in io_struct/input_processor/output_processor/entrypoints, the `num_out` path in logprobs.py, and the output-logprob tests). Each inline note marks code that exists solely for the prompt/input-logprob path — currently unreachable because `LogprobParams.verify()` rejects `prompt_logprobs`/`logprob_token_ids`, so it's untested dead code until the follow-up.

Grouped: scheduler-coupled (`request_handler.py`, `scheduler_utils.py`) · execution path (`context.py`, `types.py`, `cuda_graph_wrapper.py`, `model_executor.py`) · output-processor accumulation/shipping (`generation_output_processor.py`, `collector.py`) · assembly (`logprobs.py` prompt branch) · API/guard (`logprob_params.py`, `io_struct.py` legacy mapping) · tests (`runners.py`).

python/tokenspeed/runtime/execution/cuda_graph_wrapper.py
```diff
@@ -907,7 +909,10 @@ def __call__(
             with nvtx_range("graph_replay", color="red"):
                 self.graphs[padded_bs].replay()
 
-            output_tokens, output_lengths, output_logprobs = self.output_buffers[
+            # Captured output buffers carry _forward_step's 4-tuple; the 4th
+            # entry (input/prompt logprobs) is only produced on the eager
+            # pure-extend path, never in a captured decode graph, so discard it.
+            output_tokens, output_lengths, output_logprobs, _ = self.output_buffers[
```
qywu: The 3-tuple→4-tuple change here (plus the `None` pad at 930 and the docstring example at 188) exists only to carry the eager-path `input_token_logprobs`. With the prompt path removed `_forward_step` returns the original 3-tuple and all of this reverts.

python/tokenspeed/runtime/execution/model_executor.py
```diff
@@ -651,7 +651,60 @@ def _forward_step(
             ] = next_round_input_ids.to(torch.int32)
 
         output_logprobs = logits_output.next_token_logprobs
-        return output_tokens, accept_lengths, output_logprobs
+        input_token_logprobs = getattr(logits_output, "input_token_logprobs", None)
```
qywu: `_forward_step`'s new 4th return element — the `(input_token_logprobs, input_top_*, input_tid_*)` bundle (654–669) — is prompt path only. Revert to the 3-tuple `(output_tokens, accept_lengths, output_logprobs)`.

python/tokenspeed/runtime/execution/model_executor.py
```diff
@@ -651,7 +651,60 @@ def _forward_step(
...
+            (
+                input_token_logprobs,
+                input_top_val,
+                input_top_idx,
+                input_tid_val,
+                input_tid_idx,
+            ),
+        )
+
+    def _maybe_set_input_logprob_ctx(self, ctx, forward_op, bs, num_extends):
```
qywu: `_maybe_set_input_logprob_ctx` (672–707) is entirely the prompt-logprob activation path: it reads `forward_op.prompt_logprobs` / `logprob_token_ids` and sets the `ctx.extend_*` fields. Remove the method and its call site at ~1474.

python/tokenspeed/runtime/execution/model_executor.py
```diff
@@ -1509,6 +1573,20 @@ def execute_forward_op(
                         ),
                         **mamba_kwargs,
                     )
+                    if _input_lp_bundle is None:
```
qywu: The `_input_lp_bundle` unpack (1547–1551, 1576–1589), the `input_*=None` inits (1416–1420), the CPU copy (1646–1656), and the `ModelExecutionResult(input_*=…)` kwargs (1665–1670) are all prompt path. Remove with the 4-tuple revert above.

python/tokenspeed/runtime/layers/logits_processor.py
```diff
@@ -292,6 +292,10 @@ def forward(
         aux_hidden_states: torch.Tensor | None = None,
     ) -> LogitsProcessorOutput:
         # Get the last hidden states and last logits for the next token prediction
+        # NOTE: ``extend_return_logprob`` is only ever set True once the
```
qywu: Comment-only change. With the prompt path removed, `extend_return_logprob` is always False so this branch is always taken — the added 'Phase B' NOTE can be dropped (or reduced to a one-liner without the prompt-path framing).

test/runners.py
```diff
@@ -569,73 +572,50 @@ def forward_generation_raw(
...
-                input_token_logprobs + [output_token_logprobs[0]]
-            )
-            output_token_logprobs_lst.append(output_token_logprobs)
+            # New logprob format: meta_info["prompt_logprobs"] is a
+            # list[dict[int, Logprob] | None] (entry 0 is None for the first
+            # prompt token); meta_info["logprobs"] is the per-output-token list.
+            # OUTPUT top-k logprobs are deferred (not produced yet), so only the
+            # prompt (prefill) top-k and prompt token-id logprobs are extracted;
+            # output top-k is left empty and skipped by check_close_model_outputs.
+            prompt_lp = response["meta_info"].get("prompt_logprobs") or []
```
qywu: Heads-up: once the prompt path is gone, `meta_info` has no `prompt_logprobs` key, so `prompt_positions` is always `[]`. That makes the prefill-logprob comparison in `check_close_model_outputs` (704–708, `n_common=0`) and the prompt-token-id comparison (750–781) vacuous — they pass trivially. Better to drop the prompt/prefill comparisons and keep this runner asserting only the output (decode, `logprobs=0`) path, so the test actually exercises what the PR ships.

python/tokenspeed/runtime/engine/io_struct.py
```diff
@@ -111,7 +115,110 @@ class GenerateReqInput:
...
+        rejected, keeping legacy callers working. ``logprob_start_len >= 0``
+        still maps to ``prompt_logprobs`` (rejected by ``LogprobParams.verify``
+        until the prompt path lands)."""
+
+        def _row(rl, lsl, tid) -> LogprobParams | None:
+            rl = bool(rl)
+            tid = list(tid) if tid else None
+            if not (rl or tid):
+                return None
+            want_prompt = isinstance(lsl, int) and lsl >= 0
```
qywu: For an output-only PR the legacy coercion only needs `return_logprob → logprobs=0`. `want_prompt` + `prompt_logprobs=…` (145, 151) and `logprob_token_ids=tid` (152) map to features `verify()` rejects, so a legacy caller passing `logprob_start_len>=0` or `token_ids_logprob` would now hard-error. Either trim these mappings, or keep them and accept the reject — but call out the back-compat behavior change.

python/tokenspeed/runtime/engine/logprob_params.py
```diff
@@ -0,0 +1,120 @@
...
+TokenSpeed keeps ``SamplingParams`` focused on *how to sample*; logprob return
+config is about *what to report*, so it lives here. It exposes
+``logprobs`` / ``prompt_logprobs`` / ``logprob_token_ids``.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+MAX_LOGPROB_TOKEN_IDS = 128
```
qywu: Dead constant — `MAX_LOGPROB_TOKEN_IDS` is defined here and referenced nowhere on the branch (`git grep` finds only this line). Either wire it into `verify()` (e.g. cap `len(logprob_token_ids)`) or drop it.

python/tokenspeed/runtime/engine/logprobs.py
```diff
@@ -32,22 +32,69 @@
...
+        decoded_token: detokenized string, or None when text not requested.
+    """
+
+    logprob: float
+    rank: int | None = None
+    decoded_token: str | None = None
+
+
+# One position maps token_id -> Logprob.
+LogprobsOnePosition = dict[int, "Logprob"]
```
qywu: Dead type alias — `LogprobsOnePosition` is defined and referenced nowhere (`git grep` finds only this line). `_build_positions`/`build_position_logprobs` annotate with the inline `dict[int, Logprob]` instead. Drop it or actually use it as the return annotation.

python/tokenspeed/runtime/execution/context.py
```diff
@@ -69,3 +69,14 @@ class ForwardContext:
...
+    extend_return_logprob: bool = False
```
qywu: Beyond being prompt-only: 8 of these 9 fields are **write-only**. `_maybe_set_input_logprob_ctx` assigns `ctx.extend_return_logprob`, `extend_logprob_start_lens_cpu`, `extend_seq_lens_cpu`, `extend_logprob_pruned_lens_cpu`, `top_logprobs_nums`, `token_ids_logprobs`, `extend_return_top_logprob`, `extend_token_ids_logprob` — but nothing ever reads them back off `ctx` (only `extend_input_logprob_token_ids_gpu` is read, in model_executor ~1651). The logits kernel reads `LogitsMetadata.*`, and the PR adds no `ctx → LogitsMetadata` bridge, so these don't even drive the existing prompt kernel. Dead regardless of the prompt-path decision.

python/tokenspeed/runtime/engine/logprobs.py
```diff
@@ -56,118 +103,121 @@ def __init__(self, engine: AsyncLLM) -> None:
...
-                    return_text_in_logprobs,
-                )
-            )
-            output_top_logprobs.extend(
-                self.detokenize_top_logprobs_tokens(
-                    _get("output_top_logprobs_val"),
-                    _get("output_top_logprobs_idx"),
-                    return_text_in_logprobs,
-                )
+            logprobs_info.setdefault("logprobs", []).extend(positions)
```
qywu: **SGLang compat (breaking).** This replaces the entire SGLang `meta_info` logprob surface — `input_token_logprobs`, `output_token_logprobs`, `input_top_logprobs`, `output_top_logprobs`, `input_token_ids_logprobs`, `output_token_ids_logprobs` (each a `list[(logprob, token_id, decoded_token)]`) — with `meta_info["logprobs"]` / `["prompt_logprobs"]` as `list[dict[int, Logprob]]` (+ `cumulative_logprob`). Any SGLang-shaped consumer (the external `smg_grpc_servicer` OpenAI path, SGLang router/eval harnesses, the PD `mini_lb` merge of `input_token_logprobs`) reading the old keys silently gets nothing. The in-repo tests were rewritten to the new shape, so CI won't catch this. If SGLang wire-compat is a requirement, consider emitting both key sets during a deprecation window; if it's an intentional break, please call it out in the PR description as an API break.

python/tokenspeed/runtime/engine/logprobs.py
```diff
@@ -32,22 +32,69 @@
...
-from tokenspeed.runtime.engine.io_struct import BatchStrOut
+from tokenspeed.runtime.engine.logprob_params import LogprobParams
 
 if TYPE_CHECKING:
     from tokenspeed.runtime.engine.async_llm import AsyncLLM
+    from tokenspeed.runtime.engine.io_struct import BatchStrOut
+
+
+@dataclass
+class Logprob:
```
qywu, line 48: **SGLang compat (serialization).** SGLang returned plain tuples/lists, which JSON-serialize directly. `dict[int, Logprob]` does not: `Logprob` is a bare dataclass (no `to_dict`/`__json__`), and JSON object keys must be strings (int keys get coerced, changing the shape consumers see). No in-repo path serializes it (the only `asdict`/`json.dumps` hits are server_args + PD registry/transfer, unrelated). The OpenAI/HTTP mapping lives in the external smg package, which the PR notes "still needs the matching change" — so served logprobs are effectively unmapped for this shape until that lands. Worth a `to_dict()` / explicit wire form here so in-process and served outputs agree.

python/tokenspeed/runtime/engine/io_struct.py
```diff
@@ -111,7 +115,110 @@ class GenerateReqInput:
...
+            return LogprobParams(
+                # Output top-k unsupported -> honor return_logprob as the
+                # sampled-token logprob; top_logprobs_num is intentionally
+                # clamped to 0 rather than rejected for back-compat.
+                logprobs=0 if rl else None,
```
qywu: **SGLang compat (silent semantic changes).** This legacy translation changes behavior for existing SGLang clients:
- `top_logprobs_num` is **dropped entirely** — it isn't even a parameter of `_row`. A request with `return_logprob=True, top_logprobs_num=5` silently returns only the sampled token (`logprobs=0`), no top-k, no error.
- `logprob_start_len >= 0` (the standard SGLang way to ask for prompt logprobs) → `prompt_logprobs` → `verify()` **raises**. Previously-working requests now hard-error.
- `token_ids_logprob` → `logprob_token_ids` → `verify()` **raises** (even when `return_logprob=False`, since `_row` proceeds on `tid` alone).
- `return_logprob=True` used to return both prompt and output logprobs; now it's output-only.

`return_text_in_logprobs` is the one preserved. Suggest at minimum erroring (not silently clamping) on `top_logprobs_num > 0` so the data loss is visible, and documenting the prompt-field rejections as a back-compat break.

qywu (commented):
Two follow-ups in one review: (1) the net-new redundancy items from the offline pass, and (2) a careful **SGLang logprobs compatibility** check (this codebase is SGLang-derived).

**Redundancy:** two dead symbols (`MAX_LOGPROB_TOKEN_IDS`, `LogprobsOnePosition`) and the write-only `ForwardContext` fields.

**SGLang compatibility — three concerns**, all verified against `main` and the PR branch:
1. The public `meta_info` logprob keys are renamed + reshaped (six SGLang tuple-lists → `logprobs`/`prompt_logprobs` dict-of-`Logprob`). Breaking for any SGLang-shaped consumer; CI won't catch it since the tests were rewritten.
2. `Logprob` (dataclass) + int-keyed dicts aren't JSON/wire-native the way SGLang's lists were; no in-repo serializer handles them.
3. Legacy request-field coercion silently changes meaning: `top_logprobs_num` dropped, `logprob_start_len`/`token_ids_logprob` now hard-error.

None of these block the output-logprobs goal — but if SGLang wire-compat matters (router, eval harnesses, the external OpenAI/smg path), they need a conscious decision (compat shim vs. documented break).

qywu:
## Design proposal: support SGLang **and** vLLM logprob formats together

Following up on the compatibility review. The good news is this PR is already most of the way there — both formats are just **two views of the same per-position data**, and the neutral wire arrays this PR keeps unchanged (`*_token_logprobs_val/idx`, `*_top_logprobs_val/idx`, `*_token_ids_logprobs_val/idx`) are already a superset of what either dialect needs. So we can serve both with **one compute path, two renderers, chosen per request** — no scheduler involvement, no double work.

### Architecture

```
scheduler/model ──► neutral wire arrays (BatchStrOut: *_val / *_idx)   [UNCHANGED]
                              │
                    convert_logprob_style(params, recv_obj)            ◄── single dispatch point
                       │                         │
                 render_sglang()           render_vllm()
                       │                         │
   meta_info["output_token_logprobs"]    meta_info["logprobs"]
   meta_info["output_top_logprobs"]      meta_info["prompt_logprobs"]
   ... (6 tuple-list keys)               ... (dict[int, Logprob] + cumulative_logprob)
```

`render_sglang` is essentially the pre-PR `convert_logprob_style` (tuple-lists); `render_vllm` is the new code this PR already wrote. Both read the same neutral arrays.

### Request side — a near 1:1 mapping (PR is close)

The PR's `_coerce_legacy_logprob_fields` just needs to be lossless instead of dropping fields:

| SGLang request | vLLM `LogprobParams` | Note |
|---|---|---|
| `return_logprob=True, top_logprobs_num=0` | `logprobs=0` | sampled token only |
| `return_logprob=True, top_logprobs_num=N` | `logprobs=N` | **currently dropped** — should map N |
| `logprob_start_len>=0` (+ `top_logprobs_num`) | `prompt_logprobs=N` | `start_len` is the extra SGLang knob (caveat 2) |
| `token_ids_logprob=[...]` | `logprob_token_ids=[...]` | direct |
| `return_text_in_logprobs` | `return_text` | already preserved |

Gate unsupported *capabilities* (output top-k `N>0`, prompt logprobs) as capabilities — not by silently clamping the *API shape*.

### Output side — add a `format` flag, dispatch on it

```python
@dataclass
class LogprobParams:
    logprobs: int | None = None
    prompt_logprobs: int | None = None
    logprob_token_ids: list[int] | None = None
    return_text: bool = False
    format: Literal["vllm", "sglang", "both"] = "vllm"   # new
```

**Selection policy: the response matches the request dialect.**
- Built via legacy fields → `format="sglang"` → tuple-list keys (back-compat preserved exactly).
- Built via `logprob_params` directly → `format="vllm"` (default) → dict[int, Logprob].
- `"both"` available for migration/testing (opt-in; it doubles payload).

This gives full SGLang back-compat with zero ambiguity, computed once. `collector._APPEND_META_KEYS` becomes the **union** of both dialects' list keys (only one dialect is present per request, so the union is harmless).

### Where each change lands

| Change | File |
|---|---|
| `format` field; keep `render_sglang` (old code) + `render_vllm` (new); dispatch in `convert_logprob_style` | `engine/logprobs.py` |
| Set `format="sglang"` in legacy coercion; carry `top_logprobs_num` into the count | `engine/io_struct.py` |
| `_APPEND_META_KEYS` = union of both key sets | `engine/collector.py` |
| `Logprob.to_dict()` so the vLLM shape is JSON-safe; serving layer calls it | `engine/logprobs.py` + serving boundary |
| Parametrize logprob tests over `format ∈ {sglang, vllm}` | `test/runners.py`, `test/runtime/...` |

### Caveats to decide consciously

1. **`rank` fidelity (vLLM):** vLLM's `Logprob.rank` is the token's 1-based rank in the *full vocab distribution*; the PR uses slot index (0=sampled, 1..N). True parity needs the actual vocab rank from the sampler. Orthogonal to this architecture, but don't advertise vLLM-rank parity until it's real.
2. **`logprob_start_len`:** SGLang allows mid-prompt start; vLLM is whole-prompt. `render_sglang` honors `start_len` (slice), `render_vllm` returns the whole prompt. Doesn't bite now (prompt logprobs deferred) — just bake `start_len` into the neutral representation so it's there when the prompt path lands.
3. **OpenAI/serving is a *third* renderer.** Point the external smg servicer at the **neutral arrays**, not at either `meta_info` dialect, so it never depends on `format`.
4. **`"both"` cost:** always-emit-both doubles the logprob payload (heavy for RL/prompt-logprob workloads) — keep it opt-in, not the default.

### Net

Keep the neutral wire arrays as the contract, split `convert_logprob_style` into two pure renderers, add a `format` flag defaulting to "match the request dialect," and make `Logprob` serializable. One compute path, both APIs satisfied, scheduler untouched. This also closes the CI blind spot from the review (tests would then assert both shapes).

HJSang:
Thanks @qywu — adopted this. The branch now serves **both dialects from one compute path**, dispatched per request, as proposed.

One deliberate adaptation: I kept the vLLM API on `SamplingParams.logprobs` (per the earlier maintainer direction) rather than reintroducing `LogprobParams`. The `format` flag lives on `GenerateReqInput.logprob_format` (`vllm`|`sglang`|`both`, default = **match the request dialect**: vLLM if `sampling_params.logprobs` is set, else SGLang). Everything else follows your design:

- **Two renderers, one compute path.** `LogprobsProcessor` splits into `render_vllm` (dict + `cumulative_logprob`) and `render_sglang` (tuple-list keys); both read the unchanged neutral wire arrays. Scheduler untouched.
- **Request side.** SGLang fields restored on `GenerateReqInput` + the engine entrypoints; vLLM via `sampling_params.logprobs`.
- **`_APPEND_META_KEYS` = union** of both dialects' list keys.
- **`Logprob.to_dict()`** added for the serving boundary.
- **Tests parametrized** over `format ∈ {vllm, sglang}`.

On your caveats:
1. **`rank`** is slot-index (0 = sampled), not the full-vocab rank — not advertised as vLLM-rank parity; noted in the dataclass docstring.
2. **`logprob_start_len`** (mid-prompt) is deferred along with prompt logprobs.
3. **Serving** should read the neutral arrays, not a `meta_info` dialect — agreed; left as the contract.
4. **`"both"`** is opt-in (doubles payload), not the default.

Per your "gate *capabilities*, not the API shape": output top-k (`logprobs>0` / `top_logprobs_num>0`), full-vocab (`-1`), and prompt logprobs (`logprob_start_len>=0` / `token_ids_logprob`) now **raise loudly** in both dialects instead of silently clamping. Validated on B300 (vLLM vs HF on dense+MoE/bf16+fp8-KV; spec-decode MTP lossless; both renderers unit-tested).

qywu (approved):

PR #346, opened 2026-06-03
feat(engine): pause/resume scheduler control API

HJSang:
## Summary

Adds a **pause / resume / is-paused** scheduler control API for RLHF weight-update workflows. The gate lives in the **Python event loop** (requests are admitted to the scheduler from Python via `submit_requests`, so withholding new work while paused is handled there) and is driven by `PauseState{UNPAUSED, PAUSED_NEW, PAUSED_ALL}`.

**Modes** (how in-flight requests are treated):
- `abort` — cancel in-flight requests, drain, then reply
- `wait` — let in-flight requests finish, drain, then reply
- `keep` — freeze everything in place; reply immediately; resume later

## Design

Accept-but-defer: pausing gates the *drain* (scheduling / forward), never the *fill* (`submit_requests`), so resume is a flag flip plus a flush of buffered specs.

| Layer | Change |
|---|---|
| `io_struct.py` | `PauseMode` + Pause/Resume/IsPaused message pairs |
| `pause.py` (new) | `PauseController` + `PauseState` — the gate core |
| `event_loop.py` | buffer admitted specs while paused, skip forward on `PAUSED_ALL`, idle-drain for deferred `abort`/`wait` replies, flush buffered specs on resume |
| `request_handler.py` | control-request dispatch |
| `scheduler_control_client.py` / `engine.py` / `protocol.py` | `AsyncLLM.pause_scheduler` / `resume_scheduler` / `is_scheduler_paused` + `Engine` passthroughs |
| `generation_output_processor.py` | `abort_notify_client` flag so pause-aborts stream a terminating finish to passive clients; client-initiated aborts unchanged |

**No `clear_cache`:** pause/resume is purely scheduling control. Cache invalidation after a weight swap is already owned by the weight-update op (`update_weights_*(flush_cache=...)`), so pause does not duplicate it.

## Scope

HTTP endpoints (`http_server.py`) and the smg gRPC `Pause/Resume/IsPaused` RPCs are a **follow-up PR** — the cross-repo contract is `AsyncLLM`, which this PR implements and validates.

## Testing

- **Unit (CPU):** `test_pause_controller.py` (8 tests: keep/wait/abort/resume/is_paused, deferred-drain reply, invalid mode, FIFO buffering) + a `mark_abort` notify-client case in `test_generation_output_processor.py`.
- **GPU E2E (nv2, 8×B200, Qwen2-1.5B):** through the real `AsyncLLM` → ZMQ → event-loop → scheduler stack — `keep` freezes generation (0 tokens in a 5 s pause) and resumes to completion; `wait` returns only after the in-flight request drains; `abort` cancels the in-flight request (~4 ms drain) and terminates the client stream; `is_paused` tracks state.

Two bugs found via GPU testing and fixed: (1) resume stranded buffered specs because `_process_new_requests` early-returns before the admission gate; (2) pause-aborts left passive clients hanging (the `abort_notify_client` fix).


HJSang:
Addressed the automated review feedback (pushed in `2b8b1ab`):

- **P2 — Complete pending pause replies before resume.** `PauseController.handle_resume` now sends the in-flight `wait`/`abort` reply before clearing it, so a caller awaiting `pause_scheduler` is unblocked even if `resume` arrives mid-drain (resume uses a separate communicator and otherwise couldn't wake it). New unit test `test_resume_resolves_inflight_pause_then_acks`.

- **P1 — Abort grammar-queued requests during pause** and **P1 — Include grammar-queued requests in wait-mode drain.** Both shared one root: requests still compiling in `GrammarManager.grammar_queue` are not yet in `rid_to_state` or the scheduler, so the abort sweep and the drain check missed them. `abort`/`wait` now set a `cancel_grammar` flag; the event loop marks those queued states finished-with-abort, so the next `get_ready_grammar_requests` pass publishes them (terminating client finish) instead of admitting — they never run under post-resume weights or strand the drain. `keep` leaves them frozen (resumed later), as intended.

- **P2 — Skip aborted buffered specs on resume.** The resume flush now filters buffered specs whose registered state is already finished (`_spec_already_finished`), so a request the client aborted while paused no longer burns a scheduler slot + a forward pass after resume.

- **P1 — Implement clear_cache before acknowledging drained pauses.** Obsolete: `clear_cache` was removed from the pause API entirely. Cache invalidation after a weight swap is owned by the weight-update op (`update_weights_*(flush_cache=...)`); pause is purely scheduling control.

Unit tests: 9 pass. Re-validated on nv2 (8×B200, Qwen2-1.5B): keep/wait/abort/resume all green, no regression.

HJSang:
@codex review

qywu:
Took this for an independent spin — looks solid. The 10 unit tests pass, and a GPU E2E on Qwen2-1.5B (real `AsyncLLM`→ZMQ→`PauseController` path) confirms `keep` truly freezes an in-flight request (0 tokens for the whole pause window) and resumes to completion; `abort`/`wait` return promptly when nothing's in flight.

One latent edge case worth a guard: in `PauseController.handle_pause`, a second `abort`/`wait` pause arriving before the first drains overwrites `self._pending_reply`, so only the later reply ever fires and the first `pause_scheduler()` caller is stranded on its ZMQ await. Harmless in the one-pause-per-step RLHF flow, but rejecting/no-oping a pause while `_pending_reply is not None` would close it.

python/tokenspeed/runtime/engine/pause.py
```diff
@@ -0,0 +1,186 @@
...
+        if req.mode == "keep":
+            # Freeze in place and reply now — nothing to drain.
+            self.state = PauseState.PAUSED_ALL
+            self._send.send_pyobj(PauseSchedulerReqOutput(success=True))
+            return
+
+        # abort / wait: stop admitting new requests, keep stepping so in-flight
+        # requests drain, and reply once the scheduler is empty. Both also
+        # cancel grammar-queued (still-compiling) pre-pause requests.
+        self.state = PauseState.PAUSED_NEW
```
qywu, line 168: Guard to close the double-pause strand: a second `abort`/`wait` pause before the first drains overwrites `_pending_reply` here, so the first `pause_scheduler()` caller never gets its reply and hangs on its ZMQ await. Rejecting a pause while one is already pending fixes it (only this path needs it — `keep` doesn't touch `_pending_reply`).

```suggestion
        if self._pending_reply is not None:
            self._send.send_pyobj(
                PauseSchedulerReqOutput(
                    success=False, message="a pause is already in progress"
                )
            )
            return

        self.state = PauseState.PAUSED_NEW
```

python/tokenspeed/runtime/engine/pause.py
qywu, line 1: codex has some comments that can be addressed

python/tokenspeed/runtime/engine/generation_output_processor.py
```diff
@@ -119,6 +119,10 @@ def __init__(
         # abort related
         self.to_abort = False
         self.to_abort_message = None
+        # Client-initiated aborts skip streaming a finish (the TM already tore
+        # down its state). Pause-initiated aborts set this so the passive client
+        # still receives a terminating finish.
+        self.abort_notify_client = False
```
qywu, line 125: should abort return error to the user?

python/tokenspeed/runtime/engine/event_loop.py
qywu, line 1: codex p1 comments can be addressed

qywu (approved):

python/tokenspeed/runtime/engine/pause.py
HJSang: Good catch. Add a guard.

python/tokenspeed/runtime/engine/pause.py
HJSang: Done

python/tokenspeed/runtime/engine/generation_output_processor.py
HJSang: same to vllm, the client will recieve something like ```finish_reason={"type":"abort", "message":"Aborted by pause"}```

python/tokenspeed/runtime/engine/event_loop.py
HJSang: Done

PR #358, opened 2026-06-05
Make tokenspeed-kernel imports lazy

antiagainst:
This is a pre-step to structure kernel packages better.

### DIFF
--- a/tokenspeed-kernel/python/tokenspeed_kernel/backends.py
+++ b/tokenspeed-kernel/python/tokenspeed_kernel/backends.py
@@ -1,0 +1,132 @@
+# Copyright (c) 2026 LightSeek Foundation
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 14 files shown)


tokenspeed-kernel/python/tokenspeed_kernel/backends.py
```diff
@@ -0,0 +1,132 @@
...
+    """Import built-in registration modules for one or more op families.
+
+    This is intentionally separate from importing public API packages. Built-in
+    modules perform platform detection and optional vendor-library imports, so
+    they are loaded only when an API call needs them or tests request them.
+    """
+
+    family_names = _normalize_families(families)
+
+    with _LOCK:
```
borontion, line 120: will this cause any host-side latency as we calling  `load_builtin_kernels` before each kernel launch?

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/__init__.py
```diff
@@ -22,14 +22,8 @@
 
 import math
 
-# Backend registration (side-effect imports)
-import tokenspeed_kernel.ops.attention.cuda  # noqa: F401
```
zhyncs, line 26: QQ what’s the motivation behind this change? Will we gain any performance benefits, or regret it later?

antiagainst:
This was an early attempt to see how we publish split wheels. It is more involved going down this route so I decided not to proceed.

PR #363, opened 2026-06-05
[AMD] moe: fused single-kernel Gluon routing fast path for small-M decode (gfx950)

panditsa:
## Summary

  Adds a fused single-kernel Gluon routing fast path for the small-M decode regime on gfx950, where the generic multi-kernel MoE routing pipeline is dominated by launch overhead.

  - New kernel ops/moe/gluon_decode_routing_gfx950.py: computes routing in one Gluon kernel launch for M <= SMALLM_MAX_M (16).
  - Gates the fast path in triton_kernels_routing: small-M inputs that pass gluon_route_supported use gluon_fused_route; everything else falls back to the existing pipeline unchanged.
  - Stable per-expert ranks via a register-only O(G²) all-pairs comparison, bounded by G = M * topk <= GLUON_ROUTE_MAX_G (64) so the whole route fits one workgroup.
  - Reorders the mxfp4 backend preference in selector.py to prefer gluon_kernel over triton_kernel.
  - Correctness guards (runtime M <= 16 and G-bound assertions) and collapsed redundant overlapping prefix stores.
  - Prefill and large-M paths are untouched; only small-M routing behavior changes.

## Test Plan

- Added test/ops/moe/test_gluon_decode_routing_gfx950.py, validating the fused route against the generic pipeline across decode-shaped inputs.
- pytest test/ops/moe/test_gluon_decode_routing_gfx950.py on MI355X: 12/12 passing.
- Decode profiling on MI355X shows improved tok/s versus the generic routing baseline (gluon-first vs triton-first on main).


python/tokenspeed/runtime/layers/moe/core/selector.py
```diff
@@ -47,8 +47,8 @@
     ),
     "mxfp4": (
         "flashinfer_mxfp4",
-        "triton_kernel",
         "gluon_kernel",
+        "triton_kernel",
```
panditsa, line 51: Let me know if we are not ready to fully commit to the gluon_kernel over triton_kernel yet. I can revert this change.

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/gluon_decode_routing_gfx950.py
borontion, line 1: seems this only used in tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/triton_kernels.py? 

we are organizing code as `ops/<family>/<impl>`. so probably better to put it in tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/triton_kernels.py

python/tokenspeed/runtime/layers/moe/core/selector.py
borontion: I think both `gluon_kernel` and `triton_kernel` moe backend will be using your routing kernel. so we can use both backend.

antiagainst: Yeah fine for now. This is using ad-hoc runtime layer kernel selection. Eventually this needs to be folded into kernel selection logic. I guess @borontion you will clean it up?

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/triton_kernels.py
```diff
@@ -189,6 +194,11 @@ def triton_kernels_routing(
     n_tokens, _ = logits.shape
 
     assert sm_first is False, "sm_first=True not supported for triton_kernels_routing"
+
+    # For small M, route with the single fused Gluon kernel.
```
antiagainst: Hmm. This couples the gluon impl with triton impl so we cannot separate them out later easily. It is also intercepting logic of triton kernels rounting and completely redirecting it to another solution. Ideally we want to register it separate kernels and use that existing mechanism. Can we consider using `traits`? Is it too static for the purpose of what we want to do here?

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/gluon_decode_routing_gfx950.py
```diff
@@ -0,0 +1,417 @@
...
+    TKP: gl.constexpr,  # next_pow2(topk)
+    X_DTYPE: gl.constexpr,  # gate element type (logits dtype)
+    L1: gl.constexpr,  # 1D blocked layout used by the consuming kernel
+    LT: gl.constexpr,  # 2D blocked layout for the [MP, EP] logits tile
+):
+    """Fused in-kernel top-k matching ``topk_forward(apply_softmax=True)``.
+
+    Selects, per token row, the top ``TOPK`` experts by logit value (ties to
+    the smaller expert id, descending value order) and the softmax gate over
+    the selected logits -- reproducing the OAI ``_topk_forward`` semantics
```
antiagainst: nit: "reproducing the triton kernels ``_topk_forward``"

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/gluon_decode_routing_gfx950.py
```diff
@@ -0,0 +1,417 @@
...
+        valsf = gl.where(take, gl.gather(gat_r, tok, axis=0), valsf)
+    # cast like topk_forward's softmax(...).to(x_dtype) before the gate store.
+    return idx, valsf.to(X_DTYPE)
+
+
+# ===========================================================================
+# Small-M (M <= 16): single-workgroup, stable-order, single-block collapse.
+# ===========================================================================
+@gluon.jit
+def _fused_route_smallM(
```
antiagainst: nit: `small_m` to follow the naming style

python/tokenspeed/runtime/layers/moe/core/selector.py
borontion: yeah I will handle it.

borontion:
@panditsa also is it possible to fuse the quantize fp8 to this kernel?

image

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/gluon_decode_routing_gfx950.py
panditsa: Since the last review round the kernel is no longer only used by `triton_kernels.py`, so following `ops/<family>/<impl>` suggestion, I folded it into the existing gluon impl, `ops/moe/gluon.py`, alongside the other gluon moe kernels.

panditsa:
> @panditsa also is it possible to fuse the quantize fp8 to this kernel?
> image

The PR is ready to review/merge if no other issues. I'll try to fuse the quant op in the meantime, but I don't mind creating another PR if this one can be merged.

borontion (approved):

PR #370, opened 2026-06-06
fix(cache): Coarsely fence the compute stream behind the host loadback stream on.

LorrinWWW:
## Summary

retraction→reload→resume-decode (and while loadback is in flight), since a captured decode graph can't honor the per-layer wait_until whose events live on the non-captured load_stream, allowing the replay to read KV mid-reload.


LorrinWWW:
> This PR has been inactive for 14 days and is marked as stale. It will be closed in 3 days if there is no further activity.

this is still needed.

PR #374, opened 2026-06-07
feat(kernel): introduce unified moe kernel api

borontion:
## Summary

This PR moves moe backend concepts into tokenspeed kernel. MoE will be a platform-agnostic layer that invoke moe kernel api. 

### MoE entrypoint

Introduce a single `moe_apply` api. Under the hood it can launch a single or more multiple kernels depending on the implementation. This api is too coarse from kernel-level. Moving forward, we will replace it as a kernel execution plan.

See `tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/__init__.py`.

### Supported type and backend

Currently I keep the following weight dtype + implementation:

- Unquant (fp16): flashinfer cutlass
- fp8 (with scale): flashinfer cutlass
- mxfp4: triton / gluon / flashinfer trtllm
- nvfp4: flashinfer cutlass / flashinfer trtllm / flashinfer cutedsl

Other data types are not actively used in our current target models. Also drop the past triton backend for fp8/unquant, which has convoluted dependencies. A portable impl for fp8/unquant which can run amd gpu will be added later. Also renamed the backend name for 'triton_kernel' to 'triton'.

### Kernel early selection

MoE is different from attention where the weights need to be preshuffled after model loaded. The preshuffle and actual moe impl are tightly coupled. So we need to determine the kernel implementation earlier, before engine accepts incoming requests. 

This is done by adding a specific `moe_plan` utility function. It will select the corresbonding moe kernel based on various features and the current platform. The selection is encoded into a dict which includes the target kernel name, which will be passed to `moe_apply`.

### AMD Gluon kernel

Decouple amd gluon kernel from other kernel utilit by copying used code into `tokenspeed-kernel-amd` module. This includes 3 parts

1. fp8 quant
2. fallback triton kernel routing
3. fallback triton kernel group gemm

See: `tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/moe/utils.py` for ported triton kernel code.

Currently I lower the priority for the gluon kernel due to known performance issue.


tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/gluon/mxfp4.py
```diff
@@ -0,0 +1,240 @@
...
+            b_microblock_size=MXFP_BLOCK_SIZE,
+            out_dtype=out_dtype,
+        )
+
+        w.w13_weight_triton_tensor = w13_weight
+        w.w2_weight_triton_tensor = w2_weight
+        del w.w13_weight
+        del w.w2_weight
+
+        torch.cuda.empty_cache()
```
antiagainst, line 175: What's this for?

tokenspeed-kernel/python/tokenspeed_kernel/selection.py
```diff
@@ -426,6 +440,11 @@ def spec_matches_traits(
             every requested trait must be explicitly present on the spec.
     """
     for trait_name, trait_value in traits.items():
+        if trait_name == "ispp":
```
antiagainst, line 448: This is a special casing in the general infra. Can we add some docstr of this special treat? it's not very straightforward what it stands for just from the name. 

We might also need to think about how to extend infra to support such flexible cases better in the future (like passing in lambda?) if this is needed freqently. I do feel current mechanism might be too static. But don't need to incur the overhead just right now. We can revisit when this becomes a general issue.

antiagainst (approved):
Awesome! Looks large change but mostly moving code around and fleshing out kernel interface and registration. Just two nits from me.

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/gluon/mxfp4.py
borontion: I just copied this part from original code. I think it was trying to free some gpu mem. Del a tensor does not mean it gets deallocated.

PR #390, opened 2026-06-09
refactor(spec-decode):  simplify Llama Eagle3 attention path for #217 (1/3)

rjzhb:
## Summary

First PR in a series refactoring the spec-decode attention path introduced in #217.

- Lift fused KV pre-write into base `llama.LlamaAttention._attn` (= pure perf win for target Llama when backend supports it).
- Eagle3 `LlamaAttention` collapses the 60-line monolith forward + 4 `_maybe_*` helpers + 80-line `__init__` copy into a 2-method subclass: `_attn` overrides the draft-first-step dispatch B (correction + q-slice + DECODE / post-slice), inactive steps delegate to `super()._attn`.
- Correction (= `spec_num_tokens - accept_lengths` trim of `draft_seq_lens_buf`) moves from the drafter and the standalone `_maybe_trim_cache_seqlens` helper into a single `_apply_correction` method, called atomically alongside the q-slice in the active branch. Drafter plumbs `accept_lengths` / `draft_seq_lens_buf` via `ForwardContext`.
- Drops the temporary `pre_attention_trim` class flag + drafter-side isinstance gate now that the trim lives next to its only consumer.

`ctx.draft_first_step_reduce` stays for now (= Qwen NextN / DeepSeek V3 NextN / `comm_manager` still read it); follow-up PRs in the series migrate those families and then delete the field.

no behaviour change for target Llama or Eagle3 draft. Builds on #217.

### DIFF
--- a/python/tokenspeed/runtime/execution/drafter/eagle.py
+++ b/python/tokenspeed/runtime/execution/drafter/eagle.py
@@ -219,10 +219,11 @@ class Eagle(BaseDrafter):
             draft_input, bs, draft_input.input_num_tokens
         )
         input_ids = maybe_substitute_mm_pad(input_ids, self.mm_pad_substitute_id)
-        # Llama Eagle3 narrows for prefill catch-up too; Qwen/DeepSeek do not.
+        # Llama Eagle3 narrows for any non-idle catch-up (EXTEND/MIXED/
+        # TARGET_VERIFY/DECODE); Qwen/DeepSeek keep is_decode() only.
         draft_first_step_reduce = forward_mode.is_decode() or (
             isinstance(self.draft_model_runner.model, LlamaForCausalLMEagle3)
-            and forward_mode.is_target_verify()
+            and not forward_mode.is_idle()
         )
 
         draft_first_mode = (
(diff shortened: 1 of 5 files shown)


python/tokenspeed/runtime/execution/drafter/eagle.py
```diff
@@ -27,6 +27,7 @@
 from tokenspeed_kernel.ops.sampling import argmax as sampling_argmax
 from typing_extensions import override
 
+from python.tokenspeed.runtime.models.llama_eagle3 import LlamaForCausalLMEagle3
```
LorrinWWW: no `python.` prefix

python/tokenspeed/runtime/execution/drafter/eagle.py
```diff
@@ -219,21 +219,11 @@ def _run_first_step(
...
-            and self.attn_backend.support_kv_cache_prewrite
-            and isinstance(draft_model, BaseCausalLM)
-            and draft_model.pre_attention_trim
-        ):
-            correction = (self.spec_num_tokens - draft_input.accept_lengths).to(
-                self.draft_seq_lens_buf.dtype
-            )
-            self.draft_seq_lens_buf[:bs].sub_(correction)
+        # Llama Eagle3 narrows for prefill catch-up too; Qwen/DeepSeek do not.
+        draft_first_step_reduce = forward_mode.is_decode() or (
```
LorrinWWW, line 223: We will eventually remove this right?

rjzhb: yeah, Now Qwen/Deepseek needs them

LorrinWWW (approved):
temporal if-condition will be removed when 3/3 refactoring PRs are all in.

PR #393, opened 2026-06-09
feat(engine): Sleep / Wake Up API (release/resume_memory_occupation)

HJSang:
## Summary

Completes TokenSpeed's half-ported SGLang-style **data-plane sleep/wake**: wires `release_memory_occupation` / `resume_memory_occupation` / `is_sleeping` end-to-end with selective `weights`/`kv_cache` tags. A *release* auto-pauses + drains the scheduler (control plane) and only then frees GPU memory via `torch_memory_saver` (data plane); *resume* re-maps, repairs the KV cache, and unpauses. Primary driver: the **RL/RLHF loop** (free the GPU for the trainer between rollouts, push fresh weights, resume).

Pure-Python — the C++ scheduler `.so` is untouched.

### Before this PR (half-ported state)
`--enable-memory-saver`, the `torch_memory_saver` adapter, region-wrapped weights/KV, the engine API stubs, and client communicators existed — but the scheduler-side dispatch was **absent**, `memory_saver.pause/resume` was **never called**, and the io_struct `tags` field was missing (so the engine API would have `TypeError`'d).

## What changed
- **`PauseController`** — generalized the deferred reply into a `_PendingDrain(on_drained, on_cancelled)` action + a `released` flag, so "release after drain" reuses the proven async-drain path (pause/resume behavior unchanged).
- **`MemoryOccupationController`** (new) — GPU-memory orchestration: tag→`memory_saver.pause/resume`, `released_tags` bookkeeping (partial wake, double-release/not-released rejection), prefix-cache reset on release, KV repair on wake.
- **Wiring** — `request_handler` dispatch for Release/Resume/IsSleeping; `event_loop` constructs the controller and **skips DP `execute_idle_forward` while `released`** (weights are unmapped); `BaseTokenToKVPool.clear_kv_buffers()` zeros remapped KV.
- **Adapter / regions** — `region(tag, enable_cpu_backup)` pass-through; weights tagged `enable_cpu_backup=True` (byte-exact restore), KV `False` (discard).
- **Surface** — Engine `release/resume_memory_occupation(tags)` + `is_sleeping()`; client returns `success` outputs; `torch_memory_saver==0.0.9.post1` pinned (import-guarded).

## Validation
- **26 unit tests** (local + real GPU env), incl. adapter tag pass-through against the real `torch_memory_saver`.
- **Live E2E on B300** (Qwen2-0.5B, `--enable-memory-saver`), all cases pass:
  - release **freed 26.7 GiB**, `is_sleeping`→True; resume restored to within ~2 MiB
  - **token-identical generation across the sleep cycle with CUDA graphs** (weights restored byte-exact)
  - RL multi-stage tag flow (release both → resume weights [still sleeping] → resume kv_cache [awake])
  - error paths return `success=False`

## Scoped out (follow-ons)
- **HTTP endpoints** — require new RPCs in the external `tokenspeed-smg-grpc-proto`/`-servicer` packages (pause/resume is Python-only for the same reason).
- **deepseek_v4 KV region wrapping** — currently `del enable_memory_saver`; needed for the real serving model (validated on a small MHA model first).
- **TP/DP multi-GPU run** of the idle-forward gate (logic in place; single-GPU validated).

Design: `docs/superpowers/specs/2026-06-07-sleep-wakeup-api-design.md` · Plan + results: `docs/superpowers/plans/2026-06-07-sleep-wakeup-api.md`

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -10,7 +10,9 @@ WORKDIR /workspace
 
 COPY . /workspace
 
-RUN apt-get update && apt-get install -y --no-install-recommends libssl-dev libopenmpi-dev && \
+# libnuma1: runtime dep of torch_memory_saver (the Sleep/Wake Up memory saver),
+# which is a hard dependency of the python package; the base runner image lacks it.
+RUN apt-get update && apt-get install -y --no-install-recommends libssl-dev libopenmpi-dev libnuma1 && \
     rm -rf /var/lib/apt/lists/*
 
 RUN MAX_JOBS=${MAX_JOBS} FLASHINFER_CUDA_ARCH_LIST="9.0a 10.0a" TOKENSPEED_KERNEL_BACKEND=cuda \
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -69,6 +69,10 @@ dependencies = [
     "tokenspeed-smg-grpc-proto==0.4.10.post20260612",
     "tokenspeed-smg-grpc-servicer==0.5.5.post20260612",
     "torch==2.11.0",
+    # Tag-capable sleep/wake allocator (region(tag, enable_cpu_backup)/pause/
+    # resume). Bundles cu12/cu13 binaries; import-guarded so it's a no-op unless
+    # --enable-memory-saver is set. See runtime/utils/torch_memory_saver_adapter.py.
+    "torch_memory_saver==0.0.9.post1",
     "torchvision",
     "tqdm",
     "transformers==5.12.0",
--- a/python/tokenspeed/runtime/cache/req_to_token_pool.py
+++ b/python/tokenspeed/runtime/cache/req_to_token_pool.py
@@ -62,7 +62,9 @@ class ReqToTokenPool:
         self.size = size
         self.max_context_len = max_context_len
         self.device = device
-        with memory_saver_adapter.region():
+        # Tag as "kv_cache": this per-token page state is invalid once KV is
+        # discarded, so it is released/restored alongside the KV cache.
+        with memory_saver_adapter.region(tag="kv_cache", enable_cpu_backup=False):
             self.req_to_token = torch.zeros(
                 (size, max_context_len), dtype=torch.int32, device=device
             )
--- a/python/tokenspeed/runtime/engine/event_loop.py
+++ b/python/tokenspeed/runtime/engine/event_loop.py
@@ -41,6 +41,7 @@ from tokenspeed.runtime.distributed.process_group_manager import (
     process_group_manager as pg_manager,
 )
 from tokenspeed.runtime.engine.generation_output_processor import OutputProcesser
+from tokenspeed.runtime.engine.memory_occupation import MemoryOccupationController
 from tokenspeed.runtime.engine.pause import PauseController
 from tokenspeed.runtime.engine.request_handler import RequestHandler
 from tokenspeed.runtime.engine.scheduler_utils import (
@@ -93,6 +94,7 @@ from tokenspeed.runtime.utils.exceptions import get_exception_traceback
 from tokenspeed.runtime.utils.nvtx import nvtx_range
 from tokenspeed.runtime.utils.process import register_usr_signal
 from tokenspeed.runtime.utils.server_args import PortArgs, ServerArgs
+from tokenspeed.runtime.utils.torch_memory_saver_adapter import TorchMemorySaverAdapter
 
 logger = get_colorful_logger(__name__)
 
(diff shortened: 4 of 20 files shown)


HJSang:
Update: **deepseek_v4 KV region wrapping added** (was listed as a follow-on).

- `deepseek_v4.py` now allocates its KV buffers (swa / compressed / compressor_state / indexer / indexer_state) inside `region(tag="kv_cache", enable_cpu_backup=False)` instead of discarding `enable_memory_saver`.
- `BaseTokenToKVPool.clear_kv_buffers()` extended to zero V4's buffer names after wake.

Validated on nv2 (b300): deepseek_v4 imports cleanly with the real compiled kernel/scheduler env; the MHA sleep/wake E2E **regression still passes all cases** (release frees + `is_sleeping`→True, token-identical resume with CUDA graphs, RL multi-stage, error paths). A full DeepSeek-V4-Pro model run is pending free multi-GPU capacity (the box's free GPUs currently have only ~25 GB each; V4 needs ~4 GPUs × hundreds of GB).

Remaining follow-ons: HTTP endpoints (external smg-grpc packages) and a TP/DP multi-GPU run of the idle-forward gate.

docs/superpowers/plans/2026-06-07-sleep-wakeup-api.md
```diff
@@ -0,0 +1,1359 @@
+# Sleep / Wake Up API Implementation Plan
```
qywu: Please remove this file from the PR. It's a Claude Code `superpowers` planning/results artifact (1,359 lines) — point-in-time process scratch, not project documentation. A few reasons:
- `docs/superpowers/` doesn't exist on `main`; `docs/` is a curated VitePress site (`guides/`, `serving/`, `configuration/`, …). This drops planning scratch into a *published* docs tree.
- It won't be maintained as the code evolves, so it will drift and mislead.
- Planning notes + validation results like this belong in the PR description (where most of it already is) or a tracking issue, not committed under `docs/`.

If there's durable content here worth keeping, fold the essentials into a maintained doc or the module docstrings instead.

docs/superpowers/specs/2026-06-07-sleep-wakeup-api-design.md
```diff
@@ -0,0 +1,335 @@
+# Sleep / Wake Up API — Design
```
qywu: Same ask as the plan doc — please remove this 335-line design spec from the PR. It's a generated `superpowers/specs` artifact and the first thing under `docs/superpowers/`, which isn't part of the existing published `docs/` site.

If any of the design rationale is durable and worth keeping for future maintainers, distill it into the `memory_occupation.py` / `pause.py` module docstrings or a short maintained page under the existing `docs/` structure. As a standalone point-in-time spec it'll go stale against the code.

qywu (commented):
These two `docs/superpowers/` files (~1,700 of the PR's ~2,500 added lines) are Claude Code workflow artifacts rather than project docs. Recommend dropping them so the PR is just the sleep/wake implementation + tests; the design/results already live in the PR description. Details inline on each file.

qywu (commented):
`test/runtime/test_io_struct_memory_occupation.py` only tests `@dataclass` defaults/assignment (pure data holders, no behavior), and every contract it touches is already covered functionally by `test_memory_occupation_controller.py`. Recommend removing it — detail inline.

test/runtime/test_io_struct_memory_occupation.py
```diff
@@ -0,0 +1,29 @@
+# SPDX-License-Identifier: Apache-2.0
```
qywu: Recommend removing this file. These six dataclasses (`ReleaseMemoryOccupationReqInput/Output`, `ResumeMemoryOccupationReqInput/Output`, `IsSleepingReqInput/Output`) are pure data holders — no `__post_init__`, no validation, no methods — so all four tests just assert that Python's `@dataclass` stores values and applies declared defaults:
- `ReleaseMemoryOccupationReqInput().tags is None` / `...Output().success is True` test that a field defaulted to `None`/`True` has that value.
- `(success=False, message="x").message == "x"` tests assignment.
- `assert IsSleepingReqInput() is not None` is tautological — a constructor never returns `None`.

They're change-detectors: editing a dataclass forces editing the test in lockstep, with no independent verification, and they can't catch a real bug because there's no behavior to break.

Every meaningful contract here is already covered by `test_memory_occupation_controller.py`, which constructs the same dataclasses where the values actually drive behavior — `tags=None ⇒ all` (lines 74/91/130/155), selective `tags=["weights"]`/`["kv_cache"]` (105–116), `success` True/False (85/99/123/131/142/149), `is_sleeping` transitions (86/100/111/116). A renamed field or changed default would fail there too, with real meaning.

(Fully agree with the `AGENTS.md` instinct to test changed code — that coverage just belongs in the controller/behavior tests, where it already lives. The controller, adapter, and GPU E2E tests are the ones doing real work.)

qywu (commented):
—

docs/superpowers/plans/2026-06-07-sleep-wakeup-api.md
HJSang: updated

test/runtime/test_io_struct_memory_occupation.py
HJSang: updated

HJSang:
@qywu  thank you for the comments. I have resolved them. Please take another look.

qywu (approved):

PR #404, opened 2026-06-09
Split AMD Gluon kernels package

antiagainst:


tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/moe/fused_mxfp_gfx950.py
```diff
@@ -21,29 +21,15 @@
...
-from tokenspeed_kernel.registry import KernelRegistry, Priority, register_kernel
-from tokenspeed_kernel.signature import (
-    ScaleFormat,
-    dense_tensor_format,
-    format_signature,
-    format_signatures,
-    tensor_format,
-)
 
 with redirect_triton_to_tokenspeed_triton():
```
borontion: so we have the depedency of tokenspeed-triton-kernel. is that ok? or we should copy all used utility from triton_kernel?

antiagainst: yup we want to avoid that. copied in those definitions given they are small anywya.

borontion (approved):

.github/workflows/pr-test.yml
```diff
@@ -132,6 +135,35 @@ jobs:
             echo "changed=0" >> "$GITHUB_OUTPUT"
           fi
 
+      - name: Detect tokenspeed-kernel-amd source changes
```
borontion, line 138: looks like when there is no change to tokenspeed-kernel-amd we will not install it locally? this can cause ci failure as there is no released tokenspeed-kernel-amd package: https://github.com/lightseekorg/tokenspeed/actions/runs/27313828556/job/80689894639?pr=374

antiagainst: https://github.com/lightseekorg/tokenspeed/pull/417

PR #413, opened 2026-06-10
fix: size req pool max_num_seqs + 1 so a full batch can't collide with the scheduler's padding slot 0

LorrinWWW:
## Summary

The scheduler reserves request-pool slot 0 for padding/CUDA-graph warm-up, so real requests must use slots 1..max_bs and the pool must have max_bs + 1 rows. Python was passing max_req_pool_size = max_num_seqs, which let a full batch assign a live request to slot 0; padded forwards then corrupted that request’s per-slot state across TP ranks, causing scheduler divergence and mismatched Gloo collectives, matching the observed TP4+EAGLE3 deadlocks. This PR fixes the invariant by passing max_num_seqs + 1.


LorrinWWW:
non issue

PR #418, opened 2026-06-11
Add InstantTensor weight loader (--load-format instanttensor)

gongwei-130:
## Purpose

Port the InstantTensor weight loader from vLLM (vllm-project/vllm#36139) and SGLang (sgl-project/sglang#27800), and enable it for TokenSpeed's Kimi CI.

[InstantTensor](https://github.com/scitix/InstantTensor) accelerates loading safetensors weights on NVIDIA GPUs via distributed loading, pipelined prefetching, and direct I/O (with optional GPUDirect Storage). It loads tensors directly onto the GPU; the resulting weights are bit-for-bit identical to the default loader.

## Changes

**Loader**
- Add `LoadFormat.INSTANTTENSOR` and the `--load-format instanttensor` CLI choice.
- Add `instanttensor_weights_iterator()` in `weight_utils.py` (NVIDIA-guarded via `current_platform`, passing the world process group for multi-rank jobs).
- Route the format through `DefaultModelLoader` (treated like safetensors for shard selection).
- Add `instanttensor>=0.1.9` as a **core dependency** so it is available in the released image, CI, and user installs in one place. It is a CUDA-only wheel but installs on any x86_64 host and is imported lazily (guarded by `try/except` + an NVIDIA-platform check), so it is harmless on non-NVIDIA installs.

**Fix: GPU-direct loading OOM on Kimi**
- `KimiK25.load_weights` materialized the entire weight iterator into a Python list before loading, holding every loaded tensor alive at once. That is harmless for the CPU-staging loaders (tensors sit in host RAM), but with a GPU-direct loader it keeps the whole model resident on the device and OOMs during weight load. Stream the language weights to `language_model.load_weights` (a lazily consumed generator) instead. Correctness-preserving: the same `(name, tensor)` pairs reach the consumer in the same order and are copied into the same pre-allocated params.

**CI**
- Switch the six `Kimi-K2.5-NVFP4` jobs (eval: gsm8k / mmlu / aime25 / gpqa-diamond / ocr-bench; perf: agentic) to `--load-format instanttensor` — speeds up weight loading and exercises the loader end to end (the eval jobs double as an accuracy check).

**Docs**
- New guide (`docs/guides/instanttensor.md`) + server-param note. Documents that GPU-direct loading requires `load_weights` to consume the iterator lazily, and the InstantTensor buffer tunables (`INSTANTTENSOR_BUFFER_SIZE` / `INSTANTTENSOR_MAX_FREE_MEM_USAGE`).

## Test Result

- Unit / parity tests (`test/runtime/test_instanttensor_loader.py`): config/CLI/routing coverage plus a gpt2 **bit-exact parity** check (`torch.equal`) against the default safetensors loader. Verified on a B200 with `instanttensor==0.1.9`: **4/4 pass**.
- End-to-end on **8×B200** serving `nvidia/Kimi-K2.6-NVFP4` with `--load-format instanttensor`: loads with ~102 GiB free/GPU (the old buffered-load path OOM'd at ~172 GiB), EAGLE3 draft loads, engine ready, generations served.

> Note: the Kimi-K2.5 / tp4 CI config was not reproducible on the dev box (model not cached, box offline at the time), so it is validated by reasoning + the K2.6/tp8 run. Recommend a manual trigger of one Kimi eval job (e.g. gsm8k) to confirm in the real CI environment before relying on nightly.


LorrinWWW (approved):
lgtm

PR #421, opened 2026-06-11
fix(kernel): Fix gluon MoE GEMM numerical bug

Max191:
## Summary

The gluon MoE GEMM kernel had a numerical bug due to an incorrect schedule of global and LDS loads. For large sequence lengths, the gluon kernel uses a split output tile to interleave mfmas and loads. The global to LDS transfers for scale operands are issued for the full tile during the load issue of the first quadrant, but the LDS loads for scales are performed per-quadrant. This caused a race because the first quadrant's global transfer would overwrite the LDS tile of another quadrant before it loads its scales from LDS. This PR fixes the bug by moving the LDS load before the global to LDS transfer and updating the wait counts accordingly.

This also fixes 2 small bugs in the kernel configuration logic. The first is an issue where the M block size can fall into an unsupported size of 48. The second is for pre-shuffled cases, where the layout chosen by the kernel is different from the pre-shuffled layout for large sequence lengths. The kernel config logic now selects the correct block sizes in both cases.

Performance is roughly unchanged:

### Non-Preshuffled

| Tokens | Routed rows | Baseline median us | PR median us | Delta | Speedup |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 4096 | 8192 | 807.63 | 796.11 | -11.52 us | 1.014x / +1.45% |
| 8192 | 16384 | 1123.85 | 1108.11 | -15.74 us | 1.014x / +1.42% |

### Preshuffled

| Tokens | Routed rows | Baseline preshuffle | PR preshuffle median us | Delta | Result |
| ---: | ---: | ---: | ---: | ---: | --- |
| 4096 | 8192 | 952.89 us | 955.33 | +2.44 us | 0.997x / -0.26% |
| 8192 | 16384 | Compile failure | 1473.12 | n/a | PR fixes W_VIA_VGPR block-N static assert |

VGPR/SPGR count are unchanged

## Test Plan

A new numerical kernel test is added for the MoE gluon kernel implementation compared against the triton MoE kernel implementation, which is more stable and is currently producing accurate numerical results end to end. The new test covers the large sequence lengths, and also smaller sequence lengths to cover the different specialization tile layouts for the gluon kernel. It also runs the comparison with the preshuffled and non-preshuffled kernel.

The test constrains input, weight, and scale inputs to a set of power-of-2 values in order to keep floating-point rounding errors to a minimum, since these kernels are prone to floating-point rounding error differences.

Without the schedule fix, the test fails for large sequence lengths. With the schedule fix, the test passes.


tokenspeed-kernel/test/ops/moe/test_gluon_moe_gemm_gfx950.py
borontion, line 1: can we do this test in `tokenspeed-kernel-amd`?

Max191: Looks like I may need to rebase onto your PR to make this work: https://github.com/lightseekorg/tokenspeed/pull/374

I'll try rebasing and changing the base branch in this PR for now. Going to convert to a draft until your PR merges.

Max191: Context from offline discussion: We will need to move weight preprocessing over to tokenspeed-kernel-amd as a followup to #374, so I'll land this PR with the test in tokenspeed-kernel for now and move it over when we restructure the weight preprocessing.

borontion (approved):

PR #429, opened 2026-06-12
refactor(spec-decode): simplify Qwen3.5 NextN attention path for #217 (2/3)

rjzhb:
### Summary

This is the second PR in the series refactoring the spec-decode attention path from #217. It applies the same base-hook pattern introduced in #390 (1/3) for Llama Eagle3, now to Qwen3.5 NextN.

`Qwen3_5DraftForCausalLM` / `Qwen3_5DraftAttentionDecoderLayer` collapse the bespoke draft forward path into a two-method subclass of the base `Qwen3_5AttentionDecoderLayer`:

* `_attn` overrides the draft first-step dispatch path: correction + q-slice + `DECODE`.
* Inactive steps delegate to `super()._attn`.

The correction logic, `spec_num_tokens - accept_lengths` trimming of `draft_seq_lens_buf`, now lives in a single `_apply_correction` method next to its only consumer, and is plumbed through `ForwardContext`, mirroring #390 (1/3). `_maybe_narrow_residual` handles the NextN residual narrowing.

This is restricted to single-layer drafts for now, asserted in `__init__`. `_apply_correction` mutates per-layer state, so multi-layer NextN support needs the trim to be hoisted before this restriction can be relaxed.


python/tokenspeed/runtime/models/qwen3_5_nextn.py
```diff
@@ -24,11 +24,13 @@
 from collections.abc import Iterable
 
 import torch
+from tokenspeed_kernel.ops.activation.triton import sigmoid_mul
```
LorrinWWW, line 28: move to the below import group

LorrinWWW (approved):
lgtm. remember to clean the model-specific condition in the 3/3 PR.

PR #436, opened 2026-06-12
fix: limit internal activation dtype trait for moe plan

borontion:
## Summary

Adding `input` as a special value for `internal_activation_dtype` in moe api that uses the dtype the input activations have.

## Test Plan

```sh
CI=1 python test/runtime/models/test_generation_models.py
```


antiagainst (approved):

borontion:
> @borontion https://github.com/lightseekorg/tokenspeed/actions/runs/27437439370/job/81103234777?pr=436 h100 ut-kernel failure is caused by #432

fixed.

PR #447, opened 2026-06-15
KV-Cache Refactor Part 1: Introduce Unified Block Pool

wangbo981016:
# KV-Cache Refactor Part 1: Introduce Unified Block Pool

Introduces a **flat KV-cache** path alongside the existing radix KV-cache, providing unified memory and prefix-cache management for hybrid-attention models (GDN/mamba state layers + attention layers). The entire flat path is isolated behind `#if TOKENSPEED_FLAT_KVCACHE` (C++) and runtime flat-ext gates (Python), keeping the **radix/main path unaffected** — every commit is dual-arm verified (radix all-green + flat known-failure set unchanged).

**Design goal:** once the refactor is complete, adding a new complex hybrid model should require **no C++ changes** — a new model declares only its layer→group mapping and per-layer component bytes in Python, and the plan solver handles sizing/layout. The C++ side is already granularity-blind (pool), per-group (matching/eviction/block_size), and covers the full / sliding-window / snapshot reuse rules. A new C++ manager + kind would only be needed for a genuinely new reuse rule, not for heterogeneous granularity or a new component mix.

### C++ scheduler core (`tokenspeed-scheduler/csrc/`)

- **Unified block-pool model**: one `BlockPool` block-id space, `BlockRef` RAII ownership, `BlockTable` (per-request, per-group) logical→physical mapping. "Cached" and "free" are the same LRU free list — a ref-0 block keeps its hash (still prefix-matchable) until it reaches the LRU head and is evicted.
- **Per-attention-kind managers** (`FullAttnManager` / `SwaManager` / `MambaStateManager`): each owns its own match/register/reclaim policy (full attention keeps everything, SWA punches expired blocks, mamba retains only the aligned snapshot). Managers hold no pool — they take it as a parameter, so the same code matches both the device and host tier.
- **`KvCacheCoordinator`**: the orchestration hub. Binds the managers + pools at construction; exposes one `MatchPrefix` entry that fans out to each group, runs `Acquire`/`CacheFullBlocks`/`ReclaimExpired`/`Free`, and converges multi-group boundaries with `SweepThenConverge` (in tokens, so groups of different granularity agree).
- **Per-group block_size**: each group may carry its own block_size (constrained to a multiple of the GCD base). Content hashes are computed once at base granularity; each group folds base pages into its own coarse-block keys. This lets new hybrid models (DSv4 etc.) adopt heterogeneous granularities with zero new C++.

### Python runtime

- **Flat memory layout**: `plan_component_tensors` sizes each component tensor honestly; GDN state layers no longer allocate KV tensors (eliminates ~44% dead weight — Qwen3.5 gains ~1.77× capacity at the same budget).
- **Model bring-up**: GPT-OSS (equal-byte aliased layout) and Qwen3.5-35B (GDN hybrid, state slabs + KV slabs sharing one block-id space).

## Improvements (measured)

- **Capacity (GPT-OSS)**: equal-byte slab aliasing pairs the 12 sliding + 12 full layers onto shared slabs, giving **~2× the token pool** at the same byte budget — asserted end-to-end (slab vs forced-legacy arm, `max_total_num_tokens` ratio in [1.9, 2.1]) and behaviorally via a round-2 prefix-hit rate ≥0.9 where the halved legacy pool recycles its working set before reuse.

## Test Plan

- **Accuracy (GSM8K full set n=1319, temp 0)**: gpt-oss flat **0.9507** (measured on this branch; baseline band ~0.9484), Qwen3.5 flat **0.9363** ≈ radix 0.9265. No regression from the flat path.
- **Dual-arm C++ suite**: every commit runs radix (all-green) + flat (known-failure set unchanged), so the radix path is provably unaffected.
- **Byte-equivalence**: old vs new Qwen3.5 layout compared token-by-token (logprobs bit-identical) under eager mode, so the capacity gain carries zero accuracy cost.

## How to try it

The flat path is a separately-built scheduler ext, selected at runtime — the default (radix) build is untouched.

**Build the flat ext** (passes `TOKENSPEED_FLAT_KVCACHE=ON` through to cmake, installs to a side directory so the default radix install stays in place):

```bash
cd tokenspeed-scheduler
pip install . --target /path/to/flat_ext --upgrade --no-deps \
  --config-settings=cmake.define.TOKENSPEED_FLAT_KVCACHE=ON
# probe: should print FLAT_KVCACHE= True
PYTHONPATH=/path/to/flatsite python3 -c \
  "import tokenspeed_scheduler as t; print('FLAT_KVCACHE=', t.FLAT_KVCACHE)"
```

`/path/to/flatsite` holds a `sitecustomize.py` that front-loads the flat ext onto `sys.path` and strips the radix finder. **Setting `PYTHONPATH=…/flatsite` selects the flat arm; omitting it runs radix** — same source tree, no reinstall.

**Serve** (flat arm, via `PYTHONPATH`):

```bash
# gpt-oss-20b (default backend)
PYTHONPATH=/path/to/flatsite tokenspeed serve openai/gpt-oss-20b \
  --gpu-memory-utilization 0.6 --max-num-seqs 16 --disable-kvstore

# Qwen3.5-35B-A3B (GDN hybrid — Triton backend only, block_size 1088 is not a power of 2)
PYTHONPATH=/path/to/flatsite tokenspeed serve Qwen/Qwen3.5-35B-A3B \
  --attention-backend triton \
  --gpu-memory-utilization 0.6 --max-num-seqs 16 --disable-kvstore
```

## Known issues / WIP

- **Qwen3.5 (GDN) still rough**: prefix-cache hit rate is low and decode is limited to the Triton backend (block_size 1088 is not a power of 2, which the trtllm kernels require). Both are being addressed in the next few days.
- **DSV4 is Working in Progress, Equal-byte pool model not done**: the per-group block_size *mechanism* is in place, but under a shared pool the **cross-group dead weight** (a full-attention block still carries the state segment) is not yet reclaimed. Actually recovering it (equal-byte slabs + c128 bucketing) is the next step.
- **Two-tier cache (L2) is still working in Progress.**: paired device + host `BlockPool`s; new blocks stream down to host at end of round, host hits load back H2D — pins held by `BlockRef` on both sides for the duration of the copy.
- **Radix Cache still exists and will be removed in the future.**


nperrin-fr (approved):
LGTM. There seem to be some todo for component cache (for dsv4 and qwen3.5) but that has been noted into the future work. The SWA looks decent to me.

nayihz:
Are there any documents or materials introducing flat kv?

PR #454, opened 2026-06-15
[AMD] Support Kimi K2.5 MXFP4 serving

qedawkins:
## Summary

Add the runtime and kernel support needed to serve amd/Kimi-K2.5-MXFP4 on AMD GPUs with the current generic MLA and MoE APIs.

Kimi MXFP4 checkpoints are serialized through AMD Quark metadata rather than a plain quant_method: mxfp4 config. The runtime now recognizes the packed FP4 tensors, e8m0 scale tensors, ignored BF16 layers, W-MXFP4/A-FP8 metadata, and dynamic FP4 activation metadata directly in the MXFP4 quantization config, and mxfp4 is exposed as an explicit CLI quantization mode.

Packed dense MXFP4 linear layers are added for Kimi's dense/shared MLP weights. Those weights stay packed in VRAM, activations are quantized dynamically, and the dense layer dispatches through tokenspeed_kernel.mm(..., quant="mxfp4") instead of dequantizing checkpoint weights into BF16.

The MXFP4 MoE path now accepts precomputed topk_weights and topk_ids from the model/runtime, avoiding duplicate routing and matching the current generic MoE API. It also supports dynamic activation quantization for the packed W13 and W2 expert GEMMs, both MoE TP and EP application, and releases raw MXFP4 MoE parameters after processed kernel tensors are installed to reduce peak memory pressure.

The AMD kernel coverage grows to include fused SiLU-multiply, HIP fused RMSNorm, MXFP4 activation quantization, packed MXFP4 GEMM dispatch, MXFP4 MoE apply variants, and MLA KVStore transfer kernels used by MLA cache writeback.

The model integration selects the generic MLA backend on AMD, normalizes multi-token decode inputs, and feeds compatible tensors through the DeepSeek/Kimi MLA flow used by Kimi speculative serving. Triton MLA decode bounds its loop by the actual cache length instead of the compile-time maximum sequence length so short decode and graph capture do not scale with the configured maximum context.

Several shared runtime assumptions are hardened for local AMD serving. Optional fused LM-head execution falls back to Torch matmul when unavailable, zero-row TP logits all-gather preserves shape invariants, DeepSeek FP8 loader code no longer assumes every quantization config has FP8 block-size metadata, and DeepSeek scaled RoPE uses the common apply_rope path on AMD instead of a stale duplicate Python forward override.

## Testing Plan

test/runtime/test_mxfp4_config.py
test/runtime/test_kimi_mxfp4_dense_linear.py
test/runtime/test_logits_processor.py
test/runtime/test_cli_config_compat.py
test/runtime/test_server_args_attention_backends.py tokenspeed-kernel/test/ops/test_activation.py
tokenspeed-kernel/test/ops/test_quantization.py
tokenspeed-kernel/test/ops/test_kvcache.py
tokenspeed-kernel/test/test_kernel_api_selection.py

Additional local validation served amd/Kimi-K2.5-MXFP4 with MoE TP4 on 4 mi355x GPUs.


python/tokenspeed/runtime/layers/attention/registry.py
```diff
@@ -107,6 +107,8 @@ def register_backend(
 def _get_default_backend_name(arch: AttentionArch) -> str:
     platform = current_platform()
     if arch == AttentionArch.MLA:
+        if getattr(platform, "is_amd", False):
```
borontion: we can always default to mla backend regardless of the platform. nvidia will manually select the backend.

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/triton/mxfp4.py
borontion, line 1: I renamed the original "triton_kernels" solution to "triton" while working on #374. But looks like we are introducing it again in this PR. Can we re-use the existing `triton_mxfp4_moe_process_weights` and `triton_mxfp4_moe_apply` for kimi?

qedawkins: Done! There were still a few changes needed to support kimi (optional bias, dynamically quantized activations, precomputing top-k for now) but triton_kernels should be removed again.

python/tokenspeed/runtime/layers/attention/backends/mha.py
```diff
@@ -141,6 +141,10 @@ def init_forward_metadata(
         )
 
         if forward_mode.is_extend_or_mixed():
+            assert extend_seq_lens is not None
+            assert extend_seq_lens_cpu is not None
+            assert extend_prefix_lens is not None
+            assert extend_prefix_lens_cpu is not None
```
borontion, line 147: why change mha backend here?

qedawkins: For the draft model. For pure decode we don't need the extend attention inputs and EAGLE3 will still hit this https://github.com/lightseekorg/tokenspeed/blob/main/python/tokenspeed/runtime/execution/cuda_graph_wrapper.py#L743 (and below)

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/triton/mxfp4.py
```diff
@@ -298,54 +523,59 @@ def triton_mxfp4_moe_apply(
...
-            gemm2_input,
-            w2_weight,
-            w2_bias,
-            a_ragged_metadata=ragged_metadata,
-            scatter_indx=scatter_indx,
-            precision_config=w2_pc,
-            gammas=gate_scal,
-        )
-    return out.view(n_tokens, top_k, out.shape[-1]).sum(dim=1)
+    return _matmul(
```
borontion: nit: can we keep using the original `matmul` from `triton_kernels` instead of wrapping them into `_matmul`? the reason is that I feel these 2 gemms are quite different and there is no need to use them from the same api.

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/triton/mxfp4.py
```diff
@@ -278,18 +488,33 @@ def triton_mxfp4_moe_apply(
...
-    )
+    if topk_weights is not None or topk_ids is not None:
+        if topk_weights is None or topk_ids is None:
+            raise ValueError("topk_weights and topk_ids must be provided together")
+        topk_weights, topk_ids, num_experts = _local_topk_for_ep(
+            topk_weights,
+            topk_ids,
+            w,
+        )
+        ragged_metadata, gather_indx, scatter_indx, gate_scal = _routing_from_topk(
```
borontion, line 463: there are 2 modes for top-k: 1) compute top-k inside the `moe_apply` api (`kernel_routing`); 2) compute top-k separately and pass its result to `moe_apply` (`precomputed_topk`). 

the previous triton version `moe_apply` defaults to mode 1. but we are using mode 2 here. is there any specific reason we are using mode 2? is that something needed to support EP?

qedawkins: https://github.com/lightseekorg/tokenspeed/blob/main/python/tokenspeed/runtime/models/deepseek_v3.py#L305 needs grouped top-k, routed scaling, and EP expert remapping.  The plan was to fuse in a follow up (basically would need to compute the global top-k in every rank then mask based on the local rank).

borontion (approved):

borontion:
> may we add the kimi e2e ut on mi35x

We will add one in the next PR.

python/tokenspeed/runtime/layers/attention/registry.py
```diff
@@ -105,13 +103,8 @@ def register_backend(
...
 
 def _get_default_backend_name(arch: AttentionArch) -> str:
-    platform = current_platform()
     if arch == AttentionArch.MLA:
-        if platform.is_blackwell:
-            return "trtllm_mla"
-        if platform.is_hopper:
-            return "flashmla"
-        return "trtllm_mla"
+        return "mla"
```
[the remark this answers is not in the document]
borontion: The mla backend is just like mha backend - it will route to different kernels. For now we only add triton support. Later we will add flashmla, tokenspeed-mla and also gluon kernels. It should be a default backend to auto-select the best kernel.

For now, it seems we always specify the attention backend on nvidia side? It is probably fine to leave the default to be mla? Or do you think we should keep different defaults based on devices?

python/tokenspeed/runtime/layers/attention/backends/mla.py
```diff
@@ -302,7 +302,7 @@ def forward_decode(
         num_extends = metadata.num_extends
         q_len_per_req = q.shape[0] // bs if bs > 0 else 1
 
-        if q_len_per_req > 1 and self.is_draft:
+        if q_len_per_req > 1:
```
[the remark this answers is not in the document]
qedawkins: Target verification will still run with > 1 token but won't be the draft model and we still need to do the flatten + duplication of the page table/seqlen metadata in that case. Basically the fact that we're processing a draft model isn't special here.

PR #456, opened 2026-06-15
perf(kernel): optimize Qwen vision QKV rotary layout

qimcis:
## Summary

+7.38% average speedup for encoder

Previously, `VisionAttention.forward()` split packed QKV and forced Q, K, and V to become contiguous tensors:

```python
q = q.reshape(bsz * s, head, -1).contiguous()
k = k.reshape(bsz * s, kv_head, -1).contiguous()
v = v.reshape(bsz * s, kv_head, -1).contiguous()
```

That made three device copies per vision attention layer, qwen vision has 27 layers, so this was 81 Q/K/V layout copies per encoder forward

This pr makes two optimizations for this, first by removing the explicit Q/K/V .contiguous() copies by keeping Q/K/V as views after the packed-QKV split then adding a packed-QKV rotary triton kernel for Qwen RoPE

## Test Plan

accuracy was validated across backends;

Encoder only results:
Qwen3.5-397B-A17B-NVFP4
backend: FA4

Item 1: Remove Q/K/V `.contiguous()` materialization after packed QKV split

| | Before Avg ms | After Avg ms | Speedup |
|---|---:|---:|---:|
| grid=96,batch=32 | 760.27 | 725.83 | +4.75% |
| grid=128,batch=32 | 1709.16 | 1693.18 | +0.94% |

Item 2: Add packed-QKV + rotary Triton kernel on top of item 1

| | Before Avg ms | After Avg ms | Speedup |
|---|---:|---:|---:|
| grid=96,batch=8 | 191.99 | 177.01 | +8.46% |
| grid=96,batch=32 | 750.61 | 690.60 | +8.69% |
| grid=128,batch=8 | 438.65 | 412.62 | +6.31% |
| grid=128,batch=32 | 1769.53 | 1667.77 | +6.10% |

E2E results:
Qwen3.5-397B-A17B-NVFP4
backend: FA4
TP8
8xb300

4 images x 1024, concurrency 8

| | Before Avg | After Avg | Delta |
|---|---:|---:|---:|
| Requests/sec | 1.269 | 1.270 | +0.10% |
| Tokens/sec | 324.917 | 325.230 | +0.10% |
| p50 TTFT | 4094.646 ms | 3795.509 ms | +7.88% faster |
| p50 total latency | 6046.911 ms | 6007.632 ms | +0.65% faster |

1 image x 2048, concurrency 8

| | Before Avg | After Avg | Delta |
|---|---:|---:|---:|
| Requests/sec | 1.236 | 1.225 | -0.90% |
| Tokens/sec | 316.089 | 313.215 | -0.91% |
| p50 TTFT | 4375.853 ms | 4696.134 ms | -6.82% slower |
| p50 total latency | 6288.336 ms | 6361.221 ms | -1.15% slower |

slight regression for higher resolution image on average

### DIFF
--- a/python/tokenspeed/runtime/layers/attention/mm_encoder_attention.py
+++ b/python/tokenspeed/runtime/layers/attention/mm_encoder_attention.py
@@ -64,7 +64,10 @@ if _is_nvidia:
     )
 
 from tokenspeed_kernel.ops.attention.triton.context import context_attention_fwd
-from tokenspeed_kernel.ops.attention.triton.qkv_rotary import packed_qkv_complex_rotary
+from tokenspeed_kernel.ops.attention.triton.qkv_rotary import (
+    packed_qkv_complex_rotary,
+    packed_qkv_neox_rotary,
+)
 
 # CUDA-graph bucketing for the cuDNN vision prefill backend: batch and max
 # seqlen are quantized so a small set of captured graphs covers the request
@@ -382,6 +385,11 @@ class VisionAttention(nn.Module):
             self._backend_fn is vision_attn_fa4
             and self.position_embedding_mode == "complex_rope"
         )
+        self._use_packed_qkv_rotary = (
+            self._backend_fn is vision_attn_fa4
+            and self.customized_position_embedding_applier is None
+        )
+        self._copy_v_after_packed_qkv_rotary = False
         self._workspace_buffer = workspace_buffer
 
         self.qkv_proj = QKVParallelLinear(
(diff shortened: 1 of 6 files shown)


qimcis:
cc @chenht2022

python/tokenspeed/runtime/layers/attention/mm_encoder_attention.py
```diff
@@ -426,27 +432,52 @@ def forward(
         )
 
         qkv, _ = self.qkv_proj(x)
-        q, k, v = qkv.split([self.q_size, self.kv_size, self.kv_size], dim=-1)
-
-        q = q.reshape(bsz * s, head, -1).contiguous()
-        k = k.reshape(bsz * s, kv_head, -1).contiguous()
-        v = v.reshape(bsz * s, kv_head, -1).contiguous()
```
chenht2022: For backends other than fa4 that use fallback branch, would deleting these `.contiguous()` cause any correctness issues?

chenht2022:
BTW, why is there a regression on high-res image workload? Is this just a jitter?

qimcis:
> BTW, why is there a regression on high-res image workload? Is this just a jitter?

it's just jitter in e2e runs as far as i'm aware, the encoder improvements stay consistent

python/tokenspeed/runtime/layers/attention/mm_encoder_attention.py
qimcis: i tested for all backends other than fa4 and found that there were no correctness issues~

chenht2022 (approved):

qimcis:
ran into a couple issues with the ut i added with `Qwen/Qwen3.5-35B-A3B-FP8`
1. Loading this model failed because a GDN projection shard is 32 wide but fp8 block quant (`--quantization fp8`) requires 128-wide blocks; i fixed so now fp8 disables quant for GDN linear attention (which nvfp4 already does)

2. Qwen3.5 FP8 MoE failed because FlashInfer TRTLLM FP8 MoE was hardcoded to DeepSeekV3 routing, but Qwen3.5 uses `RenormalizeNaive`, fixed to read the proper config

3. The CI failure was because log suppression for FA only caught `ImportError`, but the import raised a runtime exception

can you re-review? cc @chenht2022

PR #470, opened 2026-06-17
perf(kernel): decode at medium batch sizes

panditsa:

### DIFF
--- a/tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/moe/fused_mxfp_gfx950.py
+++ b/tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/moe/fused_mxfp_gfx950.py
@@ -3651,6 +3651,21 @@ def _run_moe_tile_ncontig_w(
 
 
 @gluon.jit
+def _moe_masked_store(out, y_ptr, y_offs, mask, USE_BUFFER_STORE: gl.constexpr):
+    """Shared masked store for the prefill and medium-decode epilogues.
+
+    ``USE_BUFFER_STORE`` selects the AMD ``buffer_store`` intrinsic (the medium
+    dispatch path's fast store) vs generic ``gl.store`` (prefill + medium
+    combine). The caller owns all address/mask computation so each path keeps
+    its own addressing; only the final emit is shared.
+    """
+    if USE_BUFFER_STORE:
+        gl.amd.cdna4.buffer_store(out, y_ptr, y_offs.to(gl.int32), mask=mask)
+    else:
+        gl.store(y_ptr + y_offs, out, mask=mask)
+
+
+@gluon.jit
 def _pipelined_moe_tile_compute(
     # Tensors --------------------------------------------------------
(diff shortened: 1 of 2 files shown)


tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/moe/fused_mxfp_gfx950.py
```diff
@@ -4184,8 +4188,453 @@ def _group_m_swizzle(
...
+    gl.assume(stride_xm >= 0)
+    gl.assume(stride_xk >= 0)
+    gl.assume(stride_we >= 0)
+    gl.assume(stride_wn >= 0)
+    gl.assume(stride_wk >= 0)
+    gl.assume(stride_wse >= 0)
+    gl.assume(stride_wsn >= 0)
+    gl.assume(stride_wsk >= 0)
+    gl.assume(stride_ym >= 0)
+    gl.assume(stride_yn >= 0)
```
borontion, line 4424: what's the effect of these assume?

antiagainst: They establish the ranges for integer range analysis to help deducing that we can safely use buffer load for `tl.load`, if not programming directly with `gl.buffer_load`. Also these assumes help to simplify codegen down the road like [this](https://github.com/triton-lang/triton/blob/main/third_party/amd/lib/TritonAMDGPUTransforms/FoldTrueCmpIOp.cpp) and also llvm can use it simplify some logic down the stack

antiagainst (approved):

borontion (approved):

antiagainst (changes requested):
It looks this now completely reverts https://github.com/lightseekorg/tokenspeed/commit/b4333e4d13a33a3a0c4c55995dcb382b94d36fcb? Do we need to? Is it possible to land progressively and maybe just disable if that's causing issues rather than revert entirely?

antiagainst:
I took your older branch and tried to fix the correctness issue with https://github.com/lightseekorg/tokenspeed/pull/494 as incremental steps. :)

panditsa:
> It looks this now completely reverts [b4333e4](https://github.com/lightseekorg/tokenspeed/commit/b4333e4d13a33a3a0c4c55995dcb382b94d36fcb)? Do we need to? Is it possible to land progressively and maybe just disable if that's causing issues rather than revert entirely?

Hi @antiagainst  that was a bad merge, working on a fix locally. Will push it sometime later today. Fundamentally these two changes conflict although they both target different things, prefill vs. decode. I tried using the preshuffling + padding on the decode path but it regresses performance.

antiagainst:
I see; makes sense. Thanks for fixing it! While iterating on https://github.com/lightseekorg/tokenspeed/pull/494, I do find an issue for the current main, fixing it with https://github.com/lightseekorg/tokenspeed/pull/495.

antiagainst (approved):

PR #509, opened 2026-06-24
Fix: GLM5 DSA MTP TP4 IMA Error

FlamingoPg:
## Summary

Fixes the GLM5 DSA + MTP TP4 illegal memory access seen under the ShareGPT 128/16 workload by keeping draft decode metadata aligned to the accepted-prefix view used by later MTP steps.

Changes include:

- keep CUDA graph draft metadata aliased to the drafter-owned `draft_seq_lens_buf`, with the seq-lens prep kept inline at the call sites
- let GLM5 NextN correct draft seq lens after the first MTP step so rejected verify-tail KV is hidden from later draft steps
- move GLM DSA MTP top-k row selection into `glm5_nextn.py`, leaving `Eagle` to call model/backend hooks instead of knowing GLM DSA internals
- keep draft metadata advancement as an optional backend hook, matching the DeepSeek V4 style
- add the DSA backend hook to refresh sparse decode metadata and schedule state when draft seq lens changes

## Root Cause

The TP4 path let GLM5 NextN draft metadata drift from the accepted-prefix view used by later MTP decode steps. DSA top-k/window metadata and sparse decode schedule state could be computed from stale or mismatched row counts, while the first draft step still needed to fill KV from the wider verified window. That mismatch eventually surfaced as an illegal memory access in the TRTLLM MoE path.

## Validation

- DeepSeek metadata regression tests around the optional draft metadata hook
- Exact ShareGPT repro on 4 GPUs with the reported 128/16 workload shape:
  - `#Input tokens: 42015`
  - `#Output tokens: 32768`
  - `Successful requests: 128`
  - `Failed requests: 0`

### DIFF
--- a/python/tokenspeed/runtime/layers/attention/backends/dsa.py
+++ b/python/tokenspeed/runtime/layers/attention/backends/dsa.py
@@ -273,11 +273,7 @@ class DSABackend(AttentionBackend):
         self._refresh_decode_topk_schedule_metadata(bs)
         return result
 
-    def _refresh_decode_topk_schedule_metadata(
-        self,
-        bs: int,
-        seq_lens_offset: int = 0,
-    ) -> None:
+    def _refresh_decode_topk_schedule_metadata(self, bs: int) -> None:
         if deep_gemm is None:
             return
         metadata = getattr(self._dense_backend, "forward_decode_metadata", None)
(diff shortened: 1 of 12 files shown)


python/tokenspeed/runtime/layers/attention/backends/dsa.py
```diff
@@ -354,12 +345,15 @@ def init_forward_metadata(
...
         out = self._dense_backend.init_forward_metadata(
             bs=bs,
             num_extends=num_extends,
             req_pool_indices=req_pool_indices,
             seq_lens=seq_lens,
-            forward_mode=forward_mode,
+            forward_mode=dense_forward_mode,
             req_to_page=req_to_page,
             spec_info=spec_info,
             **kwargs,
```
syuoni:
```suggestion
        out = self._dense_backend.init_forward_metadata(
            bs=bs,
            num_extends=num_extends,
            req_pool_indices=req_pool_indices,
            seq_lens=seq_lens,
            forward_mode=forward_mode,
            req_to_page=req_to_page,
            spec_info=spec_info,
            **kwargs,
```

syuoni (approved):

syuoni (approved):

PR #539, opened 2026-06-27
chore(kernel): bump flashinfer to 0.6.13, add <cfloat> include

jaywme:
Update flashinfer-python/flashinfer-cubin to 0.6.13 in cuda.txt and add the missing <cfloat> include (FLT_MAX) to silu_and_mul_fuse_nvfp4_quant.cu so it builds against the new flashinfer headers.


jaywme:
pending because ci workers still uses flashinfer-jit-cache==0.6.11.post3

PR #544, opened 2026-06-28
 refactor(spec-decode): simplify deepseekV3/GLM attention path for #217 (3/3)

rjzhb:
## Summary
Replace the `draft_first_step_reduce` flag with a query-slice in the spec-decode
drafters. On the active first draft step the full `latent_cache` still writes every
KV row, but only the live rows (`gather_ids`) run the absorbed decode, narrowing the
output to `[bs, H]`. `accept_lengths` / `gather_ids` become the single narrow signal.

## Changes
- New shared `DeepseekV3DraftAttentionMLA._attn` (used by the NextN and Eagle3-MLA
  drafters): narrow `Q` to live rows via `gather_ids`; `_apply_correction` trims
  decode `cache_seqlens` by `spec_num_tokens - accept_lengths`.
- Wrap the first-step absorbed decode in `override_num_extends(0)` so the decode
  kernel's `page_table` / `seq_lens` span the full `bs` (matching the narrowed
  query) — fixes `Mismatched page_table.shape[0]` on MIXED first steps
  (`num_extends > 0`); mirrors the existing multi-step drafter loop.
- Drop `draft_first_step_reduce` from `eagle.py` and the
  deepseek-nextn / glm5-nextn / qwen3.5-nextn / llama-eagle3 drafters.


python/tokenspeed/runtime/models/deepseek_v3.py
```diff
@@ -1153,7 +1229,7 @@ def __init__(
         rope_scaling = getattr(config, "rope_scaling", None)
         max_position_embeddings = getattr(config, "max_position_embeddings", 8192)
 
-        self.self_attn = DeepseekV3AttentionMLA(
+        self.self_attn = self.ATTENTION_CLS(
```
nperrin-fr: accessing class attribute from self is weird to me. can we define
```
@property
def attention_cls(self) -> type[nn.Module]:
    return DeepseekV3AttentionMLA
```

nperrin-fr (approved):
others look fine to me

python/tokenspeed/runtime/models/deepseek_v3.py
rjzhb: Thanks for reviewing! Have fixed it.

PR #545, opened 2026-06-28
fix(sampling): CuteDSL argmax crashes the server on empty-row input

rjzhb:
## Summary
**Impact:** The server **crashes during startup** — the warmup request triggers
an idle data-parallel rank's draft forward, which runs argmax over a
`(0, vocab)` logits tensor. The CuteDSL argmax kernel launches a 0-block grid
over the vocab tiles and raises `CUDA_ERROR_INVALID_VALUE`, killing the
scheduler; the remaining ranks then time out on the DeepGEMM NVLink barrier and
the whole serve goes down. Only reproduces at a real vocab size — `(0, 128)`
survives, `(0, 163840)` does not. 

**Fix:** Add `shape[0] == 0` to the existing torch-fallback guard in `argmax`
and `_argmax_cute`, so empty input goes through `torch.argmax` (already the path
for other unsupported inputs). Non-empty inputs unchanged.

### DIFF
--- a/tokenspeed-kernel/python/tokenspeed_kernel/ops/sampling/__init__.py
+++ b/tokenspeed-kernel/python/tokenspeed_kernel/ops/sampling/__init__.py
@@ -85,7 +85,16 @@ def argmax(
     Returns:
         A tensor containing argmax indices for each row of ``logits``.
     """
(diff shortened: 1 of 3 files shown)


LorrinWWW (approved):
we can remove the comment or keep it concise

PR #558, opened 2026-06-30
Optimize GLM DSA decode graph contexts

FlamingoPg:
## Summary
- introduce backend-declared `CudaGraphProfile`s for CUDA graph capture/replay
- keep the default full-context graph as the safe fallback, and let attention backends declare/select any extra profiles they can use efficiently
- let GLM DSA request smaller decode-context profiles (`2048`, `3072`, `4096`) through the generic backend profile hook instead of hard-coding GLM policy in `CudaGraphWrapper`
- propagate a CPU-side `decode_seq_len_upper_bound` through `ForwardContext` so graph dispatch does not need to sync GPU `seq_lens`
- store TRTLLM MLA graph metadata/block-table buffers by backend profile context length and keep draft/target block-table aliasing compatible
- add a small DeepGEMM symbol-name compatibility fallback

## Root Cause
GLM 5.1 NVFP4 MTP decode was replaying decode graphs captured against the full max context length. On short/mid ShareGPT decode workloads this pays the long-context block-table/metadata cost even when the actual decode seq-len is much smaller.

## Design
This PR keeps the CUDA graph wrapper general:

- `CudaGraphProfile()` is the default/full-context fallback profile.
- `AttentionBackend.get_cuda_graph_profiles()` declares extra profiles.
- `AttentionBackend.select_cuda_graph_profile()` owns backend-specific profile selection.
- `CudaGraphWrapper` only enumerates, captures, validates, and replays profiles; it does not know GLM/DSA policy.
- `CudaGraphProfile.forward_context_fields` lets a backend make capture-time `ForwardContext` match the profile it declared.
- GLM DSA declares profiles derived from `index_topk`; TRTLLM MLA consumes the profile metadata to size its CUDA graph metadata/block-table state.

That keeps the performance optimization local to the backend that benefits from it, while leaving the runtime path usable for other backends that may want different graph profiles later.

## Performance
Benchmark setup:
- Model: `nvidia/GLM-5.1-NVFP4`
- Hardware/parallelism: 4x B200, `attn_tp_size=4`, `moe_tp_size=4`
- Workload: ShareGPT, first 128 prompts, max concurrency 16, output len 256, `temperature=0`
- Input/output shape check: `#Input tokens: 42015`, `#Output tokens: 32768`

Results:

| Version | Output tok/s | Total tok/s | Mean TPOT |
| --- | ---: | ---: | ---: |
| `origin/main` baseline | 508.40 | 1171.83 | 28.74 ms |
| This PR, first run | 642.64 | 1483.33 | 22.11 ms |
| This PR, hot repeat | 739.17 | 1704.40 | 19.69 ms |

Hot-repeat delta vs baseline:
- Output throughput: +45.4%
- Total throughput: +45.4%
- Mean TPOT: -31.5%

## Validation
- `python -m py_compile` on the modified runtime/backend files
- `pytest test/runtime/test_max_new_tokens_context_cap.py test/runtime/test_trtllm_spec_seqlen_clamp.py -q`
- GLM 5.1 NVFP4 4-card ShareGPT 128/16 benchmark described above; captured `default`, `decode_ctx_2048`, `decode_ctx_3072`, and `decode_ctx_4096` profiles successfully

### DIFF
--- a/python/tokenspeed/runtime/engine/event_loop.py
+++ b/python/tokenspeed/runtime/engine/event_loop.py
@@ -648,6 +648,9 @@ class EventLoop:
             dp_metadata.all_decode_or_idle if dp_metadata is not None else False
         )
         multimodal_context = self._get_multimodal_context_for_forward(forward_op)
+        decode_seq_len_upper_bound = self._get_decode_seq_len_upper_bound_for_forward(
+            forward_op
+        )
 
         self.model_executor.update_block_table(forward_op)
 
@@ -663,6 +666,7 @@ class EventLoop:
                     dp_all_decode_or_idle=dp_all_decode_or_idle,
                     grammar_inputs=grammar_inputs,
                     multimodal_context=multimodal_context,
+                    decode_seq_len_upper_bound=decode_seq_len_upper_bound,
                     **stats,
                 ),
                 None,
@@ -692,6 +696,7 @@ class EventLoop:
                         dp_global_bs=dp_global_bs,
                         dp_all_decode_or_idle=dp_all_decode_or_idle,
                         multimodal_context=multimodal_context,
+                        decode_seq_len_upper_bound=decode_seq_len_upper_bound,
                         **stats,
                     ),
                     None,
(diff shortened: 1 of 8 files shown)


FlamingoPg:
Closing this draft. The CUDA graph context-specialization experiment is superseded by the cleaner kernel/backend ragged top-k PR in #573.

PR #574, opened 2026-07-01
Fix DP cache sizing

tuanzhangCS:
Summary:
- Cap KVStore host sizing by cgroup-aware per-rank memory budget.
- Use rank-local DP padding rows and safe cache/gather indexing.

Test:
- pre-commit run --all-files

### DIFF
--- a/python/tokenspeed/runtime/execution/drafter/eagle.py
+++ b/python/tokenspeed/runtime/execution/drafter/eagle.py
@@ -146,6 +146,30 @@ class Eagle(BaseDrafter):
             getattr(hf_config, "index_share_for_mtp_iteration", False)
         )
 
+    def _accepted_output_indices(
+        self,
+        accept_lengths: torch.Tensor,
+        row_count: int,
+        *,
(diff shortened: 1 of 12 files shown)


python/tokenspeed/runtime/execution/drafter/eagle.py
```diff
@@ -146,6 +146,30 @@ def __init__(
...
+        than producing ``row * N - 1``.
+        """
+        safe_accept_lengths = (
+            accept_lengths[:row_count].to(torch.int64).clamp(1, self.spec_num_tokens)
+        )
+        return (
+            self.padded_gather_ids_offsets_buf[:row_count]
+            + safe_accept_lengths
+            + base_offset
+        )
```
syuoni, line 171: Hi @tuanzhangCS 

`accept_lengths` should follow a contract that its values are in `[1, self.spec_num_tokens]`. If somewhere produces values against this contract, we should fix that very place, instead of adding a patch on the common (and hot) path.

PR #583, opened 2026-07-03
perf(deepseek-v4): enable MTP overlap scheduling with paged cache

dongjiyingdjy:
## Summary

- Enable the existing overlap scheduler for DeepSeek V4's speculative paged-cache groups.
- Propagate the runtime verify width and one-step overlap depth through scheduler admission, page-table reservations, retraction recovery, CUDA Graph sizing, and V4 attention metadata.
- Keep the already-dispatched speculative tail protected while publishing snapshots only at the accepted token boundary.
- Fix mixed prefill/decode MTP metadata so target and draft step 0 use the runtime verify width.

## Reservation contract

Let `W = decode_input_tokens` and `D = overlap_schedule_depth` (`D` is currently restricted to 0 or 1). A request whose committed token length is `C` reserves through `C + (D + 1) * W`. When a result accepts through `A`, the table rewinds only to `A + D * W`, preserving the next forward that is already in flight. Prefix-cache snapshots still publish only the real accepted/checkpoint boundary, not the reservation horizon.

## Validation

- Targeted runtime/scheduler Python tests: 141 passed, 1 skipped, plus 145 subtests
- Scheduler ASAN/LSAN suite: 216 passed (1 disabled)
- DeepSeek V4-Flash deterministic GSM8K, 5-shot/50 questions, concurrency 1, temperature 0: strict and flexible exact match both 0.96 (required >= 0.92); all four ranks reported `W=4, D=1`
- V4-Pro SWE-smith matched overlap-off/on endpoint run: 350/350 requests succeeded on each side, with no C4/page-allocation/OOM/runtime fatal
  - TPS/GPU: p1 -3.19%, p2 +5.36%, p4 +15.43%, p8 -1.64%
  - TPOT: p1 -12.16%, p2 -19.00%, p4 -19.69%, p8 +4.45%
- Matched p4 Nsight Systems capture with CUDA Graph nodes and OSRT:
  - post-sync commit work overlapped the current graph in 0/768 off steps and 824/824 on steps
  - median inter-replay GPU bubble decreased from 7.197 ms to 0.670 ms

## Scope and limitations

- This change supports one in-flight overlap step only (`D <= 1`). A deeper pipeline would require additional table-versioning or page-lifetime work.
- Endpoint performance is intentionally reported as mixed: p2/p4 improve, while p1 has a TTFT tradeoff and p8 regresses slightly. Follow-up profiling is needed for the p1 prefill/TTFT path and p8 saturation; this PR does not claim a whole-curve performance win.

### DIFF
--- a/python/tokenspeed/runtime/configs/paged_cache_spec.py
+++ b/python/tokenspeed/runtime/configs/paged_cache_spec.py
@@ -47,6 +47,8 @@ def compute_paged_cache_group_page_counts(
     max_scheduled_tokens: int,
     max_total_tokens: int,
     max_context_len: int,
+    decode_input_tokens: int = 1,
+    overlap_schedule_depth: int = 0,
     safety_margin: int = 0,
 ) -> dict[str, int]:
     if max_live_requests < 0:
@@ -59,6 +61,16 @@ def compute_paged_cache_group_page_counts(
         raise ValueError(f"max_total_tokens must be >= 0, got {max_total_tokens}")
     if max_context_len < 0:
         raise ValueError(f"max_context_len must be >= 0, got {max_context_len}")
+    if decode_input_tokens < 0:
+        raise ValueError(f"decode_input_tokens must be >= 0, got {decode_input_tokens}")
+    if overlap_schedule_depth not in (0, 1):
+        raise ValueError(
+            f"overlap_schedule_depth must be 0 or 1, got {overlap_schedule_depth}"
+        )
+    if overlap_schedule_depth > 0 and decode_input_tokens == 0:
+        raise ValueError(
+            "overlapped paged-cache sizing requires decode_input_tokens > 0"
+        )
     if safety_margin < 0:
         raise ValueError(f"safety_margin must be >= 0, got {safety_margin}")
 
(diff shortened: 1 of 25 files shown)


dongjiyingdjy:
CI follow-up after rebasing onto `d8e54692` and updating the PR head to `cc601d549`:

Resolved checks:
- lint: passed, including scheduler C++ clang-format and clang-tidy
- DeepSeek V4 Flash GSM8K: `0.9469` >= `0.90`
- DeepSeek V4 Flash MTP GSM8K: `0.9500` >= `0.90`
- Qwen3.5 DP4EP4 AIME25: `0.9000` >= `0.86`
- Kimi K2.5 EAGLE3 B200 AIME25: `0.9333` >= `0.93`

The only remaining failures are pre-existing Qwen agentic perf-reference gates:
- B200 8-GPU p1 TPS/user: `429.18 < 522`; the prior PR583 run was `434.78`, and the direct base PR582 run was `431.03`, so this is the same baseline failure.
- GB200 4-GPU: p4/p8 throughput and TPS/user plus p16 TPS/user miss the historical floors. The direct base and adjacent main runs fail the same job with the same concurrency pattern. All requests completed successfully.

Final check summary: 44 success, 2 known perf-reference failures, 3 skipped, 0 pending. All three workflow `finish` jobs passed.

PR #584, opened 2026-07-03
fix(output): prevent prompt-tail tokens leaking into streamed output_ids

minedec:
## Summary

The inline-detokenizer branch of OutputProcessor incorrectly extended `state.output_ids` with `recv_obj.decode_ids[i]`, which includes a 5-token prompt-tail context window (INIT_INCREMENTAL_DETOKENIZATION_OFFSET) needed by the incremental detokenizer. This caused the first 5 tokens of every streamed response to echo the prompt suffix (e.g. <|im_start|>assistant\n<think>\n) as if they were freshly sampled output. Changed to `recv_obj.output_ids[i]` to match the BatchStrOut and raw-token branches, which already use the correct field containing only actually-sampled tokens.


minedec:
> Hi @minedec may you open a new one Thanks!

ok

PR #586, opened 2026-07-03
perf(glm-5.2): drop full-topk path and hadamard transform, use flashinfer LayerNorm for DSA decode path

syuoni:
## Summary

This PR trims dead code and per-decode-step kernels from the DSA lightning-indexer path, and swaps the one torch norm for a fused kernel. Behavior-preserving except the Hadamard drop, which is accuracy-verified (below).

### Changes
- **Remove the dead full-context decode top-k fast path** (`glm5.py`). The static-shape proxy gating it is always false under CUDA graph (no D2H sync), so it never fired for long context; it only added redundant index-math kernels and a double top-k probe.
- **Indexer `k_norm`: `nn.LayerNorm` → flashinfer `LayerNorm`.** New `LayerNorm` layer keeps fp32 affine params (flashinfer requires it; matches the reference). Threaded through the `tokenspeed_kernel` boundary.
- **Drop the indexer Hadamard** (`_glm_dsa_hadamard_rotate_pair`). It's an orthogonal, dot-product-preserving rotation whose only purpose is fp8-quant robustness; **e2e accuracy (AIME26) confirmed unchanged**. Aligns with vLLM and TRT-LLM, which run fp8 DSA without it.
- **Remove dead `forward_cuda` hooks** (`glm5.py`, `deepseek_v3.py`, `deepseek_v4.py`). Nothing defines `forward_cuda`; the base `RotaryEmbedding.forward` is already the `apply_rope` path, so the `hasattr` swap never executed.


python/tokenspeed/runtime/models/glm5.py
```diff
@@ -32,10 +32,6 @@
     dsa_plan,
     dsa_prefill_topk,
 )
-from tokenspeed_kernel.ops.attention.triton.dsa_sparse_layout import (
```
borontion, line 35: can we also delete `tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/triton/dsa_sparse_layout.py`?

PR #592, opened 2026-07-04
[AMD] Initial support for qwen3.5-397b-a17b-mxfp4 on AMD hardware

raikonenfnu:
## Summary

This branch brings the AMD Qwen3.5 MXFP4 bringup path closer to production readiness. It fixes the Qwen3.5 quantization exclusions, keeps the known-bad dynamic MXFP4 MoE `ispp=256` shape on the stable Triton fallback while leaving Gluon enabled for safe shapes, routes GDN chunk prefill through the tokenspeed-kernel registry, and addresses two distributed/Iris synchronization issues seen at larger tensor-parallel sizes. It also adds CI eval coverage and a public GDN op test.

### `fix(quant): keep Qwen3.5 GDN projections bf16 under MXFP4`

Qwen3.5 MXFP4 checkpoints leave the GDN linear-attention projections in bf16, while runtime fuses the input projection into a single `in_proj_qkvzba` module. Without a fused-name ignore rule, the loader tried to MXFP4-pack that bf16 module and hit a checkpoint shape mismatch. This commit teaches quantization filtering to ignore the fused module only when all source shards are ignored, and adds alias normalization for `model.language_model.*` checkpoint excludes that runtime sees as `model.*` module names.

### `fix(comm): add trailing barrier to Iris fused AR+RMSNorm`

Iris fused AR+RMSNorm reuses a symmetric-heap `_input_buf` across calls. The existing leading barrier ensures every rank publishes its input before peer reads begin, but did not ensure every peer finished reading before a faster rank overwrote the buffer in the next call. This could corrupt the all-reduce input under the overlap scheduler and lead to wrong output or decode hangs. This commit adds the trailing barrier needed to protect `_input_buf` reuse.

### `fix(distributed): bind init_process_group to rank device`

At higher TP, Iris symmetric-heap setup uses distributed object collectives to exchange fd/IPC handles. Without an explicit `device_id`, NCCL guesses each rank's device from global rank, which is unsafe when global rank and local GPU assignment diverge. This commit threads the assigned `gpu_id` into `ProcessGroupManager.init_distributed()` and binds the default process group to `torch.device(device, gpu_id)`.

### `adding e2e eval`

Adds Qwen3.5 397B MXFP4 evalscope CI configs for AIME25 and GSM8K on AMD MI35x TP4, plus `tokenspeed-kernel/test/ops/test_attention_gdn.py`. 

## Test Plan

- Added e2e accuracy eval test for TP=4 qwen3.5-397b-a17b-mxfp4 to run on GSM8k and AIME

### DIFF
--- a/python/tokenspeed/runtime/distributed/process_group_manager.py
+++ b/python/tokenspeed/runtime/distributed/process_group_manager.py
@@ -22,6 +22,7 @@
 
 from datetime import timedelta
 
+import torch
 import torch.distributed as dist
 
 from tokenspeed.runtime.distributed.mapping import Group, Mapping
@@ -52,6 +53,7 @@ class ProcessGroupManager:
         distributed_init_method: str = "env://",
         backend: str = "nccl",
         timeout: int | None = None,
+        device_id: "torch.device | None" = None,
     ) -> None:
         if not dist.is_initialized():
             if distributed_init_method is None:
@@ -71,6 +73,7 @@ class ProcessGroupManager:
                 world_size=mapping.world_size,
                 rank=mapping.rank,
                 timeout=timeout,
+                device_id=device_id,
             )
 
     def register_process_group(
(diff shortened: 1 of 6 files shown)


antiagainst (approved):

raikonenfnu:
> @raikonenfnu The Qwen-related CIs failed. Could you help check whether this is flaky or introduced by this PR? Thanks!

Yeah, I am seeing 0.875 locally, will check what is happening and make sure accuracy is good before landing. :)

PR #594, opened 2026-07-06
fix(perf): fix qwen3.5 tp8 agentic ci perf regression

minedec:
## Summary
Commit a6578e5 ("fix(dflash): fix server hang issue") introduced DFLASH-specific safety logic that ran unconditionally on all speculative algorithms, causing ~23% decode TPS regression (580→446) on Qwen3.5 TP8 MTP.

### Changes Causing Regression
- model_executor.py: Inflated max_num_pages_per_req with spec_num_tokens * 64 slack, enlarging page-table buffers and CUDA graph metadata for every spec model.
- model_executor.py: Added _cap_accept_to_context_len and _clamp_committed_to_context_len on every decode step, creating fresh tensors that force the slow two-D2H fallback path.
- cuda_graph_wrapper.py: Changed padding from zero (slot 0) to sentinel (max_req_pool_size), adding per-step overhead for non-DFLASH models that don't need it.
- event_loop.py: Replaced the fast local short-circuit with a collective all_reduce on every event-loop iteration across all TP ranks.

### Fix
   - Guard all four changes behind self.config.spec_algo == "DFLASH", restoring the original fast paths for MTP/EAGLE3 while preserving DFLASH correctness.

NOTE: This is a temporary fix — DFLASH-specific perf issues will be addressed in a follow-up PR.

### DIFF
--- a/python/tokenspeed/runtime/engine/event_loop.py
+++ b/python/tokenspeed/runtime/engine/event_loop.py
@@ -535,11 +535,17 @@ class EventLoop:
         # _pending_cache_event_payloads) diverges transiently. A rank-local skip
         # would let some ranks gather while others return, deadlocking the group.
         # Agree on the skip via a cheap single-int all_reduce.
+        # NOTE: For non-DFLASH algorithms, cache ops are deterministic across
+        # ranks, so the local short-circuit is safe and avoids collective overhead.
         local_has_work = bool(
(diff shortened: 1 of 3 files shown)


yweng0828 (approved):

yweng0828:
Thanks for the fix!

Issue #589, opened 2026-07-04
RFC: route GDN prefill through tokenspeed-kernel registry


raikonenfnu:
## Summary

This RFC documents the design we chose for making GDN chunk prefill registry-selected in `tokenspeed-kernel`, and how runtime handles the different checkpoint layouts exposed by the Triton/FLA and FlashInfer implementations.

The high-level direction is the same pattern used by MHA: runtime calls a public `tokenspeed-kernel` op, while concrete implementations register under a shared operator family/mode and are selected by platform, dtype, and traits.

```text
MambaAttnBackend / HybridLinearAttnBackend
  -> tokenspeed_kernel.ops.attention.gdn_chunk_prefill(...)
      -> select_kernel("attention", "gdn_chunk_prefill", traits=...)
          -> triton_gdn_chunk_prefill or flashinfer_gdn_chunk_prefill
```

## Motivation

Before this change, `MambaAttnBackend` owned too much backend choice logic for GDN prefill. It directly knew about the FlashInfer Blackwell fast path and the portable Triton/FLA fallback. That made GDN unlike MHA, where runtime calls a stable public op and `tokenspeed-kernel` owns solution selection.

Moving GDN chunk prefill into the registry gives us:

- vendor-neutral runtime wiring;
- a portable Triton fallback for AMD/non-Blackwell;
- a FlashInfer fast path on supported NVIDIA Blackwell shapes;
- a future extension point for Gluon/CuteDSL without touching runtime call sites.

## Public API Shape

We add a public attention op:

```python
gdn_chunk_prefill(
    q,
    k,
    v,
    g,
    beta,
    *,
    scale,
    initial_state,
    cu_seqlens,
    qk_l2norm=False,
    output_final_state=True,
    output_h=False,
    override=None,
    solution=None,
)
```

The wrapper derives selection traits from the tensors and options:

```python
traits = {
    "head_dim": q.shape[-1],
    "head_v_dim": v.shape[-1],
    "head_v_eq_head_k": v.shape[-1] == k.shape[-1],
    "num_v_gte_num_q": v.shape[-2] >= q.shape[-2],
    "qk_l2norm": qk_l2norm,
    "output_h": output_h,
}
```

The selected backend is then invoked with the same runtime-facing arguments.

### Triton / FLA

The Triton implementation is registered as:

```text
family="attention"
mode="gdn_chunk_prefill"
name="triton_gdn_chunk_prefill"
solution="triton"
```

[...]

PR #595, opened 2026-07-06
feat(gdn): Initial support for routing chunk prefill through kernel registry

raikonenfnu:
## Summary

Qwen3.5 is one of of the main models being used and deployed today. This family of model depends on hybrid linear attention. (3x linear GDN and 1x gated regular attention). However, the only GDN kernel that is available is Flashinfer's which seem to mainly support Blackwell GPU. Main purpose of this PR/commit is to bring a Triton implementation from FLA to enable GDN on other hardware architectures. In that effort, we generalized the kernel selection for GDN through the kernel registery way which is standard and we have done for other kernels.

## Details

GDN prefill previously kept backend choice in the hybrid runtime backend: runtime knew about the portable FLA/Triton path and the FlashInfer Blackwell fast path directly. That made GDN unlike MHA, where runtime calls a stable tokenspeed-kernel API and the registry owns backend selection by platform, dtype, and traits.

Add a public attention.gdn_chunk_prefill op in tokenspeed-kernel and register Triton/FLA as the portable implementation plus FlashInfer as the specialized Blackwell implementation. The public wrapper derives the selection traits from the q/k/v shapes and options, so future Gluon or CuteDSL implementations can register under the same mode without another runtime call-site split.

Vendor the FLA-derived Triton chunk-prefill helper stack under ops/attention/triton/fla so the portable kernel implementation is self-contained in tokenspeed-kernel. The runtime copies stay in place for runtime-only decode, conv, gating, and state helpers; this commit only moves the implementation needed by the public gdn_chunk_prefill op.

Runtime now calls gdn_chunk_prefill and handles backend-native checkpoint layouts at the boundary: Triton returns FLA h states, while FlashInfer returns compact full-chunk checkpoints. Keeping native layouts avoids converting megabyte-scale state rows solely for API uniformity, while localizing the layout distinction in MambaAttnBackend metadata.

More details in RFC https://github.com/lightseekorg/tokenspeed/issues/589

## Test Plan

```
python -m py_compile tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/triton/fla/*.py ...
passed

pytest -q tokenspeed-kernel/test/ops/test_attention_gdn.py tokenspeed-kernel/test/test_kernel_api_selection.py
29 passed, 27 skipped

pytest -q test/runtime/layers/test_mamba_checkpoint_metadata.py test/runtime/layers/test_gdn_qkv_split_fused.py test/runtime/layers/test_gdn_flashinfer_fastpath.py
22 passed, 25 skipped
```


raikonenfnu:
FYI, this is an effort to break https://github.com/lightseekorg/tokenspeed/pull/592 to smaller PRs. :)

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/__init__.py
```diff
@@ -68,6 +69,113 @@ def _attention_format_signature(**roles: torch.Tensor):
...
+    | tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]
+):
+    """Run Gated Delta Net chunked prefill through kernel selection.
+
+    Args:
+        q: Query tensor shaped ``[1, total_tokens, num_q_heads, head_dim]``.
+        k: Key tensor shaped ``[1, total_tokens, num_k_heads, head_dim]``.
+        v: Value tensor shaped ``[1, total_tokens, num_v_heads, head_v_dim]``.
+        g: Log-space forget gate shaped ``[1, total_tokens, num_v_heads]``.
+        beta: Beta gate shaped ``[1, total_tokens, num_v_heads]``.
```
antiagainst, line 106: Do we need the leading `[1,` here?

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/__init__.py
```diff
@@ -68,6 +69,113 @@ def _attention_format_signature(**roles: torch.Tensor):
...
+        qk_l2norm: Whether the selected kernel should L2-normalize Q/K.
+        output_final_state: Whether to return the final recurrent state.
+        output_h: Whether to return intermediate recurrent checkpoints. Backends
+            may return native checkpoint layouts.
+        override: Optional kernel override name.
+        solution: Optional kernel solution to force through normal selection.
+
+    Returns:
+        ``(out, final_state)``, Triton/FLA-style ``(out, final_state, h)``, or
+        FlashInfer-style ``(out, final_state, h, h_cu_starts)``.
```
antiagainst: They should unify? The purpose of public API is meant for the callsite to avoid special casing towards different solutions. If due to current impl limitations we cannot do that, maybe consider returning a dict or something to be clear what's being returned.

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/triton/linear/op.py
```diff
@@ -0,0 +1,40 @@
...
+# SOFTWARE.
+
+# -*- coding: utf-8 -*-
+
+import os
+
+from tokenspeed_kernel._triton import libdevice as tldevice
+from tokenspeed_kernel._triton import tl, triton
+
+if os.environ.get("FLA_USE_FAST_OPS", "0") == "1":
```
antiagainst: Do we still need this?

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/flashinfer/gated_delta_rule.py
```diff
@@ -36,39 +36,46 @@
...
+platform = current_platform()
 SUPPORTED_HEAD_DIM = 128
 
-_chunk_gated_delta_rule = None
-_AVAILABLE = False
+_chunk_gated_delta_rule = error_fn
+_has_blackwell_prefill = False
 
-if current_platform().is_nvidia:
+if platform.is_nvidia and platform.is_blackwell:
```
antiagainst: `is_blackwell` checks `is_nvidia` internally already.

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/flashinfer/gated_delta_rule.py
```diff
@@ -36,39 +36,46 @@
...
-        )
     except ImportError:
-        _AVAILABLE = False
+        pass
 
 
 def is_available() -> bool:
     """Whether the sm100 GDN kernel can run on this platform."""
-    return _AVAILABLE
+    cuda_major = int(torch.version.cuda.split(".")[0]) if torch.version.cuda else 0
```
antiagainst, line 72: You can use `current_platform().arch_version`

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/flashinfer/__init__.py
```diff
@@ -37,6 +37,11 @@
 
 platform = current_platform()
 
+# Side-effect registration for FlashInfer GDN kernels.
+from tokenspeed_kernel.ops.attention.flashinfer import (  # noqa: E402,F401
```
antiagainst: Do we need this?

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/__init__.py
raikonenfnu: Yeap, this is because fused_qkv_split_gdn_prefill who is producer of gdn_chunk_prefill emits it in that layout per 

usage:
https://github.com/lightseekorg/tokenspeed/blob/3e577bbc178d9d85fdc27d90219eeca9a91f20f5/python/tokenspeed/runtime/layers/attention/backends/hybrid_linear_attn.py#L1062-L1070 

and 

def:
https://github.com/lightseekorg/tokenspeed/blob/3e577bbc178d9d85fdc27d90219eeca9a91f20f5/tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/triton/gdn_qkv_split.py#L174-L194

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/flashinfer/gated_delta_rule.py
raikonenfnu: FYI this was taken from the original code as well. but we might need this since this is taking care of `cuda_version == 13` while `current_platform().arch_version` should be telling you the NV arch/blackwell but in number form (blackwell == 10.0)

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/__init__.py
raikonenfnu: > If due to current impl limitations we cannot do that, maybe consider returning a dict or something to be clear what's being returned.

Added dataclass for resultype and enum for layout for more concrete and easier identification :)

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/triton/linear/op.py
raikonenfnu: Done! that was from original runtime code, who copied from FLA. but it is indeede not used, so removed :D

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/flashinfer/gated_delta_rule.py
raikonenfnu: done!

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/flashinfer/__init__.py
raikonenfnu: If your question was wrt the import, then yes, because iiuc the `register_kernel` mechanism needs the function to be imported. similar to 

https://github.com/lightseekorg/tokenspeed/blob/3e577bbc178d9d85fdc27d90219eeca9a91f20f5/tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/gluon/__init__.py#L21

and 

https://github.com/lightseekorg/tokenspeed/blob/3e577bbc178d9d85fdc27d90219eeca9a91f20f5/tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/flashinfer/__init__.py#L21-L29

If the question was wrt E402, it was since it used to come after platform = current_platform() which is a "executable code" and hence needed E402, but we moved it to a higher position and hence we not need it anymore.

python/tokenspeed/runtime/layers/attention/backends/hybrid_linear_attn.py
```diff
@@ -26,12 +26,17 @@
...
 from tokenspeed_kernel.ops.attention.triton.gdn_qkv_split import (
     fused_qkv_split_gdn_prefill,
 )
+from tokenspeed_kernel.ops.attention.triton.linear.chunk_delta_h import (
+    CHUNK_SIZE as FLA_CHUNK_SIZE,
+)
+from tokenspeed_kernel.ops.attention.triton.linear.index import (
+    set_total_chunks_hint,
+    set_total_chunks_hint_uniform,
+)
```
antiagainst, line 39: These are still leaking solution specific stuff.. But I guess step by step to clean it up..

antiagainst (approved):

python/tokenspeed/runtime/layers/attention/backends/hybrid_linear_attn.py
raikonenfnu: Thanks for the review Lei! Indeed my next step (after smaller Qwen3.5-mxfp4 on AMD piece land) is to unify them. I'll track them in an epic. :)

PR #596, opened 2026-07-06
fix(spec): fix Kimi DP EAGLE3 mixed-step hang

chenht2022:
## Summary

Fixes a Kimi K2.5 DP + EAGLE3 hang when different DP ranks enter different forward modes in the same scheduler step. In the failing case, some ranks were in EXTEND while others were in DECODE. EAGLE3 first-step catch-up narrows activations to one row per request before later draft collectives, but the runtime only applied that reduced sizing on decode ranks. This could make ranks in the same DP step size draft collectives differently and hang.

This PR makes first-step catch-up reduction an explicit draft-model capability and uses it consistently for both active EAGLE forward and idle draft replay. It also guards fused `lm_head_gemm` against zero-token routing / launch.

## Changes

- Add `draft_first_step_reduce_for_catchup` capability to EAGLE/NextN draft models.
- Use the same capability predicate in active EAGLE forward and idle draft replay.
- Guard fused `lm_head_gemm` against zero-token routing / launch.

## Validation

- Before fix: DP8 + EAGLE3 hang.
- After fix: DP8 + EAGLE3 AIME25 completed with mean_acc 0.9333 (28/30).
- Ran `git diff --check` and import/capability smoke.

### DIFF
--- a/python/tokenspeed/runtime/execution/drafter/eagle.py
+++ b/python/tokenspeed/runtime/execution/drafter/eagle.py
@@ -36,13 +36,8 @@ from tokenspeed.runtime.execution.forward_batch_info import (
     CaptureHiddenMode,
     ForwardMode,
 )
-from tokenspeed.runtime.models.llama_eagle3 import LlamaForCausalLMEagle3
-from tokenspeed.runtime.models.qwen3_5_nextn import Qwen3_5ForConditionalGenerationNextN
 from tokenspeed.runtime.multimodal.inputs import maybe_substitute_mm_pad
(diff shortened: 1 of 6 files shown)


rjzhb:
https://github.com/lightseekorg/tokenspeed/pull/544      
hi, thanks for finding this.  This hang is already fixed at the root in PR#544 — it deletes `draft_first_step_reduce` and sizes all draft collectives uniformly via `report_collective_sizing` (no per-model isinstance list, active/idle consistent). You can take a look.  

But the `lm_head_gemm` zero-token guard here is orthogonal and worth keeping.

PR #598, opened 2026-07-07
Add gfx950 Gluon dense16 GEMM

antiagainst:
Added a few variants of fp16/bf16 GEMM kernels:

- `warp`: very small-M (1/2/4) warp-reduction path.
- `splitk`: small-M high-K MFMA/LDS path with a producer kernel and a second partial-sum reduction kernel. Unused due to not yet performant right now.
- `medium`: medium-M  (8 to 128) MFMA/LDS path.
- `largem`: large-M (>= 2048) 8-wave warp pipeline kernel adapted from https://github.com/ROCm/gfx950-gluon-tutorials by @zhanglx13
- `fallback`: fallback to torch.

Bench methodology:

-  `triton.testing.do_bench` for both Gluon and `torch.mm`, with `warmup=25` and `rep=100`. on mi350x
- `speedup = torch_ms / gluon_ms`; values above `1.0x` mean Gluon is faster.

| path | M | N | K | Gluon ms | torch.mm ms | speedup |
|---|---:|---:|---:|---:|---:|---:|
| warp | 1 | 1280 | 1024 | 0.0066 | 0.0093 | 1.41x |
| warp | 1 | 4096 | 2048 | 0.0121 | 0.0126 | 1.04x |
| warp | 2 | 1280 | 1024 | 0.0067 | 0.0093 | 1.39x |
| warp | 2 | 4096 | 2048 | 0.0109 | 0.0127 | 1.16x |
| warp | 4 | 1280 | 512 | 0.0069 | 0.0078 | 1.13x |
| warp | 4 | 4096 | 1024 | 0.0106 | 0.0096 | 0.90x |

| path | M | N | K | Gluon ms | torch.mm ms | speedup |
|---|---:|---:|---:|---:|---:|---:|
| medium | 8 | 1280 | 64 | 0.0066 | 0.0067 | 1.01x |
| medium | 8 | 1280 | 512 | 0.0067 | 0.0078 | 1.17x |
| medium | 8 | 1280 | 1024 | 0.0074 | 0.0081 | 1.10x |
| medium | 16 | 1280 | 768 | 0.0073 | 0.0086 | 1.17x |
| medium | 16 | 1280 | 1024 | 0.0075 | 0.0080 | 1.06x |
| medium | 32 | 2560 | 2048 | 0.0101 | 0.0117 | 1.16x |
| medium | 64 | 1280 | 1024 | 0.0070 | 0.0090 | 1.30x |
| medium | 64 | 1280 | 2048 | 0.0099 | 0.0113 | 1.14x |
| medium | 64 | 1280 | 2880 | 0.0141 | 0.0145 | 1.03x |
| medium | 64 | 2560 | 2048 | 0.0119 | 0.0125 | 1.05x |
| medium | 128 | 1280 | 2048 | 0.0120 | 0.0114 | 0.94x |
| medium | 128 | 1280 | 2880 | 0.0146 | 0.0150 | 1.03x |
| medium | 128 | 2560 | 2048 | 0.0130 | 0.0141 | 1.09x |
| medium | 128 | 4096 | 2048 | 0.0154 | 0.0154 | 1.00x |
| medium | 128 | 4096 | 4096 | 0.0258 | 0.0240 | 0.93x |

| path | M | N | K | Gluon ms | torch.mm ms | speedup |
|---|---:|---:|---:|---:|---:|---:|
| largem | 2048 | 4096 | 4096 | 0.0832 | 0.1311 | 1.58x |
| largem | 2048 | 8192 | 8192 | 0.2451 | 0.2464 | 1.01x |
| largem | 4096 | 4096 | 4096 | 0.1227 | 0.1308 | 1.07x |
| largem | 4096 | 8192 | 8192 | 0.4862 | 0.4838 | 1.00x |


tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/gluon.py
```diff
@@ -0,0 +1,90 @@
...
+
+if current_platform().is_cdna4:
+    try:
+        from tokenspeed_kernel_amd.ops.gemm.mm_a16w16_gfx950 import (
+            gluon_mm_a16w16_gfx950 as _dense16_impl,
+        )
+    except ImportError:
+        _dense16_impl = None
+
+if _dense16_impl is not None:
```
raikonenfnu, line 45: Would it be possible to skip this since we have it in register kernel capabilityRequirements?
```
        capability=CapabilityRequirement(
            min_arch_version=ArchVersion(9, 5),
            max_arch_version=ArchVersion(9, 5),
            vendors=frozenset({"amd"}),
            required_features=frozenset({"tensor_core:f16"}),
        ),
```

tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/gluon.py
```diff
@@ -0,0 +1,90 @@
...
+        B: torch.Tensor,
+        A_scales: torch.Tensor | None,
+        B_scales: torch.Tensor | None,
+        out_dtype: torch.dtype,
+        *,
+        alpha: torch.Tensor | None = None,
+        block_size: list[int] | None = None,
+    ):
+        if A_scales is not None or B_scales is not None or block_size is not None:
+            return None
```
raikonenfnu: Do we expect to return None here btw, should we raise error?

tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/gluon.py
```diff
@@ -0,0 +1,90 @@
...
+
+        output = _dense16_impl(A, B, out_dtype, alpha=alpha)
+        if output is not None:
+            return output
+
+        # TODO: Optimize M >= 256 and M <= 1024 dense16 cases in Gluon.
+        output = F.linear(A, B)
```
raikonenfnu, line 89: Do we expect to fallback to `F.linear` within this function? Any thoughts on moving some heuristic pieces into `traits` so we can let the register_kernel do the fallback?

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gemm/mm_a16w16_gfx950.py
```diff
@@ -0,0 +1,1473 @@
...
+WARP_REDUCE_BLOCK_K = 1024
+MFMA_LDS_BLOCK_N = 256
+MFMA_LDS_BLOCK_K = 64
+MFMA_LDS_NUM_WARPS = 4
+MFMA_LDS_REDUCE_M = 4
+MFMA_LDS_SPLIT_K = 8
+GROUP_SIZE_M = 2
+MAX_M = 256
+LARGEM_DISPATCH_MIN_M = 2048
+KERNEL_NAME = "gluon_mm_a16w16_gfx950"
```
qedawkins: Not used anywhere?

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gemm/mm_a16w16_gfx950.py
```diff
@@ -0,0 +1,1473 @@
...
+            mask=valid_out[:, None] & k_valid[None, :],
+            other=0.0,
+        ).to(gl.float32)
+        b = cdna4.buffer_load(
+            ptr=b_ptr,
+            offsets=b_offsets,
+            mask=valid_out[:, None] & k_valid[None, :],
+            other=0.0,
+        ).to(gl.float32)
+        acc += gl.sum(a * b, axis=1)
```
qedawkins, line 399: Why reduce every iteration instead of once at the end?

tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/gluon.py
antiagainst: This is meant to be temporary so that later once we have optimized the fallback paths too it will go away. I'll work on that next. We are not trying to beat hipblaslt all around; more focusing on shapes really used in Q/K/V projection etc. 

With that context, using `trait` would be too much churn for such purpose given we can create new kernel subvariants or tweak what shape maps to what subvariants. It would also make `traits` more complicated given all the tricks trying to "emulate" various logic combinations. So I didn't take that apporach yet. Currently treating `traits` more high level categorical characteristics rather than the only solution for switches. But we can evolve in the future once we have more kernels in stable state.

tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/gluon.py
antiagainst: For the `try: import` it's not given that we need to guard against case that amd package not installed on nvidia. For the `if current_platform` yeah we can drop it given it's sort of overlapping with what capability declaration is. Benefit of keeping it might be reducing on the amount of kernel candidates in the cache as this is clearly for gfx950 only. So both are okay actually.

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gemm/mm_a16w16_gfx950.py
antiagainst: Good catch. Agents followed style from somewhere else. Dropped. :)

tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/gluon.py
antiagainst: Yeah make sense. Done.

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gemm/mm_a16w16_gfx950.py
antiagainst: For such extreme small shapes the bottleneck is really global memory access so the amount of compute doesn't matter that much. Warp reduction helps to distribute the problem differently with more workgroup launched to fill the GPU. Switching to do reduction outside of the loop means we are carrying more live fp32 registers which would hurt occupancy so can hurt memory access requests, comparing to the compute savings it's doing. Currently `gl.sum` or other `tt.reduce` is already lowering to dpp cross-lane wave reduce so not a lot there tbh.

raikonenfnu (approved):
Overall LGTM!

PR #602, opened 2026-07-07
fix(deepseek-v3): Route DeepSeek merge state through unified backend

Max191:
## Summary

DeepseekV3's prefix-cache replay path used the CUDA-specific merge_state attention helper. On AMD, this caused TokenSpeed to try loading thirdparty/cuda/objs/merge_state/merge_state.so when caching prefix tokens.

This PR fixes the issue by routing through the unified `tokenspeed_kernel.ops.attention.attn_merge_state` backend, which has a triton implementation. On NVIDIA, this routes to the same merge_state helper that was used before.

## Test Plan

Ran `tokenspeed bench serve` with prefix caching enabled, which fails without the fix, and succeeds with the fix.

### DIFF
--- a/tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/cuda/__init__.py
+++ b/tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/cuda/__init__.py
@@ -36,11 +36,15 @@ if platform.is_nvidia and platform.is_hopper_plus:
         out_b: torch.Tensor,
         lse_b: torch.Tensor,
         lse_scale_log2: float,
+        inplace: bool = False,
+        enable_pdl: bool = False,
     ) -> tuple[torch.Tensor, torch.Tensor]:
         return merge_state(
             out_a,
             lse_a,
             out_b,
             lse_b,
+            inplace=inplace,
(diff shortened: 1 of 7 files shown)


tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/cuda/__init__.py
```diff
@@ -20,7 +20,7 @@
         name="cuda_attn_merge_state",
         solution="cuda",
         capability=CapabilityRequirement(
-            min_arch_version=ArchVersion(10, 0),
+            min_arch_version=ArchVersion(9, 0),
```
antiagainst, line 23: Is this intentional?

tokenspeed-kernel/test/test_kernel_api_selection.py
```diff
@@ -981,7 +981,7 @@ def _case(
         "hopper",
         "attention",
         "attn_merge_state",
-        "triton_attn_merge_state",
+        "cuda_attn_merge_state",
```
antiagainst, line 984: Do we need a corresponding test case for cdna4 to check it's routing to trition?

PR #608, opened 2026-07-08
perf(multimodal): optimize Qwen3.5 M-RoPE preparation

yechank-nvidia:
## Summary

  - Build Qwen3.5 M-RoPE positions directly from validated multimodal placeholder offsets instead of rescanning token IDs.
  - Reuse repeated video grid positions within each request without introducing a process-wide cache.
  - Reduce M-RoPE prefill and decode host overhead by reusing scalar deltas, avoiding unnecessary H2D copies, and bypassing single-chunk concatenation.
  - Preserve the existing generic implementation as a fallback for missing or inconsistent offset metadata.

  The offset path improves M-RoPE preparation by 3.08x for images and 7.20x-19.13x for tested video sequence lengths.

  ## Test Plan

  - Verified image, split-video, timestamped-video, mixed-item, uniform-delta, mixed-delta, and invalid-offset fallback cases.
  - Compared optimized output against the existing generic implementation, including 100 randomized mixed image/video layouts.
  - Replayed an actual 9,890-token SMG timestamp-video request:
    - HTTP 200
    - Confirmed the validated offset path was used
    - Exact output and M-RoPE delta matched the expected layout
  - Ran long-output Qwen3.5-4B E2E at concurrency 1/8/32 with OSL 512/1024. Results remained within normal run variance, as generation dominates these
  workloads.
  - `pytest -q test/runtime/test_multimodal_mrope.py`


chenht2022:
Reproduced on Qwen3.5-4B with a single B200: +3.15% E2E throughput for the 8×1080p image workload at concurrency 32. M-RoPE latency dropped from 4.50 ms to 0.81 ms.
LGTM.

chenht2022 (approved):

PR #610, opened 2026-07-08
perf(multimodal): optimize SHM input transport

yechank-nvidia:
## Summary

  - Report the loaded vision encoder dtype so SMG can serialize multimodal tensors in the destination dtype.
  - Interleave large SHM-to-pinned copies with asynchronous H2D transfers.
  - Overlap SHM unlink cleanup with H2D for single large tensors.
  - For large tensor-parallel inputs, copy SHM on one vision TP rank and broadcast the GPU tensor instead of repeating the host copy on every rank.
  - Copy single TP source tensors directly from SHM into the destination GPU buffer.
  - Preserve the existing staged path for small, mixed-dtype, inline, and non-CUDA inputs.

  ## Test Plan

  - Verified vision encoder dtype discovery for direct and nested vision towers.
  - Verified SHM copy, pinned-memory staging, direct destination copy, cleanup, and dtype validation.
  - Verified TP broadcast selection thresholds and fallback behavior for mixed dtype and inline tensors.
  - Verified CUDA value preservation for both small staged tensors and large interleaved tensors.
  - Benchmarked SHM/H2D paths on B200 with single-GPU and TP configurations.
  - `pytest -q test/runtime/test_multimodal_encoder_dtype.py test/runtime/test_multimodal_shm_transport.py test/runtime/test_multimodal_tp_broadcast.py`


chenht2022:
Reproduced on NVIDIA B200 GPUs. A direct SHM-to-GPU transport benchmark with 64–128 MiB payloads showed 11–37% lower latency at TP1/TP2/TP4. On Qwen3.5-4B TP2 with 8×1080p images at concurrency 32, E2E throughput improved by 1.74%

chenht2022:
Pushed a small follow-up to report multimodal_encoder_dtype from disaggregated encode workers.

chenht2022 (dismissed):

chenht2022 (dismissed):

chenht2022 (approved):

PR #611, opened 2026-07-08
feat: enable the prefill CUDA graph by default

nperrin-fr:
Default prefill_graph_max_tokens: None = auto = min(2048, chunked-prefill size, max_total_tokens) -- covers the launch-bound small-prefill range while keeping capture time and graph-pool memory modest (matches sglang's MLA default of 2048). 0 still disables; explicit > 0 caps as before.

Capture OOM fails the boot loudly with an actionable message (free --gpu-memory-utilization headroom, lower --prefill-graph-max-tokens, or 0 to disable) -- the operator decides, nothing degrades silently.


nperrin-fr:
The AMD failure was the new default reaching an unvalidated platform: with the graph on by default, the mi355 boot never became ready (most likely the per-bucket capture warmups triggering a Triton/Gluon JIT storm on ROCm, possibly a hipGraph multi-segment capture issue — no AMD box here to confirm). Fixed in d5742206b: the default now enables the prefill graph on NVIDIA only (where it is validated); other platforms stay eager unless `--prefill-graph-max-tokens` is set explicitly. AMD enablement should be its own validated change.

nperrin-fr:
Correction + rollback of the NVIDIA-only gate: the mi355 failures are node flakiness, not the default — with the gate in place (graph off on AMD) a *different* mi355 job failed the same way, and the same jobs flip red/green across unrelated PRs all day (qedawkins, yechan). Better: the first graph-on AMD round actually PASSED 7/8 mi355 jobs, and their logs show the graph genuinely capturing and replaying on ROCm (`captured buckets [16, 32, ...]` + `ENGAGED (text)`) with evals green — so AMD works and is now CI-validated. The default is platform-uniform again.

PR #620, opened 2026-07-09
feat: add DeepSeek V4 L2 KV cache offload and perf optimize.

SimonCqk:
## Summary

This PR adds DeepSeek V4 L2 KV cache offload support for host-side CPU KV cache reuse.

Main changes:
- Add scheduler/runtime support for DeepSeek V4 paged L2 KV cache writeback/loadback.
- Add history-family + terminal-state prefix reuse handling for host L2 prefix hits.
- Optimize fragmented paged L2 transfers with coalescing and H2D scatter transfer support.
- Optimize DeepSeek V4 HCA cached-prefill cache insert by compacting active rows.
- Add focused tests for scheduler cache ops, Python binding shape, transfer behavior, and prefix reuse correctness.

## Performance Setup

For the benchmark comparison we use `dp=1, tp=4`.

Reason: each DP rank owns its own device/host KV cache pool, so `dp=4` shards requests across independent cache pools and can dilute prefix reuse. `dp=1, tp=4` keeps reuse in one KV pool while still using tensor parallelism.

To construct a **capacity-bound** scenario, the server is started with:

```bash
--max-total-tokens 81920
```

Baseline without kvstore:

```bash
tokenspeed serve --model deepseek-ai/DeepSeek-V4-Flash --port 30101 --trust-remote-code --data-parallel-size 1 --tensor-parallel-size 4 --enable-expert-parallel --kv-cache-dtype fp8_e4m3 --chunked-prefill-size 8192 --attention-use-fp4-indexer-cache --moe-backend mega_moe --dist-init-addr 127.0.0.1:14013 --prometheus-port 12211 --enable-mixed-batch --gpu-memory-utilization 0.85 --max-total-tokens 81920 --disable-kvstore
```

L2 kvstore enabled:

```bash
tokenspeed serve --model deepseek-ai/DeepSeek-V4-Flash --port 30101 --trust-remote-code --data-parallel-size 1 --tensor-parallel-size 4 --enable-expert-parallel --kv-cache-dtype fp8_e4m3 --chunked-prefill-size 8192 --attention-use-fp4-indexer-cache --moe-backend mega_moe --dist-init-addr 127.0.0.1:14013 --prometheus-port 12211 --enable-mixed-batch --gpu-memory-utilization 0.85 --max-total-tokens 81920 --kvstore-ratio 20
```

Overall performance comparison:

| Metric | Disable KVStore | Enable KVStore | Delta |
| --- | ---: | ---: | ---: |
| Success | 42 / 43 | 48 / 48 | disable has 1 retracted/abort request |
| Output throughput | 53.23 tok/s | 69.57 tok/s | +30.70% |
| Total throughput | 6925.72 tok/s | 9174.15 tok/s | +32.46% |
| Avg latency | 17.01s | 13.94s | -18.05% |
| TTFT avg | 10316.53ms | 7333.52ms | -28.91% |
| TTFT p50 | 12537.04ms | 7054.09ms | -43.73% |
| TTFT p90 | 13374.15ms | 9983.89ms | -25.35% |
| TTFT p99 | 13769.64ms | 12218.83ms | -11.26% |
| Subsequent-turn TTFT | 10286.49ms | 7133.57ms | -30.65% |
| TPOT avg | 13.42ms | 13.24ms | -1.34% |
| Steady cached prompt throughput | 6597.57 tok/s | 8846.47 tok/s | +34.09% |
| Steady completion throughput | 53.34 tok/s | 70.02 tok/s | +31.27% |
| Trace cache hit mean | 87.51% | 91.10% | +3.59pp |

## Test Regression

- GSM8K limit 200, accuracy: 0.983;


tokenspeed-scheduler/csrc/resource/hybrid_prefix_cache/hybrid_prefix_cache.cpp
```diff
@@ -758,6 +816,19 @@ TreeNode* CapNodeToDepth(TreeNode* from, std::int32_t depth) {
     return node;
 }
 
+enum class PagedCacheResidency { kDevice, kHost };
+
+const PagedCacheSnapshot* GetPagedCacheSnapshotForResidency(TreeNode* node, PagedCacheResidency residency) {
+    if (node == nullptr) return nullptr;
+    switch (residency) {
+        case PagedCacheResidency::kDevice:
+            return node->GetPagedCacheSnapshot();
```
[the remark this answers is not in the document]
SimonCqk: done

python/tokenspeed/runtime/engine/event_loop.py
```diff
@@ -309,7 +322,11 @@ def __init__(
             mamba_l2_layout=server_args.mamba_l2_layout,
             mamba_l2_io_backend=server_args.mamba_l2_io_backend,
         )
-        if not token_to_kv_pool.supports_hierarchical_kv_cache:
+        paged_cache_host_group_pages = {}
+        if not create_memory_executor:
+            self.memory_executor = None
+            num_host_pages = 0
```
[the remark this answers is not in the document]
SimonCqk: done

tokenspeed-scheduler/csrc/resource/radix_tree/radix_tree.cpp
```diff
@@ -174,7 +178,8 @@ WalkResult RadixTree::WalkDownUtilMismatch(token_slice aligned_tokens, TreeNode:
         }
         if (matched_num_pages != static_cast<std::int32_t>(child->Tokens().size() / page_size_)) {
             // Refuse to split a snapshot-bearing node; borrowed ids rely on it.
-            if (child->HasPagedCacheSnapshot()) {
+            if (child->HasPagedCacheSnapshot() || child->HasPagedCacheHostSnapshot() ||
+                child->HasPagedCachePendingHostSnapshot()) {
```
dongjiyingdjy: Snapshot splits can stay unsupported, but a partial match here leaves the walk at the parent with all tokens unconsumed. KVPrefixCache::Insert then adds the same first-page key, and AddChild replaces the existing snapshot branch, leaving adjunct raw pointers dangling. Please return an explicit no-insert/blocked result for this path and reject duplicate child keys.

tokenspeed-scheduler/csrc/resource/hybrid_prefix_cache/hybrid_prefix_cache.cpp
```diff
@@ -863,151 +957,204 @@ bool AssemblePagedCacheStateGroupPagesToTarget(MatchResult::PagedCache& out, con
...
-        cap_to_root();
-        return;
-    }
-
-    std::int32_t worst_window = 0;
-    for (const auto& gid : paged_cache_state_groups_) {
-        auto it = paged_cache_sliding_window_per_group_.find(gid);
-        if (it != paged_cache_sliding_window_per_group_.end()) {
-            worst_window = std::max(worst_window, it->second);
+        // Missing continuation-only State groups cap PrefixReuse to the replayable
```
dongjiyingdjy: Compressed History has no valid rebuild point when continuation State is missing: SWA cannot be reconstructed from it, so returning prefix_len_tokens = history_depth makes the suffix read freshly allocated, uninitialized State pages. Since partial replay is unsupported, please restore the terminal-State guard and treat this as a cache miss/full prefill.

tokenspeed-scheduler/csrc/scheduler/outside_event_handler.cpp
```diff
@@ -157,10 +157,20 @@ void Scheduler::handleEvent(const cache::WriteBackDone& event) {
 
     if (!spec.request_id.empty()) {
         if (auto* req = find_request(spec.request_id)) {
-            req->Apply(
-                fsm::WriteBackDoneEvent{&kv_prefix_cache_, hybrid_prefix_cache_ ? &*hybrid_prefix_cache_ : nullptr});
+            // A finishing request no longer needs its paged-cache table once
+            // D2H has completed. Release borrowed pages before device demotion;
+            // retracting requests keep the table for recovery.
+            if (hybrid_prefix_cache_ && req->Is<fsm::WritingBack>()) {
```
dongjiyingdjy: Retracting requests keep their V4 paged table here. After D2H, `OnKVDeviceDemote` therefore preserves borrowed snapshots, while unpublished/overlap tail pages were never transferred, so retract does not release the victim’s paged-group capacity. If admission failed on group exhaustion, requests can be retracted until scheduling stalls. Please fully offload then `ReleaseRequest`; with `--disable-kvstore`, abort/release (or recompute) the victim instead of acknowledging a no-op retract.

python/tokenspeed/runtime/cache/executor/memory_executor.py
```diff
@@ -124,107 +162,140 @@ def __init__(
...
-            actual_draft_pool = draft_device_pool
-            if hasattr(draft_device_pool, "inner") and not isinstance(
-                draft_device_pool, kv_pool_types
-            ):
-                actual_draft_pool = draft_device_pool.inner
-            if not isinstance(actual_draft_pool, kv_pool_types):
-                raise ValueError(
-                    f"draft_device_pool only supports DSA, MHA and MLA, "
-                    f"got {type(actual_draft_pool)}"
+        if isinstance(actual_pool, DeepseekV4TokenToKVPool):
```
dongjiyingdjy, line 179: When the target is V4, this branch never handles `draft_device_pool`. V4 MTP keeps target and draft KV in independent tensors but feeds both the same scheduler page IDs, so after a host-prefix loadback only target state is restored and the draft reads stale/reused KV. Target verification protects committed tokens, but MTP acceptance can collapse. Please mirror/copy draft tensors in each paged transfer (and budget them), or reject V4 KVStore + MTP explicitly.

tokenspeed-scheduler/csrc/resource/radix_tree/radix_tree.cpp
SimonCqk: done

tokenspeed-scheduler/csrc/scheduler/outside_event_handler.cpp
SimonCqk: Decode/recovery admission now distinguishes paged-group exhaustion from generic resource pressure. When no forward/cache work remains in flight and all attempted decodes are blocked by paged capacity, the scheduler deterministically aborts one victim and ReleaseRequest frees its request-local paged tables. Mixed generic pressure keeps the existing retract behavior. Focused scheduler and runtime-abort tests cover this path.

python/tokenspeed/runtime/cache/executor/memory_executor.py
SimonCqk: Startup now rejects KVStore when grouped paged caches and a speculative draft KV pool coexist, before host-pool allocation. --disable-kvstore, configurations without a draft pool, and non-paged KVStore draft paths remain allowed. The behavior is covered by the parametrized grouped-KVStore draft-pool test.

And MTP with enabled L2 kvstore will be in next coming follow up seperate PR.

tokenspeed-scheduler/csrc/resource/hybrid_prefix_cache/hybrid_prefix_cache.cpp
```diff
@@ -1169,6 +1323,179 @@ void HybridPrefixCache::RewindRequest(const std::string& request_id, std::int32_
...
+
+        PagedCacheTransferPair transfer{gid, host_pages, device_pages.Ids()};
+        auto [table_it, inserted] = tables.emplace(gid, PagedCacheGroupTable(device_alloc_it->second.get()));
+        if (!inserted) {
+            ReleaseRequest(request_id);
+            throw std::logic_error("HybridPrefixCache::PreparePagedCacheDeviceLoadBack: duplicate table for group " +
+                                   gid);
+        }
+        table_it->second.ImportPrefixOwned(std::move(device_pages), base_logical_page, host_hit.prefix_len_tokens);
+        transfers.push_back(std::move(transfer));
```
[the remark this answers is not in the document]
SimonCqk: Each paged H2D loadback now stores a HostNodeRef in the scheduler’s cache-op tracker. The paged HostExecutor emits LoadBackDoneEvent only after the final layer H2D event completes, and the scheduler then releases the pin. Ordinary radix KV/Mamba loadbacks remain unchanged, and no CUDA synchronize or stream wait was added. The scheduler lifecycle and HostExecutor completion paths have focused regression coverage.

tokenspeed-scheduler/csrc/scheduler/outside_event_handler.cpp
```diff
@@ -177,12 +175,12 @@ void Scheduler::handleEvent(const forward::ExtendResult& event) {
 }
 
 void Scheduler::handleEvent(const forward::Abort& event) {
+    // Terminal for this request's forward stream: late results are ignored.
+    pending_forward_results_.erase(event.request_id);
```
dongjiyingdjy, line 225: **[P1] Defer client abort until request-scoped cache operations acknowledge**

Unlike `Scheduler::abortRequest`, this client-abort path does not check whether the request still owns an entry in `cache_op_tracker_`. If an abort races a request-scoped cache operation, a request in `WritingBack` hits no valid `AbortEvent` transition and throws; a request in `Retracting` or with a paged H2D loadback can become `Finished` and be erased on the next plan, dropping its page pins and request table while DMA is still reading or writing those pages. The later ACK then has no live request through which to complete or roll back the operation.

Please keep the aborted request unschedulable but alive until its last tracked cache operation acknowledges, then terminalize and release it. The existing `Prefetching -> Aborting -> PrefetchDone` ownership handoff should remain intact.

tokenspeed-scheduler/csrc/fsm/forward_events.cpp
```diff
@@ -578,14 +653,24 @@ std::variant<Draining, Finished> FinishEvent::apply(ForwardStateT&& state) {
...
+        std::vector<PagedCacheTransferPair> paged_cache_writeback_transfers;
+        std::vector<TreeNode*> paged_cache_writeback_nodes;
         if (hybrid_prefix_cache_ != nullptr) {
             auto mamba_pairs = hybrid_prefix_cache_->PrepareMambaHostWriteBack(write_diff);
             mamba_writeback_nodes = MambaNodesForTransferPairs(write_diff, mamba_pairs);
             pages_to_transfer.insert(pages_to_transfer.end(), std::make_move_iterator(mamba_pairs.begin()),
                                      std::make_move_iterator(mamba_pairs.end()));
+            paged_cache_writeback_nodes = PagedCacheSnapshotNodesForWriteBack(match.device.last_node);
+            paged_cache_writeback_transfers =
+                hybrid_prefix_cache_->PreparePagedCacheHostWriteBack(paged_cache_writeback_nodes);
```
dongjiyingdjy, line 612: **[P1] Roll back prepared host state when aborting `Draining`**

`FinishEvent` has already attached host KV resources and prepared the Mamba/paged host state before returning `Draining`, but no cache operation or future ACK exists until the next plan submits the D2H work. If a client abort arrives in that window, `AbortEvent::operator()(Draining&&)` simply returns `Finished`; destroying the state drops its node refs but does not remove the unpublished host KV resources or resolve the pending Mamba/paged snapshots. A later prefix hit can therefore treat host pages that were never written as valid, while the pending host pages and flags leak and can block future writebacks.

Please perform the same failure rollback used by an unsuccessful writeback before terminalizing a `Draining` request: resolve the Mamba/paged preparations with `success=false` and release the unwritten host KV resources after dropping the host pin.

tokenspeed-scheduler/csrc/fsm/forward_events.cpp
```diff
@@ -634,11 +729,18 @@ Retracted WriteBackDoneEvent::operator()(Retracting&& state) {
     std::int32_t page_size = state.GetPageSize();
     TreeNode* device_node = state.DeviceNode();
     if (hybrid_prefix_cache_ != nullptr) {
-        hybrid_prefix_cache_->OnMambaHostWriteBackDone(state.MambaWriteBackNodes());
+        hybrid_prefix_cache_->OnMambaHostWriteBackDone(state.MambaWriteBackNodes(), success_);
+        hybrid_prefix_cache_->OnPagedCacheHostWriteBackDone(state.PagedCacheWriteBackNodes(), success_);
+    }
+    if (!success_) {
```
dongjiyingdjy: **[P1] Publish a scheduler abort when retract writeback fails**

This failure branch removes the invalid host destinations, but the transition still returns `Retracted` unconditionally below. That state claims the request has recoverable offloaded state even though the D2H copy failed; after the device pin is dropped, the original prefix can be evicted under the same memory pressure, and `scheduleDecodeFromRetracted` may then resume without the complete KV history.

There is currently no retry state, error classification, or retry budget, so `success=false` should be terminal: after all request-scoped cache operations have acknowledged, publish exactly one `SchedulerAbort` (for example, `Retract writeback failed`), release the request safely, and do not expose it as `Retracted`. A future retry path would instead need to retain the original device state and pins until a bounded retry succeeds.

SimonCqk: done

tokenspeed-scheduler/csrc/scheduler/outside_event_handler.cpp
SimonCqk: `ForwardEvent::Abort` now keeps request-scoped writeback/loadback operations alive in a unified deferred-abort ledger and excludes the request from further scheduling.

Finalization waits until tracked cache work, and scheduler-owned forward debt where applicable, has retired. A writeback ACK received after the no-cache abort is treated as a logical failure even if the physical D2H succeeded, so KV, Mamba, and paged host state is rolled back instead of published. Paged H2D source pins remain held until LoadBackDone.

tokenspeed-scheduler/csrc/fsm/forward_events.cpp
SimonCqk: Draining aborts now use the same rollback recipe as failed writebacks: pending Mamba and paged snapshots are resolved with `success=false`, the host pin is dropped, unwritten host KV resources are released, and the device pin is released before terminalization.

tokenspeed-kernel/python/tokenspeed_kernel/thirdparty/cuda/kvcacheio.py
```diff
@@ -64,19 +120,268 @@ def _check_direct_copy_args(
...
 
+def transfer_kv_direct_h2d_scatter_prepared(
+    plan: DirectH2DScatterPlan,
+    src_indices: torch.Tensor,
+    dst_indices: torch.Tensor,
+    page_size: int,
+    entry_begin: int,
+    entry_end: int,
+) -> _DirectH2DScatterResult:
+    """Launch a layer-range H2D scatter from a prepared full-group plan."""
```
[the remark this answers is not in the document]
SimonCqk: done

tokenspeed-kernel/python/tokenspeed_kernel/ops/kvcache/cuda.py
```diff
@@ -20,37 +20,48 @@
...
 """CUDA KV cache transfer kernels."""
 
+from typing import Any
+
 from tokenspeed_kernel.registry import error_fn
 
 try:
     from tokenspeed_kernel.thirdparty.cuda.kvcacheio import (
+        DirectH2DScatterPlan,
+        prepare_kv_direct_h2d_scatter_plan,
```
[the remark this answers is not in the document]
SimonCqk: done

dongjiyingdjy (approved):
Approving — all seven of my earlier P1 threads are substantively fixed (verified against the latest head: the split-safety + duplicate-child-key rejection, restored continuation-state guards with dedicated tests, exhaustion victim abort, grouped-KVStore/draft-pool startup rejection, deferred-abort ledger, Draining rollback, and retract-writeback failure terminalization). The lifecycle design held up under a deep pass: no bare free-during-DMA races found.

Non-blocking follow-ups for a subsequent PR, roughly in priority order:

1. `forward.cpp` (`schedulePrefillFirstChunk` / `scheduleDecodeFromRetracted`): the mamba-slot early returns after a successful `PreparePagedCacheDeviceLoadBack` leak the request tables — next round either skips the transfer and feeds never-filled device pages to forward (retracted path), or throws `logic_error` in Prepare (prefill path). `ReleaseRequest()` at those exits. Only reachable with a mamba adjunct + paged L2 combined, but the failure mode is silent corruption.
2. `outside_event_handler.cpp` `handleEvent(pd::FailedEvent)`: applies `AbortEvent` directly, bypassing the deferred-abort ledger (no `hasInFlightCacheOp` gate; `Retracting` has no transition and throws; the event carries no prefix-cache pointers so a Draining rollback is a no-op). Route it through `deferAbort(discard_writeback=true)` + `tryFinalizeDeferredAbort`.
3. `memory_executor.py` loadback: an op whose paged transfers are empty is silently skipped, but event_loop counts its op_ids in-flight — no ack ever arrives (the writeback side has an immediate-completion fallback; loadback needs the symmetric one). The existing test constructs this exact case but only asserts the enqueue.
4. `ImportPrefixOwned` sets State-group `committed = hit depth`, so a **required** State group makes CommitChunk's preflight report `target_not_advanced` forever after a host loadback — no device snapshot is ever republished. Worth a dedicated required-State + host-hit test to settle reachability.
5. Smaller items: loadback failure is a hard `_assert` (terminalize the request instead); page indices are never bounds-checked (D2H overrun silently corrupts registered host heap — a cheap optional assert would help); the HCA compress kernel silently writes NaN when all window sources are non-resident (debug assert on `valid_source`); an end-to-end writeback→loadback data roundtrip test is missing; `io_backend="kernel"` is accepted but silently runs the direct path.

python/tokenspeed/runtime/cache/executor/host_executor.py
```diff
@@ -339,7 +453,65 @@ def _start_loading(self) -> None:
...
+                for layer_start in range(0, num_layers, layer_chunk_size):
+                    layer_end = min(layer_start + layer_chunk_size, num_layers)
+                    self.paged_pool.loadback_prepared_paged_range(
+                        prepared_transfers,
+                        layer_start,
+                        layer_end,
+                    )
+                    for layer_index in range(layer_start, layer_end):
+                        producer_event.complete(layer_index)
+                op_ids = _ordered_unique(unit.op_id for unit in self.paged_load_queue)
```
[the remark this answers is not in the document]
SimonCqk: This expression already derives the IDs directly from self.paged_load_queue. The unit name is bound by the generator expression itself and does not depend on the preceding regular-loadback loop. The paged-only executor case is covered by test_host_executor_paged_cache_loadback_reuses_prepared_transfers.

PR #655, opened 2026-07-12
feat(profile): drive Proton via the /start_profile control plane

antiagainst:
Proton could not profile serving workloads: GPU kernels run in scheduler subprocesses (mp.Process spawned in _launch_subprocesses), and those are torn down with SIGKILL on both crash and clean engine shutdown. The import-time TOKENSPEED_KERNEL_PROFILE bootstrap finalizes via atexit, which never runs under SIGKILL, so no profile was ever written; all ranks also shared one default output path.

Fix by running Proton where the kernels launch, controlled through the existing profile control plane:

* request_handler: new PROTON activity for /start_profile and /stop_profile. Each scheduler rank starts its own Proton session writing <output_dir>/<profile_id>-TP-<rank>[-STAGE].proton.<fmt> and finalizes it on /stop_profile while the process is alive, so nothing depends on atexit or graceful exit. The existing start_step / num_steps / profile_by_stage step windows apply to Proton unchanged.

* init_profile validation: reject PROTON combined with GPU or CUDA_PROFILER (CUPTI/roctracer allows one GPU profiling client per process; host-side CPU/MEM/VIZTRACER still compose), and fail cleanly when Proton is unavailable or an env-bootstrap session already owns the process.

* tokenspeed_kernel.profiling: expose proton_available() and profile_config_from_env(output=...) so the control plane reuses the TOKENSPEED_KERNEL_PROFILE_* env config (data/backend/mode/ hook) with a per-rank output override. Env bootstrap behavior is unchanged and remains the right tool for single-process scripts.

* bench serve: new --profile-activities flag (e.g. PROTON) sent in the /start_profile body. The profile request now uses a dedicated body dict, so profiler fields like num_steps no longer leak into every generation request payload.

* Document the serving flow in the tokenspeed-kernel README; add unit tests for the new helpers and the scheduler-side PROTON dispatch, activity-conflict rejection, and step-window finalize.

### DIFF
--- a/python/tokenspeed/runtime/engine/request_handler.py
+++ b/python/tokenspeed/runtime/engine/request_handler.py
@@ -324,13 +324,14 @@ class RequestHandler:
                 message="Profiling is already in progress. Call /stop_profile first.",
             )
 
-        self.profile_by_stage = profile_by_stage
-
         if output_dir is None:
             output_dir = envs.TOKENSPEED_PROFILER_DIR.get()
         if activities is None:
             activities = ["CPU", "GPU"]
 
+        # All validation must precede any state mutation: the event loop runs
+        # _profile_batch_predicate on every batch, so a rejected request that
+        # left partial profiler state behind would crash the scheduler.
         if "PROTON" in activities:
             conflicting = sorted({"GPU", "CUDA_PROFILER"} & set(activities))
             if conflicting:
@@ -354,6 +355,7 @@ class RequestHandler:
                     "be controlled through /start_profile.",
                 )
 
+        self.profile_by_stage = profile_by_stage
         self.profiler_output_dir = output_dir
         self.torch_profiler_with_stack = with_stack
         self.torch_profiler_record_shapes = record_shapes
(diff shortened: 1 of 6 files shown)


python/tokenspeed/runtime/engine/request_handler.py
```diff
@@ -266,8 +285,11 @@ def handle_generate_request(
         )
 
     # ------------------------------------------------------------------
-    # Profiling: torch / cuda / viztracer / mem-snapshot, driven by
-    # /start_profile and /stop_profile control requests.
+    # Profiling: torch / cuda / viztracer / mem-snapshot / proton, driven
+    # by /start_profile and /stop_profile control requests. Proton must be
+    # driven from this process (not the frontend): its GPU hooks are
+    # per-process and the scheduler subprocess is torn down with SIGKILL,
+    # so an atexit-based finalize would never write the profile.
```
qedawkins, line 292: Since the second half of this comment is specific to Proton it might be better to move it closer to the Proton specific stop_profiling call.

qedawkins (approved):
Makes sense to me

borontion (approved):

PR #665, opened 2026-07-13
fix(prefill-graph): illegal memory access on the flat KV-cache path

nperrin-fr:
## Bug

The prefill CUDA graph (default-on since #611) pads q/k/v rows to the capture bucket, but flat per-group KV write locations cover only the real tokens. On any flat-built serve the KV store kernel walks past the loc array on the first padded replay — an illegal memory access that kills the engine (gpt-oss + mha reproduces on the first non-bucket-sized prompt; warmup can survive by landing exactly on a bucket). The trtllm backend has a second, silent variant: its extend prewrite (fused RoPE KV write) runs inside the graphed segment, so capture bakes that forward's write-location tensor into the graph — every replay rewrites capture-time locations (the dummy page on flat, a stale input-buffer tail on radix). No crash, just corrupted KV.

## Fix (backend-agnostic)

- `FlatCacheGroupsMixin._select_out_cache_loc` extends the group's write locs to the caller's (bucket-padded) row count with dummy slot 0 — the radix tail convention (padded rows are scrubbed and land in the null page). Memoized per group per forward; a no-op off the padded path. One fix point: every flat-capable backend inherits it instead of patching each KV write site.
- Prefill-graph capture feeds all-zero flat per-group tables (the reserved null block 0, the decode-capture convention) so capture exercises the same flat metadata branch as replay. State groups are skipped.
- trtllm `support_kv_cache_prewrite` returns False while a breakable capture is active, so the captured branch hands KV to the eager attention break, which resolves fresh per-step locations from metadata.

## Validation (gpt-oss-120b, B200, flat scheduler ext, PFG-on vs PFG-off)

| Gate | mha | trtllm |
|---|---|---|
| GSM8K | 0.9625 / 0.9621 (400q / 1319q eager) | 0.9650 / 0.9625 (400q) |
| GPQA-Diamond ×3 | 0.6953 / 0.7020 | — |
| AIME25 ×3 (64k gens) | 0.7778 / 0.7444 | — |
| Churn 2×2 (16k pool, varied/equal lengths) | 4 cells, 19,878 reqs, 0 errors | — |
| Heavy churn (retract × replay interleave) | 877/877, 0 errors | — |
| Graph engages + health | ✅ | ✅ |

Pre-fix IAE reproduced deterministically; gone post-fix. Graph-vs-eager on padded buckets is not bit-exact for either backend — a radix control shows the identical divergence class, so the residual is the pre-existing padded-bucket numeric noise floor, not a flat defect. 6 unit tests pin the three seams; 104 prefill-graph/flat tests green.


nperrin-fr:
Second commit extends the fix to the trtllm backend (goal: every attention backend works under flat + prefill graph), plus the full validation matrix.

**trtllm-specific bug**: its extend prewrite (fused RoPE KV write) runs inside the graphed segment, so capture bakes that forward's write-location *tensor* into the graph — every replay rewrites capture-time locations (dummy page on flat; stale input-buffer tail rows on radix). Silent corruption, no crash. Fix: `support_kv_cache_prewrite` returns False while a breakable capture is active, so the captured branch hands KV to the eager attention break (fresh per-step metadata locations), plus the same padded-tail trim as mha.

**Validation matrix (gpt-oss-120b, B200, flat scheduler ext, PFG-on vs PFG-off):**

| Gate | mha | trtllm |
|---|---|---|
| GSM8K | 0.9591 / 0.9621 (1319q) | 0.9600 / 0.9625 (400q) |
| GPQA-Diamond ×3 | 0.6953 / 0.7020 | — |
| AIME25 ×3 (64k gens) | 0.7778 / 0.7444 | — |
| Churn 2×2 (16k pool) | 4 cells, 19,878 reqs, 0 errors | — |
| Heavy churn (retract×replay interleave) | 877/877, 0 errors | — |
| Graph engages + health | ✅ | ✅ (ENGAGED, health 200) |

Graph-vs-eager on padded buckets is not bit-exact for either backend — a radix control shows the identical divergence class, so that's the pre-existing padded-bucket numeric noise floor, not a flat defect. 7 unit tests pin the three seams (mha trim, capture dummy tables, trtllm prewrite gate + trim).

PR #668, opened 2026-07-13
feat(kernel): Add batch matmul API

qedawkins:
Add a public gemm.bmm API matching the GEMM dispatch style for true batched inputs. The API is for true batched GEMMs only and is kept distinct from typical GEMM as implementations can vary in non-trivial ways depending on the size of the batch. Currently registers the torch reference implementation while performant backend implementations are left for follow-up work.

Also add out= plumbing to mm and bmm with torch reference dense GEMM/BMM write through native torch out= paths when possible. Triton GEMM paths that already store through explicit output strides write directly into out=. Gluon GEMM paths were updated to do the same as well. This is motivated by bmm callsites needing to write directly into concatenated outputs in ways that wasn't yet encountered for normal GEMM. Bias and enable_pdl handling now follow the same ownership model for mm and bmm: kernels that opt in receive those kwargs, and dispatch only applies bias for kernels without bias support.

Share the GEMM and BMM numerics input generation paths so shape/layout handling stays consistent, and add shape presets, throughput accounting, API selection coverage, generator coverage, and focused API correctness tests.

Testing: pytest tokenspeed-kernel/test/ops/test_bmm.py tokenspeed-kernel/test/ops/test_gemm.py; pytest tokenspeed-kernel/test/test_kernel_api_selection.py; pytest tokenspeed-kernel/test/test_numerics.py; pytest tokenspeed-kernel/test/test_benchmark.py -q -k 'bmm_throughput'.

### DIFF
--- a/tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/__init__.py
+++ b/tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/__init__.py
@@ -32,6 +32,7 @@ import tokenspeed_kernel.ops.gemm.trtllm  # noqa: F401
 import torch
 from tokenspeed_kernel.platform import ArchVersion, Platform
 from tokenspeed_kernel.profiling import ShapeCapture, kernel_scope
+from tokenspeed_kernel.registry import KernelRegistry
 from tokenspeed_kernel.selection import select_kernel
 from tokenspeed_kernel.signature import (
     ScaleFormat,
@@ -42,16 +43,17 @@ from tokenspeed_kernel.signature import (
 
 logger = logging.getLogger(__name__)
 
-__all__ = ["mm"]
+__all__ = ["bmm", "mm"]
 
 _platform = Platform.get()
 _fp8_dtype = _platform.fp8e4m3fn.dtype
 
-# Kernels that natively fuse a bias-vector add inside their GEMM kernel.
-# For any kernel not listed here, ``mm`` applies the bias with a post-GEMM
+# Kernels that accept and own bias application inside their GEMM wrapper.
+# For any kernel not listed here, dispatch applies the bias with a post-GEMM
 # add instead of passing it to the kernel.
 _KERNELS_WITH_FUSED_BIAS: frozenset[str] = frozenset(
     {
+        "torch_bmm",
         "torch_mm",
         "triton_mm_fp8_scaled",
     }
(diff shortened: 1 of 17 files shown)


tokenspeed-kernel/test/ops/test_gemm.py
borontion, line 1: what's the purpose of this test?

tokenspeed-kernel/test/ops/test_bmm.py
borontion, line 1: what's the purpose of this test? it seems we not really testing any kernel implementation but just using torch reference implementation here?

tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/__init__.py
```diff
@@ -245,13 +246,32 @@ def ensure_row_major_scales(
...
+
+
 def mm(
     A: torch.Tensor,
     B: torch.Tensor,
     *,
     A_scales: torch.Tensor | None = None,
     B_scales: torch.Tensor | None = None,
     bias: torch.Tensor | None = None,
+    out: torch.Tensor | None = None,
```
borontion, line 280: is there any runtime use case, for this change? basically adding a pre-allocated output buffer

qedawkins: It's not just pre-allocated, it could be strided (e.g. concatenated halves produced by two different kernels). Though there is no use for it in standard gemm today, since we needed it for batch matmul I added it here for consistency. It would be odd to carry API surface unique to batch matmul that has no reason to be.

tokenspeed-kernel/test/ops/test_gemm.py
qedawkins: Cleaned up and consolidated the file to just check a couple basic API rejection cases (out dtype mismatch, shape mismatch). AFAICT this matches the location of other similar dispatch-level check tests. Also combined with the bmm one. LMK if there is a better place for tests like this.

tokenspeed-kernel/test/ops/test_bmm.py
qedawkins: Addressed in other comment (file is dropped now).

borontion:
@qedawkins the gluon bf16 gemm registration has been removed in https://github.com/lightseekorg/tokenspeed/pull/689 due to performance issue in decode. can we also follow it to make sure not registering it until performance is fixed? just for normal mm not batch mm

qedawkins:
> @qedawkins the gluon bf16 gemm registration has been removed in #689 due to performance issue in decode. can we also follow it to make sure not registering it until performance is fixed? just for normal mm not batch mm

I kept bmm disabled too for consistency. The situations where we're faster are quite narrow for bmm also so fine to enable slowly there as well.

tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/__init__.py
```diff
@@ -245,13 +246,32 @@ def ensure_row_major_scales(
...
+    op: str,
+) -> None:
+    if tuple(out.shape) != shape:
+        raise ValueError(f"{op} out expects shape {shape}, got {tuple(out.shape)}")
+    if out.dtype != dtype:
+        raise ValueError(f"{op} out expects dtype {dtype}, got {out.dtype}")
+    if out.device != device:
+        raise ValueError(f"{op} out expects device {device}, got {out.device}")
+    if out.stride(-1) != 1:
+        raise ValueError(f"{op} out must have stride(-1) == 1")
```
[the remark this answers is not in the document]
qedawkins: Racy writes is the expected outcome in such cases. This would be an instance of API misuse, not a mishandled case. Inner unit stride is different because that's an explicit restriction of the API where we're rejecting otherwise valid output tensors to `torch.mm(..., out=)`. Overlapping outputs like this aren't well defined usage and shouldn't be accounted for. Also AFAICT no other `out=` APIs bother checking for this.

borontion (approved):

PR #698, opened 2026-07-16
perf(kernel): tune gfx950 A4W4 MXFP4 MoE for Kimi-K2.5 (decode cap + prefill tile)

panditsa:
## Summary

   This PR makes three related improvements to the gfx950 (CDNA4) A4W4 MXFP4 MoE path for Kimi-K2.5. First, it centralizes the CDNA4 MXFP4 scale-swizzle constants and byte permutation into a single `mxfp4_cdna4_scale_layout` module (removing the unused, slower `GLUON_MXFP4_LDS_BSCALE` LDS-staged path and dead reference code). Second, it raises the kernel-routing decode cap from M≤2 to M≤8 so Kimi's M=8 decode reaches the tuned MFMA decode kernels instead of the slow ragged fallback. Third, it sets a flat stage2 down-projection `BLOCK_M=128` in package prefill (dropping a redundant sort), which is 10–17% faster on the Kimi shape at M≤1024 with no gpt-oss regression.

   ## Testing

   Validated on MI350X/gfx950: the full op suite (139 tests) passes, including new bit-exact parity tests (route-owned decode M∈{1,2,3,4,8} vs MFMA reference, and package prefill BLOCK_M=128 vs the generic ragged reference) and dispatch-cap assertions; Microbenchmarks on the Kimi TP4 shard (E=384, D=7168, I=512, topk=8) confirmed the wins:

   - **Decode cap (M≤8):** with the cap raised, Kimi's M=8 decode now enters the tuned MFMA decode path (routing kernels switch from the ragged `_fused_biased_grouped_route_small_m` to `_sigmoid_bias_topk_route_gluon` + `_fused_precomputed_topk_route_small_m`), cutting per-layer MoE matmul time ~3.5× (≈1065µs → ≈305µs on TP0)  end-to-end serving throughput rose ~2.5× (20.5 -> 51.7 tok/s total) in the profiled window.
   - **Prefill stage2 BLOCK_M=128:** 10–17% faster than the old 32/64 tiers at M∈{128,256,512,1024} (e.g. M=256: 872µs → 744µs = 1.17×; M=1024: 958µs → 867µs = 1.10×), fastest at every M
 
   End-to-end, serving Kimi-K2.5-MXFP4 (TP4) on a fixed chat prompt produces byte-identical output before/after the prefill tiling change (SHA-256 match), confirming the perf changes are output-preserving.

### DIFF
--- a/tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/moe/mxfp4_cdna4_scale_layout.py
+++ b/tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/moe/mxfp4_cdna4_scale_layout.py
@@ -1,0 +1,149 @@
+# Copyright (c) 2026 LightSeek Foundation
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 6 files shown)


tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/moe/mxfp4_cdna4_scale_layout.py
```diff
@@ -0,0 +1,149 @@
...
+preprocessor and the activation-scale gather. Centralizing them here makes a
+scale-layout change a single-file edit: update the constants / permutation
+below, keep the stage-kernel scale addressing in sync, and the bit-exact
+package tests gate correctness.
+
+This module is intentionally a leaf (only depends on ``torch``) so every
+producer/consumer can import it without pulling in the Triton stage kernels.
+
+
+Why a swizzle at all
```
qedawkins:
```suggestion
Why swizzle at all
```
?

qedawkins (approved):

PR #710, opened 2026-07-17
feat(kernel): add mxfp MoE Gluon kernel on gfx1250

knwng:
initial support for gfx1250 MoE Gluon kernel

### DIFF
--- a/tokenspeed-kernel-amd/README.md
+++ b/tokenspeed-kernel-amd/README.md
@@ -1,17 +1,3 @@
 # AMD-specific High Performance Kernels
 
 This directory contains high-performance kernel implementations for AMD GPUs.
-
-## MoE kernels
-
-The MXFP/FP8 Gluon MoE kernels live under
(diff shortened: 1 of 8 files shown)


borontion:
@panditsa @antiagainst probably we need to organize files in tokenspeed-kernel-amd a little bit. it looks messy at this moment. like architecture - operation (moe / attn) - variants (moe a8w4|a4w4, mha bf16|fp8, mla fb16|fp8, dsa, etc).

antiagainst:
Yes agreed. #880 makes a good first step but I think we should further reorganize and decouple some logic, like the gigantic [`fused.py` file](https://github.com/lightseekorg/tokenspeed/blob/main/tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/moe/mxfp4/fused.py) which contains very intertwined logic. Good to break it down into multiple files, even if it means we are gonna duplicate some logic. It would make maintenance burden lower long term.

@panditsa can you take a look later?

PR #720, opened 2026-07-18
[WIP][AMD] Implement MTP support for qwen3.5 MXFP4

raikonenfnu:
## Summary

Enables Multi-Token Prediction (MTP / speculative decoding) for **Qwen3.5-397B-A17B MXFP4 on AMD**. The MXFP4 checkpoint keeps its MTP/NextN draft branch in bf16 even though the target model is MXFP4, so the draft path needs an unquantized MoE apply kernel that AMD did not previously have. This PR adds that path, wires the draft branch through the existing kernel-registry boundary, and stabilizes MoE routing so greedy MTP stays target-equivalent.

## Key changes and why

**1. AMD Triton unquant MoE draft path** (`tokenspeed-kernel/.../ops/moe/triton/unquant.py`)
- Adds a registry-selected unquant MoE apply kernel for precomputed top-k routing on AMD.
- Reuses the existing MXFP4 ragged-matmul routing flow and transposes dense expert weights into the `[expert, K, N]` layout consumed by `triton_kernels.matmul`.
- Registers TP and EP precomputed-topk variants.
- *Why:* the bf16 draft branch had no registry path on AMD; this lets the draft model load/execute through the same boundary as the rest of MoE instead of a runtime special case.

**2. Treat the mxfp4 draft branch as unquantized** (`qwen3_5_nextn.py`)
- The MTP branch is dense bf16 in both nvfp4 and mxfp4 checkpoints; treating it as quantized wrongly selected MXFP4 expert weights.
- *Why:* fixes draft-model construction for MXFP4 targets.

**3. Deterministic MoE routing under MTP** (`runtime/layers/moe/topk.py`)
- Native top-k helper now resolves exact routing ties with a stable descending sort for both normal and bias-corrected routing.
- *Why:* MTP changes target-verify batch shape from `batch_size` to `batch_size * draft_width` rows, which can perturb expert selection on exact ties. The Triton MXFP4 precomputed MoE path consumes these ids directly, so stable ids are required for MTP/non-MTP target equivalence.

**4. Disable SMG health checks under `ts serve` by default** (`cli/serve_smg.py`)
- *Why:* the gateway health probe was being scheduled as real `HEALTH_CHECK_*` generation work during long eval batches, changing batch composition while measuring deterministic greedy evals.

**5. CI eval exercises MTP** (`test/ci/eval/qwen3.5-397b-a17b-mxfp4-evalscope-aime25.yaml`)
- Adds `--speculative-algorithm MTP --speculative-num-steps 3 --speculative-num-draft-tokens 4 --speculative-draft-model-quantization unquant` and disables thinking traces in the chat template to measure the faster speculative path.

**6. Tests**
- Kernel-selection coverage + a torch-reference test for the bf16 unquant MoE layout.
- Top-k tie contract tests (exact-tie and bias-corrected) in `test/runtime/layers/test_moe_topk.py`.
- Gateway-args test for the health-check default.

## Local validation (TP=4, Qwen3.5-397B-A17B MXFP4)

- AIME25 limit-8: **score 1.0** with `enable_thinking=false` (~3.5 min) and **1.0** with thinking on (~6 min).
- MTP accept length ~3.3–3.7 of 4 draft tokens; accept rate ~0.85.
- Avg TPOT ~13 ms (thinking off) / ~15 ms (thinking on).

## Known limitations (why WIP)

- The stable-sort fix only guarantees **exact-tie** determinism. **Near-tie** batch-shape invariance is still open (identical concurrent requests can diverge in trajectory while remaining correct) — tracked as production work.
- The stable full sort is O(E log E); a deterministic O(E log K) top-k is the preferred production form.
- Only the torch-native routing path is covered; other routing paths (grouped/CUDA/flashinfer) are not yet audited to the same tie contract.
- Disabling SMG health checks is a mitigation; the health probe should instead be excluded from scheduler batch composition.

## Test plan

- [ ] TP=4 Qwen3.5 397B MXFP4 AIME25 limit-8 MTP eval meets threshold in CI
- [ ] Unit tests pass (MoE top-k tie contract, unquant MoE torch reference, kernel selection, gateway args)
- [ ] No `HEALTH_CHECK_*` generation requests during `ts serve` eval after readiness

### DIFF
--- a/python/tokenspeed/cli/serve_smg.py
+++ b/python/tokenspeed/cli/serve_smg.py
@@ -72,6 +72,7 @@ DEFAULT_SMG_POLICY = "passthrough"
 _DEFAULT_SMG_DISABLE_FLAGS = (
     "--disable-circuit-breaker",
     "--disable-retries",
+    "--disable-health-check",
 )
 
 
--- a/python/tokenspeed/runtime/layers/moe/topk.py
+++ b/python/tokenspeed/runtime/layers/moe/topk.py
@@ -196,6 +196,14 @@ def _mask_topk_ids_padded_region(
     topk_ids[indices >= num_token_non_padded, :] = -1
 
 
+def _stable_descending_topk(
+    scores: torch.Tensor,
+    topk: int,
+) -> tuple[torch.Tensor, torch.Tensor]:
+    values, indices = torch.sort(scores, dim=-1, descending=True, stable=True)
+    return values[:, :topk], indices[:, :topk]
+
+
 def torch_native_fused_topk(
     hidden_states: torch.Tensor,
     gating_output: torch.Tensor,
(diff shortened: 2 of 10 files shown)


python/tokenspeed/cli/serve_smg.py
```diff
@@ -72,6 +72,7 @@
 _DEFAULT_SMG_DISABLE_FLAGS = (
     "--disable-circuit-breaker",
     "--disable-retries",
+    "--disable-health-check",
```
borontion, line 75: @lightseek-bot @LorrinWWW is that ok we disable smg health check by default?

PR #740, opened 2026-07-20
feat(profile): Link VizTracer to Proton CPU scopes

antiagainst:
Emit a VizTracer flow at kernel scope entry using Proton's Python scope ID. During trace merging, match that ID against Proton's exported CPU event and add a rank-namespaced endpoint flow. This keeps the VizTracer-to-CPU links distinct from Proton's existing CPU-to-GPU launch flows.


antiagainst:
This allows us to associate viztracer events with proton cpu events and then gpu kernel like following (up arrow shoots out of the screen due to huge stack trace but it points to a frame in viztracer):

Screenshot 2026-07-19 at 8 07 16 PM

PR #741, opened 2026-07-20
fix: clear padded breakable graph handoffs

FlamingoPg:
## Summary

Fix padded prefill CUDA-graph replays leaking undefined eager-break output rows into the following captured graph segment.

### Reproduction

1. Capture a breakable prefill graph for a token bucket larger than the live request, for example a bucket of 8 rows.
2. Replay it for 1 real token.
3. Let an eager varlen operation return a bucket-shaped tensor while initializing only the real prefix. This matches kernels that honor live cu-seqlens but allocate an output for the captured bucket.
4. The old handoff copied that whole result into the stable destination. Rows 1-7 remained undefined, but the next graph segment consumed all 8 rows.

The added regression test makes this deterministic by filling the unwritten rows with NaN and placing a captured matrix multiplication after the eager break. Before this fix, the NaNs propagate through the padded rows. In a full model they can reach later quantization or MoE routing code and surface as an asynchronous CUDA error far from the original attention handoff.

### Fix

- Add an optional `valid_rows` argument to `BreakableCapture.replay()`.
- Have `PrefillGraph` pass the real token count before it temporarily pads the forward context to the selected bucket.
- Clear every eager-break handoff after landing its result and before replaying the following graph segment.
- Restore the previous replay state in `finally` so it cannot leak between replays.

This is a generic breakable-graph correctness fix. It does not depend on MiniMax M3, sparse attention, MoE, or flat KV cache behavior.

## Test Plan

- `python -m pytest test/runtime/execution/test_breakable_cuda_graph.py -q` (20 passed)
- `python -m pytest test/runtime/test_flat_prefill_graph.py -q` (8 passed)

### DIFF
--- a/python/tokenspeed/runtime/execution/breakable_cuda_graph.py
+++ b/python/tokenspeed/runtime/execution/breakable_cuda_graph.py
@@ -168,6 +168,7 @@ class BreakableCapture:
         self._stream_ctx: Any | None = None
         # Break-output handoff buffers keyed by (shape, dtype, device); see break_point.
         self._handoff: dict[Any, torch.Tensor] = {}
+        self._valid_rows: int | None = None
 
     @classmethod
     def current(cls) -> BreakableCapture | None:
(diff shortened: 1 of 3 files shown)


nperrin-fr (approved):
lgtm

PR #762, opened 2026-07-22
refactor(cache): make BlockRef a shared handle

wangbo981016:
## Summary

Replace intrusive cache block reference counting with a pool-managed,
`shared_ptr`-like `BlockRef`.

- Make `BlockRef` copyable and movable.
- Return blocks to the pool after the last reference is released.
- Replace raw block pointers and manual reference management.
- Acquire references only after cache matches are finalized.


nperrin-fr (approved):
lgtm

PR #765, opened 2026-07-22
perf(m3): flashinfer cute-dsl backend for MXFP8 (1,32) dense GEMM

syuoni:
The portable Triton w8a8 kernel was the only registration matching the (1,32)-block ue8m0 MXFP8 signature, so it served every MXFP8 linear of MiniMax-M3 and dominated GPU time. Register flashinfer's mm_mxfp8 cute-dsl backend (SM100/103) at SPECIALIZED priority, quantize activations online with mxfp8_quantize, and swizzle weight scales once at load in Fp8LinearMethod, which then pins the kernel via override. Sub-floor shapes and non-Blackwell platforms keep the Triton fallback through the new n_min/k_min/k_align_32 selection traits.


syuoni:
> Hi @syuoni do we need to wait for the new flashinfer release

No. This PR relies on the released version.

PR #769, opened 2026-07-22
[WIP] feat(kernel): Add validation per family/mode for kernel registration

Max191:
## Summary

Adds a new `OperationRegistry` to carry operation schemas for each family + mode combination. The purpose of the OperationRegistry is to provide validation for registered kernels and to maintain proper definitions of the semantics of each family and mode. Each operation schema has a reference implementation, which defines the kernel ABI and the semantics of its implementation. Any kernel that is registered under the family + mode matching a specific operation schema is expected to match the semantics and ABI of its reference. The operation schema also provides validation functions for traits and signatures at registration time. When a kernel is registered under a specific family + mode, its traits and signatures will be verified by the schema, ensuring that the traits and signatures match what the operation expects.

The OperationRegistry does not change any of the existing functionality, but it adds hooks during registration for validation. Currently, only mha_prefill is added to the registry as an example, but later PRs will add support for each operation.

## Test Plan

Added unit tests for the new registry and schemas.


borontion:
Hi @Max191 some high-level questions:

> Adds a new OperationRegistry to carry operation schemas for each family + mode combination.

What's the difference between this and existing `KernelRegistry`? And what's the drawback of registering the reference implementation using the same kernel api?

>  The purpose of the OperationRegistry is to provide validation for registered kernels and to maintain proper definitions of the semantics of each family and mode

I feel we are duplicating some definitions. I was expecting the op definition is in the public api, like mha prefill.

https://github.com/lightseekorg/tokenspeed/blob/633279213b3a4344a8267ca4969dc861fb6b1437/tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/__init__.py#L1027-L1044

Max191:
> What's the difference between this and existing `KernelRegistry`?

This is not registering kernels, it is registering family + mode op definitions. My goal was to create a recipe for how we define operations without reducing any of the flexibility. That's why this exists as an abstract class structure, but is simply used as a validation hook in the kernel registry.

> And what's the drawback of registering the reference implementation using the same kernel api?

It's mostly about organization. The operation definitions define what an operation actually means, and the only way to really do that is with a reference implementation, so IMO it makes sense for the reference implementations to live on the op definitions.

> I feel we are duplicating some definitions. I was expecting the op definition is in the public api, like mha prefill.

This is a registration-time validation, so I don't see an easy way to make it part of the public API.

I wanted to keep this PR small and reviewable, but eventually, we can organize the public API pieces and the operation definition pieces together. I want to move at least some of the stuff that the public API does (e.g., like creating the traits/signatures structs) into the operation contracts. Most of the public APIs actually do the same things, so if we want, we can even make a generic kernel selection + launch function using the new abstract class, and then most of the public APIs can be reduced to some simple wrappers (parts of the more complex APIs could also wrap such a function).

So, a bit bigger picture, this adds new kernel registration validation at the time of registering kernels, but it also introduces a natural place to keep a central definition of that the operation (i.e., family + mode) actually is, so we can move more logic into the op contracts to maintain a single source of truth on the kernel selection and dispatch surface for each operation.

borontion:
> I wanted to keep this PR small and reviewable, but eventually, we can organize the public API pieces and the operation definition pieces together. I want to move at least some of the stuff that the public API does (e.g., like creating the traits/signatures structs) into the operation contracts.

I think we have to keep a single place for operation definitions as a first step. you can see that there are a growing number of new apis, especially for different attention families, and it will continue to grow.

Max191:
> > I wanted to keep this PR small and reviewable, but eventually, we can organize the public API pieces and the operation definition pieces together. I want to move at least some of the stuff that the public API does (e.g., like creating the traits/signatures structs) into the operation contracts.
> 
> I think we have to keep a single place for operation definitions as a first step. you can see that there are a growing number of new apis, especially for different attention families, and it will continue to grow.

Sure, I can implement that for mha in this PR so we have an idea of how it looks.

Max191:
Closing for now. Will re-open when I have more time to finish this up and also migrate the rest of the operations to the new style.

PR #780, opened 2026-07-23
fix(multi-node): harden Qwen3.5 multi-node execution

tuanzhangCS:
Route cross-node collectives through NCCL, use overlap-safe Mamba staging, and stabilize DeepEP backend selection and NVFP4 planning.


python/tokenspeed/runtime/execution/input_buffer.py
```diff
@@ -428,30 +414,41 @@ def write_decode_input_ids(
...
+                mamba_pool_indices_cpu,
+                mamba_cow_src_indices_cpu,
+                mamba_branching_seqlens_cpu,
+                mamba_track_pool_indices_cpu,
+            ) = self._bulk_pinned(
+                (batch_size, torch.int32),
+                (batch_size, torch.int32),
+                (batch_size, torch.int32),
+                (batch_size, torch.int32),
+            )
```
tuanzhangCS, line 427: This change might introduce correctness-related bug. I will review again.

PR #791, opened 2026-07-24
fix(epd): install all encoder graph wrappers

qimcis:
## Summary
- consume multimodal models’ existing encoder CUDA-graph wrapper maps in EPD
- install each available modality wrapper instead of calling the obsolete singular builder
- route video through its capturable encoder seam

## Validation
- Python byte-compilation
- focused install and modality-routing tests added

Focused pytest collection is blocked locally by missing `pyzmq`.


qimcis:
> Hi @qimcis why do you close this pr

I didn't notice my agent had opened PRs without me noticing, i haven't finished properly testing/benchmarking this yet - will re-open it shortly once i've done so

PR #793, opened 2026-07-24
feat(kernel): support small-batch Gluon MLA decode

Max191:
## Summary

A new specialization is added for gluon mla_decode_with_kvcache targeting small batch sizes. This kernel is needed for data parallel Kimi 2.5 configurations. With data parallelism, each rank gets either 1 or 2 requests (depending on concurrency), which can become a batch of 1, 2, or 4 with speculative decoding.

The new specialization relies on splitting along kv for parallelism rather than splitting along the head dim. This strategy performed better during benchmark experiments.

## Test Plan

Validated by running the agentic benchmark on Kimi 2.5 using the attn_dp8-moe_tp8 configuration. Without the new specialization, the triton MLA kernel is selected, and MLA takes up 95% of model runtime. With the specialization, MLA is only 10%, and the agentic benchmark is able to complete in a reasonable amount of time.

New unit test cases are also added, and a new test is added to verify the specialization cases selected by the kernel.

### DIFF
--- a/tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/attention/mla/decode.py
+++ b/tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/attention/mla/decode.py
@@ -26,7 +26,7 @@ MIT-licensed ``aiter/ops/triton/gluon/mla_gluon.py`` at commit
 ``ae0bae8954110b12655e3232f68262dd63cd694e``. The implementation here retains
 TokenSpeed's paged-cache interface (page IDs plus an arbitrary page size).
 
-The shared kernel supports three regimes:
+The shared kernel supports five fixed regimes:
 
 * ``bh16bn128`` -- BF16/FP8 Q + FP8 KV, BLOCK_H=16, BLOCK_N=128 and
   ``num_q_heads <= 16`` and arbitrary batch sizes.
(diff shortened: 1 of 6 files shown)


tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/attention/gluon/mla_decode_bf16_gfx950.py
```diff
@@ -20,12 +20,17 @@
 
 """MLA decode Gluon kernels for AMD GFX950 (bf16 Q + bf16 KV).
 
-Two regimes share one kernel, dispatched by ``num_q_heads``:
+Two compute regimes share one kernel, dispatched by ``num_q_heads``:
```
Yu-Zhewen: Thanks @Max191! Instead of introducing `launch_policy` and `GRID_MODE`, Would it be cleaner to register `bh16bn64`, `bh64`, `bh16-multiblock`, and `bh64-small` as separate kernels/regimes? What do you think

Max191: Yeah, I'll go with this so we're not mixing approaches. It matches the current implementation better.

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/attention/gluon/mla_decode_gfx950.py
```diff
@@ -26,14 +26,16 @@
...
+The shared kernel supports five fixed regimes:
 
 * ``bh16bn128`` -- BF16/FP8 Q + FP8 KV, BLOCK_H=16, BLOCK_N=128 and
   ``num_q_heads <= 16`` and arbitrary batch sizes.
 * ``bh16bn64`` -- BF16 Q + BF16 KV, BLOCK_H=16, BLOCK_N=64,
   ``num_q_heads <= 16``.
 * ``bh64`` -- BLOCK_H=64, BLOCK_N=64, ``num_q_heads in {64, 128}``, 3-D
   XCD-aware grid, ``batch_size`` divisible by 64.
+* ``bh16-multiblock`` -- BF16 Q/KV, BLOCK_H=16, 64 heads, small-batch 3-D grid.
+* ``bh64-small`` -- BF16 Q/KV, BLOCK_H=64, 64 heads, small-batch 3-D grid.
```
Yu-Zhewen:
```suggestion
* ``bh16-multiblock`` -- BF16 Q/KV, BLOCK_H=16,
  ``num_q_heads == 64``, ``batch_size == 1``, 3-D grid.
* ``bh64-small`` -- BF16 Q/KV, BLOCK_H=64,
  ``num_q_heads == 64``, ``batch_size in {2, 4}``, 3-D grid.
```

Yu-Zhewen (dismissed):
LGTM, thanks!

Yu-Zhewen (approved):

PR #804, opened 2026-07-25
feat: jenga two level allocation

wangbo981016:
## Two-level cache management

The previous Flat cache layout used padding to equalize the physical storage
owned by heterogeneous cache groups. That physical block size was also exposed
to the scheduler as the group's logical block size, coupling three independent
concerns:

- the number of tokens covered by a cache key and prefix match;
- the number of logical cache blocks packed into one physical allocation; and
- the byte layout required by attention, SWA and recurrent-state kernels.

For hybrid models such as Qwen3.5, one recurrent-state checkpoint is much
larger than one Full-KV page. Propagating that byte ratio into scheduling either
coarsens prefix matching or requires separately sized pools, which wastes
capacity through fragmentation.

This PR separates those concerns:

- `P` is the domain-wide logical `CacheBlock` size used for allocation, hashing
  and prefix matching.
- `K_g` is the number of `CacheBlock`s from group `g` that fit in one
  equal-sized physical `LcmBlock`.

All groups therefore retain the same fine-grained prefix semantics while
sharing one physical arena. A cache hit is still an ordinary whole
`CacheBlock` hit, so the design does not require partial-block hits or
copy-on-write.

The current model recipes use `P=128`. This page size has been validated end to
end by the integrated attention, state, MTP and PD paths. It also keeps one
common contract for model families whose kernel layouts require 128-token
pages, including the interleaved MXFP8 scale layout. It is not an architectural
minimum. The planner and C++ scheduler are parameterized by `P`, and a smaller
Qwen3.5 geometry is mathematically valid. For example, under the validated TP4
FP8 geometry, changing `P` from 128 to 64 would change `K_full` from 32 to 64
while the State groups remain at `K_state=1`. Using such a configuration in
production still requires every participating attention and page-table layout
to support the smaller page, followed by new end-to-end correctness and
performance validation. A smaller `P` also increases page-table, hash,
metadata and checkpoint frequency.

## Implementation

lcm-python-memory-cpp-contract-en

**Python/runtime is byte-aware.**

- Model recipes declare real fields, planes, shapes, dtypes and stride
  constraints.
- `prepare_lcm_setup` applies the model recipe and derives the usable parent
  count from the cache budget after accounting for fixed workspace.
- `plan_lcm_fields` solves one pool's per-group packing ratio `K_g`, physical
  `lcm_block_bytes`, planes, field offsets and page strides using integer
  geometry.
- `LcmCachePool` performs one backing allocation for the plan and exposes stable
  typed tensor views over it.
- Adding another supported hybrid model is limited to describing its field
  recipe and kernel constraints; the planner and scheduler contract remain
  model-independent.

For the validated Qwen3.5-397B TP4 FP8 configuration:

```text
P = 128 tokens
lcm_block_bytes = 31,457,280
K_full = 32
K_state_0/1/2 = 1
N_lcm = 3,805
```

The 60 model layers repeat one three-GDN-plus-one-Full pattern 15 times.
`unit` is the occurrence index inside each cache group:

| Layer | Cache group | Group-local unit | Plane A view | Plane B view |
|---:|---|---:|---|---|
| 0 | `linear_attention_0` | 0 | `unit.0.a` (SSM) | `unit.0.b` (conv) |
| 1 | `linear_attention_1` | 0 | `unit.0.a` (SSM) | `unit.0.b` (conv) |
| 2 | `linear_attention_2` | 0 | `unit.0.a` (SSM) | `unit.0.b` (conv) |
| 3 | `full_attention` | 0 | `unit.0.a` (K) | `unit.0.b` (V) |
| 4 | `linear_attention_0` | 1 | `unit.1.a` (SSM) | `unit.1.b` (conv) |
| 5 | `linear_attention_1` | 1 | `unit.1.a` (SSM) | `unit.1.b` (conv) |
| 6 | `linear_attention_2` | 1 | `unit.1.a` (SSM) | `unit.1.b` (conv) |
| 7 | `full_attention` | 1 | `unit.1.a` (K) | `unit.1.b` (V) |
| ... | ... | ... | ... | ... |
| 56 | `linear_attention_0` | 14 | `unit.14.a` (SSM) | `unit.14.b` (conv) |
| 57 | `linear_attention_1` | 14 | `unit.14.a` (SSM) | `unit.14.b` (conv) |
| 58 | `linear_attention_2` | 14 | `unit.14.a` (SSM) | `unit.14.b` (conv) |
| 59 | `full_attention` | 14 | `unit.14.a` (K) | `unit.14.b` (V) |

Equal plane names are intentional aliases, not separate simultaneous
allocations. Every physical parent contains 15 Plane-A/Plane-B pairs. When a
parent is bound to `full_attention`, those bytes are viewed as K/V pages; when
it is bound to one State group, the same byte ranges are viewed as SSM/conv
checkpoints. The model recipe describes that reinterpretation, so neither the
generic planner nor the C++ scheduler needs per-layer Qwen3.5 branches.

One physical parent can therefore hold 32 independent Full-KV `CacheBlock`s,
while one State `CacheBlock` occupies a whole parent. Both still allocate,
hash, match and evict at the same 128-token logical grain.

**The C++ scheduler is byte-blind.**

- `ProbePrefix` performs read-only matching across all groups and converges on a
  common resumable boundary. `Admit` consumes that probe. It rejects before
  committing when capacity is unavailable, then claims hits, evicts victims and
  allocates pages. An internal mismatch after commit starts is fatal rather than
  rolled back.
- Each `KvCacheManager` owns a stable list of logical cache entries for each
  pool. Non-owning `by_key` and `by_location` indexes provide lookup, while the
  entries hold recency and boundary metadata. The Manager owns the group's
  retention policy.
- `CacheBlockRef` provides shared-pointer-like RAII ownership across manager
  cache entries and request block tables. Destroying the last reference
  releases the physical slot.
- `BlockPool` tracks one `LcmBlock` placement record for each equal-sized
  physical parent. It owns no backing bytes, cache keys, recency metadata or
  `CacheBlockRef`s. When the last child is released, the parent becomes unbound
  and can be reused by any group.

**`CacheBlockRef` ownership flow.**

The Manager cache and request `BlockTable` share the same non-atomic,
shared-pointer-like `CacheBlockRef`, so active requests naturally pin cached blocks.
Destroying the final reference destroys the `CacheBlock` and returns its physical
slot to `BlockPool`; no separate ownership state machine is needed.

cacheblockref-ownership-transfer-en-v2

**Qwen3.5 cache and eviction policy.**

Qwen3.5 uses one Full-KV group and three State groups. All four groups allocate,
hash and match in `P = 128` token `CacheBlock`s, even though a physical
`LcmBlock` packs 32 Full-KV blocks but only one State block.

- Full KV is prefix-closed: every completed 128-token page is published, and a
  hit must form a contiguous prefix. Within one request epoch, eviction starts
  from the longest suffix, preserving the shorter prefix that is more likely to
  be shared again.
- A State block is a resumable checkpoint rather than token history. Each State
  group publishes the single checkpoint page required at an aligned prefill
  chunk boundary, an aligned request endpoint, or a boundary promoted because
  Full KV found a longer reusable prefix.
- A newly created ordinary chunk checkpoint starts as probationary. A real
  prefix acquisition proves reuse and updates its access epoch; endpoint and
  promoted checkpoints are retained as established boundaries immediately.
- Admission considers all groups together. It protects blocks selected by the
  current prefix hit, then evicts the oldest request epoch first. Within the
  same epoch, request-only uncached blocks go first, followed by shallow
  probationary State checkpoints, established State checkpoints, and finally
  Full-KV pages from suffix to prefix.

MTP and PD use the same `P`/`K_g` placement contract, while the existing Radix
path remains unchanged.

**Validation.**

Qwen3.5-397B-A17B-NVFP4 was evaluated on GSM8K 4-shot with 200 samples:

| Configuration | `mean_acc` | Notes |
|---|---:|---|
| Flat LCM | 0.980 | Prefill and decode CUDA Graph enabled |
| Radix | 0.975 | Same-machine control |
| Flat LCM + MTP | 0.990 | Prefill and decode CUDA Graph enabled |
| Flat LCM, single-node PD | 0.975 | P eager, D decode CUDA Graph enabled |

At the same Qwen3.5 memory setting, Flat exposed 15,585,280 addressable logical
tokens versus 8,205,632 for Radix (`1.899x`). This is the physical
layout's maximum addressable capacity, not a claim that every workload keeps
`1.899x` more complete multi-group prefixes resident.

The earlier GPU capacity regression
`test/runtime/test_slab_capacity_prefix_hits.py` also checks that extra physical
capacity becomes useful cache residency under a fixed byte budget. Its
same-commit gpt-oss A/B measured 168,832 versus 84,416 tokens (`2.00x`);
the second-pass prefix-hit ratio was about 98.7% with the shared layout and 0%
with the forced legacy layout (`1 passed`, 63 seconds).


python/tokenspeed/runtime/layers/attention/backends/hybrid_linear_attn.py
minedec: Hi, when using attn tp4dp2 + moe ep8 on Qwen3.5, I get the error: `AttributeError: 'NoneType' object has no attribute 'update_current_inputs_after_verify'`. In flat mode, the linear backend doesn't create a SimpleMambaPool, so it should return early before calling `update_current_inputs_after_verify`.
I've added a check condition below that works on both tp8 and tp4dp2ep8.

```
def update_mamba_state_after_mtp_verify(self, accepted_length, model):
        # Flat contract pool: the per-position verify states live in the
        # scratch; commit the accepted position back to the state slab.
        commit = getattr(self.linear_attn_backend, "flat_commit_verified_state", None)
        if commit is not None and getattr(
            self.linear_attn_backend, "_verify_commit_ctx", None
        ) is not None:
            commit(accepted_length)
            return
        # mamba_cache_indices are input rows during target-verify. The first
        # output row is always the scheduler-owned working slot, so use the
        # output index table to update the next-round input pointer.
        output_indices = self.linear_attn_backend.forward_metadata.mamba_output_indices
        if output_indices is None:
            return
        req_pool_indices = (
            self.linear_attn_backend.forward_metadata.mamba_req_pool_indices
        )
        if req_pool_indices is None:
            return
        pool = getattr(self.linear_attn_backend, "pool", None)
        if pool is None:
            if getattr(self.linear_attn_backend, "flat_state_active", False):
                return
            raise RuntimeError(
                "MTP verify produced mamba output indices but the linear "
                "attention backend has no SimpleMambaPool and flat state "
                "paging is inactive"
            )
        request_number = accepted_length.shape[0]
        pool.update_current_inputs_after_verify(
            req_pool_indices[:request_number],
            output_indices[:request_number],
            accepted_length,
        )
```

python/tokenspeed/runtime/layers/attention/lcm_setup.py
```diff
@@ -0,0 +1,688 @@
...
+    cache_budget_bytes: int,
+    decode_input_tokens: int,
+    overlap_schedule_depth: int,
+) -> LcmSetup:
+    """Apply one model recipe and size target/draft arenas from one budget."""
+    if server_args.block_size != _LOGICAL_BLOCK_TOKENS:
+        server_args.block_size = _LOGICAL_BLOCK_TOKENS
+    attn_config.page_size = 64
+    if draft_attn_config is not None:
+        draft_attn_config.page_size = 64
```
minedec, line 595: I try TP4DP2EP8 on qwen3.5, causing RuntimeError   
```
[ts]   File "tokenspeed/python/tokenspeed/runtime/layers/attention/backends/trtllm.py", line 458, in init_forward_metadata
[ts]     self._init_decode_metadata(bs, req_pool_indices, seq_lens, req_to_page)
[ts]   File "tokenspeed/python/tokenspeed/runtime/layers/attention/backends/trtllm.py", line 524, in _init_decode_metadata
[ts]     else self._build_page_table(
[ts]          ^^^^^^^^^^^^^^^^^^^^^^^
[ts]   File "tokenspeed/python/tokenspeed/runtime/layers/attention/backends/trtllm.py", line 218, in _build_page_table
[ts]     page_table_buf[:bs].copy_(
[ts] RuntimeError: The size of tensor a (4096) must match the size of tensor b (2049) at non-singleton dimension 1
```
Both attn_config.page_size and draft_attn_config.page_size are set to 64, while server_args.block_size uses _LOGICAL_BLOCK_TOKENS (128). This discrepancy causes the page_table_buf (sized by attn_config.page_size) to diverge from req_to_page (sized by server_args.block_size) in both target and draft attention backends.

wangbo981016: looking into it

zhyncs: @minedec @wangbo981016 Let's fix in the followup.

PR #820, opened 2026-07-27
feat: wire flashinfer autotuner

syuoni:
## Summary

Opens a single autotune window during engine startup — one dummy prefill at
`chunked_prefill_size` tokens, before CUDA-graph capture — and makes it the
**only** place kernel tuning can happen. Serving is a pure cache-lookup /
heuristic-fallback path by construction.

### Changes

- **`ops/tuning.py`**: `autotune()` context manager + `set/get_autotune_max_num_tokens()`;
  `ModelExecutor._autotune()` runs the dummy prefill under it. Flashinfer skips
  cache hits, so shapes seeded from a pre-swept table cost nothing.
- **`nvfp4_gemm_swiglu_nvfp4_quant`** (CuteDSL) joins the autotune framework:
  12 `(mma_tiler, cluster)` tuple tactics with a `TunableRunner`, replacing the
  fixed heuristic.
- **Cleanup**: drop `autotune_frozen`/`freeze_autotuning` (a switch to suppress
  serving-time tuning admits serving-time tuning is possible — it no longer is),
  the MXFP4 SiTU lazy in-situ tuning pass (carried the TP rank-divergence
  deadlock hazard the window removes), and the tuning-cache-active flag.
  Pre-swept table loads become fire-and-forget cache seeding. The sweeper and
  all MoE ops now key buckets from `get_autotune_max_num_tokens()`.

### Measured (B200)

- MiniMax-M3: MXFP8 GEMM **1.22×** overall / **1.68×** decode; NVFP4 MoE **1.11×**
  (tuned vs. default tactic, nsys kernel-level, cold L2).

### Notes

- Requires flashinfer ≥ 0.6.16: startup tuning of the trtllm NVFP4 MoE exposed
  an OOB read in the trtllm-gen dynB kernels' route map (IMA on Kimi-K2.5),
  root-caused with compute-sanitizer and fixed upstream in
  flashinfer-ai/flashinfer#4237 (cherry-picked into v0.6.16).
- Tables swept at `tune_max ≠ runtime` no longer match bucket keys; the default
  (8192 floor) covers all `chunked_prefill_size ≤ 8192` deployments.


syuoni:
Waiting for https://github.com/flashinfer-ai/flashinfer/pull/4237

syuoni:
The autotuner increases the server startup time by 3~5 mins. @zhyncs

borontion (approved):

PR #834, opened 2026-07-29
[WIP] perf(k3): optimize decode

nperrin-fr:


nperrin-fr:
this is one is probably too big and I'll break it up later.

nperrin-fr:
this will be broken up into smaller PRs, keep the branch for reference

PR #847, opened 2026-07-30
fix(kimi3): correct AMD KDA safe gate

panditsa:
## Summary

  Fix the Kimi K3 safe-gate calculation in the AMD Triton KDA prefill and decode paths.

  The kernel previously applied the standard softplus gate and clamped it to the lower bound. K3 instead uses:

  ```python
  lower_bound * sigmoid(exp(A_log) * gate_input)
```
  This now matches the FLA reference implementation (https://github.com/fla-org/flash-linear-attention/blob/0f0f0c97af39343855b43bbbaddcedfda5cb9d77/fla/ops/kda/gate.py#L58-L70).

  ## Tests

  - Added PyTorch references for the K3 safe gate and recurrent KDA update.
  - Added coverage that distinguishes the safe gate from the old clamped-softplus calculation.
  - Compared the AMD recurrent kernel output and final state against the reference.

### DIFF
--- a/tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/triton/kda.py
+++ b/tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/triton/kda.py
@@ -64,11 +64,12 @@ def _kda_prepare_gate_beta_kernel(
         other=0.0,
     ).to(tl.float32)
     x += bias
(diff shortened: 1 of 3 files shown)


raikonenfnu (approved):
LGTM!

PR #850, opened 2026-07-30
fix(flatkv): resolve decode starvation and drafter page table crash

minedec:
### Problem
Two independent defects surface only on FlatKV builds.

**1. Long decodes stall forever with an idle GPU.**
`AcquireForRequest` is guarded by `#if !TOKENSPEED_FLAT_KVCACHE`, so on flat builds the paged-cache adjunct never acquires a page and a request never gets an adjunct table. `AdmitChunk` had no such guard, so it kept bounding every chunk by a budget it never draws from: `absolute_have` stays 0 while `required` is derived from the whole history. Once the history exceeds `total_pages * tokens_per_page` the request is deferred every round forever.
Observed on Qwen3.5 TP4DP2EP8 + MTP: all three GDN state groups starved at `free=303, required=304, target_tokens=19398` (= 303 x 64), so every AIME25 sample generating past ~19.4k tokens hung until the client timed out.

**2. The MTP drafter crashes when kernel pages are smaller than logical pages.** On EXTEND the trtllm backend prefills the drafter's step-1+ decode metadata, but that call dropped the flat per-group tables, so `_init_decode_metadata` fell back to the radix `req_to_page` gather. On flat LCM builds that table is unpopulated and laid out in logical pages (P=128 -> 2049 columns) while the backend's buffer is kernel-paged (64 -> 4096 columns), failing with `RuntimeError: The size of tensor a (4096) must match the size of tensor b (2049)`. Every sibling call site already forwards the tables, as does the MHA backend at the same spot.

### Changes

- `forward.cpp`: gate all four `AdmitChunk` call sites to radix builds, pairing them with `AcquireForRequest`. Flat placement remains the only gate that can trigger retraction or OOM terminalization, as `flatPoolWedged` intends.
- `trtllm.py`: forward `flat_page_tables` / `flat_out_cache_locs` into the drafter's decode-metadata prefill, mirroring the MHA backend.
- `test_flat_kvcache_scenarios.cpp`: `ReserveRefundBalances` needs one more planning round now that flat placement, not the adjunct, records the rejection that starts wedge escalation.

## Test Plan
- AIME25 TP4DP2EP8 + MTP completed 30/30 (0.8667 -> 0.9333, no `ReadTimeout`) and a single 22.6k-token request finished cleanly.

### DIFF
--- a/python/tokenspeed/runtime/configs/paged_cache_spec.py
+++ b/python/tokenspeed/runtime/configs/paged_cache_spec.py
@@ -191,26 +191,6 @@ def hybrid_slab_group_size(
     return max(counts.values())
 
 
-def _validate_kernel_page_size(
-    kv_pool: object, attn_backend: object, pool_name: str
-) -> None:
-    """Reject a kernel page size that disagrees with the LCM plan."""
-    plan = getattr(kv_pool, "_lcm_memory_plan", None)
-    if plan is None:
-        return
-    plan_tokens = int(plan.logical_block_tokens)
-    backend = _flat_kv_backend(attn_backend)
-    page_size = getattr(backend, "page_size", None)
-    if page_size is None or int(page_size) == plan_tokens:
-        return
-    raise RuntimeError(
-        f"attention backend {type(backend).__name__} uses page_size="
-        f"{int(page_size)} but KV pool {pool_name} plans P={plan_tokens} tokens "
-        "per logical block; the backend's page table and req_to_page would be "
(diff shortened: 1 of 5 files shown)


minedec:
@claude review

minedec:
Need to wait for https://github.com/lightseekorg/tokenspeed/pull/851 to be merged before merging this PR.

tuanzhangCS (approved):

PR #854, opened 2026-07-30
fix(runtime): reject ambiguous prompt sources

ShiroKSH:
## Summary

Reject requests that specify more than one prompt source rather than silently discarding one.

## Root cause

The request normalizer only rejected zero and three prompt sources. Any pair of text, input_ids, and input_embeds reached runtime with one value ignored.

## Testing

- Full repository pre-commit suite.
- Python syntax compilation.
- Direct validation of each pairwise conflicting prompt-source combination.


LorrinWWW (approved):
We can check later if multiple inputs are required in any cases, but tightening this constraint looks good to me.

PR #857, opened 2026-07-30
[WIP] perf(kda): CuTe DSL fused decode for KDA linear attention

nperrin-fr:
Replaces the Triton single-token decode megafusion with a CuTe DSL kernel. The Triton program committed its conv window in place and relied on the NV column-split programs staying co-resident to order that write against its siblings' reads; that stops holding once the grid exceeds one wave, and double-shifted windows reproduce at large decode batches (run-to-run nondeterminism from B=192 at the K3 TP8 geometry). The CuTe kernel orders the shift with a cluster barrier, so correctness no longer depends on occupancy.

One launch per (sequence, value-head) does the 4-tap causal conv + SiLU, the low-rank decay-gate GEMV, QK L2 normalization, and the delta-rule recurrence against the paged fp32 state. At NV=1 the state slab is staged by a single bulk-TMA copy and stays smem-resident across both recurrence passes; at NV>1 it streams per-row cp.async into registers. PDL is wired so the state stream, conv taps and conv windows all issue before griddepcontrol_wait, hiding behind the producer GEMV's tail.

The verify (MTP) path keeps its own Triton megafusion, which writes per-position windows to its own scratch and never shared the race.

Validation on 8x B300 (K3 TP8, non-spec): GSM8K 0.9700, bs=1 decode 93.7 tok/s vs 93.3 on main, 5005-request mixed soak with 0 failures, 47/47 attention kernel tests, and compute-sanitizer racecheck / memcheck / initcheck clean over the decode sweep.


nperrin-fr:
hold for this one and there will be verify kernel later. we can review when both are ready.

nperrin-fr:
low priority for now, but will revisit later

PR #873, opened 2026-07-31
feat: improve multi-node usability

syuoni:
## Summary

Multi-node `tokenspeed serve` needed three topology flags, `NCCL_*`/`GLOO_*`
exports, and a different entrypoint on followers. This PR removes all of it —
under a Slurm step the same command works on every node:

```bash
srun --nodes=2 --ntasks-per-node=1 tokenspeed serve \
    --model <model> --attn-tp-size 8 --moe-tp-size 8 ...
```

### Commits

- `ea44c05` — `ts serve` runs the engine directly on `node_rank >= 1` instead
  of parking on a dummy HTTP server the gRPC gateway can never probe.
- `f89fc6d` — new `runtime/utils/launcher.py` derives `nnodes`, `node_rank`,
  `dist_init_addr` from the Slurm step env. Flags win but must agree; a
  contradiction is a hard error.
- `15c31cf` — no control-plane port scan when `nnodes > 1`; every node was
  probing its own wildcard address for ports that exist only on rank 0.
- `78d0e82` — derives `GLOO_SOCKET_IFNAME`/`NCCL_SOCKET_IFNAME` from the route
  to the head node.
- `c260baa` — locks the `snapshot_download` calls in `get_config` /
  `get_tokenizer` / `prepare_model_and_tokenizer`.

### Notes

- Slurm-only, evidence-gated: no MPI/PMI fallback chain. Engages only when
  `SLURM_NNODES > 1`; unresolvable topology fails loudly rather than falling
  back to single-node.
- Measurement corrected the assumption: `NCCL_SOCKET_IFNAME` and `NCCL_IB_HCA`
  are *not* needed — NCCL selects correctly unaided. `GLOO_SOCKET_IFNAME` is,
  and no prior note mentioned it: gloo binds whatever the hostname resolves to
  (`127.0.1.1` here), so cross-node gloo fails without it.
- `c260baa` is hardening, not a bug fix — the tree-cache race was reported but
  never reproduced. It does close a gap: `get_lock` previously wrapped only
  weight downloads, while vLLM also locks tokenizer resolution.
- Rendezvous port is the constant 23456. Deriving it from `--port` is wrong:
  under `ts serve` that is a per-node `get_free_port()`.

### DIFF
--- a/docs/serving/parallelism.md
+++ b/docs/serving/parallelism.md
@@ -75,6 +75,44 @@ tokenspeed serve <model> \
 Each node must use the same model, backend, precision, and scheduler settings.
 Only `--node-rank` should differ between nodes.
 
+Run one `tokenspeed serve` per node. Node rank 0 serves the HTTP API; higher
+ranks run the engine only and expose no endpoint.
+
+### Under a launcher
+
+Inside a multi-node Slurm step, `--nnodes`, `--node-rank` and
+`--dist-init-addr` are all derived from the step environment when they are not
+given, so the same command line runs on every node:
+
+```bash
+srun --nodes=2 --ntasks-per-node=1 tokenspeed serve <model> --attn-tp-size 16
+```
+
+| Argument | Derived from |
+| --- | --- |
+| `--nnodes` | `SLURM_NNODES` |
+| `--node-rank` | `SLURM_NODEID` |
+| `--dist-init-addr` | first host of `SLURM_STEP_NODELIST`, port 23456 |
+
+Rules:
+
+- An explicit `--nnodes` or `--node-rank` that contradicts the environment is
+  an error, not an override. Omit the flag to accept the launcher's value.
+- An explicit `--dist-init-addr` is always used as given.
+- Derivation only engages when `SLURM_NNODES` is greater than 1. Outside a
+  launcher, or in a single-node step, behaviour is unchanged.
+- If a multi-node step is detected but the topology cannot be resolved,
+  startup fails with the reason rather than falling back to a single node.
+- The derived address is the one the head node's hostname resolves to. Where
+  that is not the interface you want carrying NCCL bootstrap traffic, set
+  `--dist-init-addr` explicitly; `NCCL_SOCKET_IFNAME` still has to be set
+  separately.
+- The rendezvous port is a fixed constant, not a function of `--port`. Every
+  node has to arrive at the same port without talking to any other node, and
+  under `tokenspeed serve` the engine's own port is allocated per node. The
+  constant also stays clear of the kernel's ephemeral range, which is checked
+  at startup. Pass `--dist-init-addr` to use a different port.
+
 Apply the same NCCL transport and channel settings on every node as well. In
 particular, do not mix IB and Socket selection or different
 `NCCL_MIN_NCHANNELS` / `NCCL_MAX_NCHANNELS` values across ranks.
--- a/docs/serving/parallelism.md
+++ b/docs/serving/parallelism.md
@@ -104,9 +104,15 @@ Rules:
 - If a multi-node step is detected but the topology cannot be resolved,
   startup fails with the reason rather than falling back to a single node.
 - The derived address is the one the head node's hostname resolves to. Where
-  that is not the interface you want carrying NCCL bootstrap traffic, set
-  `--dist-init-addr` explicitly; `NCCL_SOCKET_IFNAME` still has to be set
-  separately.
+  that is not the interface you want carrying bootstrap traffic, set
+  `--dist-init-addr` explicitly.
+- `GLOO_SOCKET_IFNAME` and `NCCL_SOCKET_IFNAME` are set from the interface that
+  routes to the head node, unless already present in the environment. Gloo
+  needs this: it has no peer-address heuristic and otherwise binds whatever the
+  local hostname resolves to, which is a loopback entry on many hosts. NCCL
+  normally selects correctly on its own; it is set for consistency.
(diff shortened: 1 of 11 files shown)


nperrin-fr (commented):
Reviewed on the GB200 cluster where we independently hit every problem this PR fixes (GLOO_SOCKET_IFNAME binding loopback, follower entrypoints, hand-derived rendezvous addresses) — the design calls are right: contradiction-is-an-error instead of flag override, fail-loud on unresolvable topology, the constant rendezvous port with the honest TODO, and especially the port-scan analysis (followers scanning their own wildcard and silently diverging from the head was a real hang class; taking the cluster as derived and erroring on the head is the correct shape).

All 29 launcher unit tests pass here (aarch64/GB200). One confirmed issue and a few smaller ones:

**P1 — batch-script env false-positive → silent rendezvous hang (confirmed, repro below).** `detect_topology` gates on `SLURM_NNODES` and falls back to `SLURM_JOB_NODELIST`. Both are set in the **batch script env of a multi-node sbatch, outside any srun step** — measured on this cluster:

```
# inside `sbatch -N2 --wrap 'env | grep …'` (no srun):
SLURM_JOB_NODELIST=inkwell-iron-cn[02-03]
SLURM_NNODES=2
SLURM_NODEID=0            # no SLURM_STEP_* at all
# inside `srun -N1` within the same job:
SLURM_NNODES=1            # step-scoped, safe
SLURM_STEP_NUM_NODES=1
```

Feeding the measured batch env to `detect_topology` derives a topology instead of returning `None`:

```python
>>> detect_topology({'SLURM_NNODES':'2','SLURM_NODEID':'0',
...                  'SLURM_JOB_NODELIST':'inkwell-iron-cn[02-03]'})
LauncherTopology(nnodes=2, node_rank=0, head_host='10.40.1.87', dist_init_port=23456, ...)
```

So a plain `ts serve` run directly in a multi-node batch script — a deliberate single-node serve inside a bigger allocation (serve on node 0, bench clients on the rest, a common pattern) — silently derives `nnodes=2` and parks at the rendezvous waiting for a follower that will never come. That's the one failure mode the PR set out to exclude ("unresolvable topology fails loudly"): this one resolves cleanly and hangs.

Suggested fix: gate on step-scoped evidence only — `SLURM_STEP_NUM_NODES > 1` (present only inside srun steps) and drop the `SLURM_JOB_NODELIST` fallback. The batch env then returns `None` (unchanged single-node behaviour), and `srun -N2` still derives everything. The docs table only mentions `SLURM_STEP_NODELIST` already, so this also brings the code in line with the docs.

**P2 — ephemeral-range check fires for the constant port even when `--dist-init-addr` is explicit.** `detect_topology` calls `check_dist_init_port(DIST_INIT_DEFAULT_PORT)` unconditionally, but `resolve_launcher_topology` only uses the derived port when `dist_init_addr is None`. On a host with a lowered `ip_local_port_range` covering 23456, a user who already passes `--dist-init-addr` gets a hard error about a port they never use. Move the check to the branch that actually adopts the derived address.

**P2 — `_get_from_args` takes the *first* occurrence; the old `_user_host_port_from_gateway_args` loop took the *last*.** Argparse semantics are last-wins, and append-to-override is a common wrapper pattern. `args.index(flag)` flips that for duplicated flags. `len(args) - 1 - args[::-1].index(flag)` (or an explicit reverse scan) keeps the old semantics.

**P3 — `_local_routable_ipv4` probes `8.8.8.8`,** which assumes a default route; on an air-gapped cluster the UDP connect raises and the rank-0-loopback path turns into a hard error even though a routable interface exists. Probing another host from the step nodelist (any peer works — no packet is sent) uses what's already at hand.

**Nit** — follower `os.execv` drops the `ENGINE_TAG` stream prefixing, so follower logs lose the tag rank-0 engine lines carry. Cosmetic; a note in the docs would do.

None of the smaller points block. With the P1 gate switched to step-scoped variables this is a clear improvement over the status quo — our own sbatch templates carry three topology flags plus two `*_SOCKET_IFNAME` exports and a hand-rolled `MASTER_IP` resolution (with the same loopback pitfall this PR documents), all of which this deletes.

syuoni:
> Reviewed on the GB200 cluster where we independently hit ..., all of which this deletes.

Thanks @nperrin-fr  — the P1 repro was decisive. Our tooling always launches via `srun`,
where `SLURM_NNODES` is rewritten to the step's count, so we could never hit
it; the unit tests fed synthetic envs that always looked like a step.

Fixed in the latest push:

- **P1** — gate is now `SLURM_STEP_NUM_NODES`, nodelist from
  `SLURM_STEP_NODELIST` only, `SLURM_JOB_NODELIST` fallback gone. The real
  defect was mixed scoping: job-scoped gate, step-scoped nodelist, and
  `SLURM_NODEID` set in both contexts (reading `0` in the batch script), which
  made the false positive look like a coherent rank-0-of-2. Docs updated; the
  two fallback tests are replaced by a case built from your measured batch env.
  Deliberately dropped: multi-node launches outside an srun step (`mpirun` in
  `sbatch`) are no longer detected — that path was already broken, since
  `SLURM_NODEID` forwards as `0` to every rank.
- **P2 (port check)** — moved to the branch that adopts the derived address. It
  was slightly worse than described: the CLI now calls `detect_topology()` just
  to learn its rank, so the gateway could die on a port it never binds.
- **P2 (`_get_from_args`)** — reverse scan, last-wins restored, with a test.
- **P3** — agreed, deferred: probing a peer needs a second host from the
  nodelist and `first_host` returns us, so it needs an expander we don't have.
- **Nit** — correct; `os.execv` leaves nothing to prefix the stream. Doc note.

PR #876, opened 2026-07-31
chore(kimi-k3): serve on GB200 -- LCM packing, layer derivation, kda la…

nperrin-fr:
…bel, subgroup timeout

Four fixes surfaced by bringing Kimi-K3 up on GB200 (aarch64 Grace + B200-class sm_100, 184 GB HBM per GPU, 4 nodes):

* ~~LCM packing: the KDA state page scales as 1/attn_tp while the MLA latent page is tp-invariant, so the fixed FULL_ATTENTION packing of 12 only lines up at tp=8 (padding 13.4%); at tp=16 the same block pads the halved KDA page by 126.8% and planning aborts. Scale the MLA-side packing with tp (12*8/tp: tp=8 unchanged, tp=16 -> 6, grain 1536 -> 768 tokens) so the block/KDA ratio -- and the padding -- stay at the tp=8 value. K3 needs tp=16 on GB200: the full checkpoint does not fit 8x184 GB, and attn-DP shapes with a TP group spanning nodes are not wired in the DP controller.~~

* Layer counts: derive the KDA/MLA split from the config instead of hardcoding 93/69/24, keeping the exact-partition and group-divisibility checks. Structurally-identical reduced-layer checkpoints (same per-layer specs) now plan one plane per MLA layer; used for cheap 2-node crash-clearing runs of the full kernel stack.

* KDA prefill label: the runtime policy label for the portable fla-backed scan is "fla", but the registered NVIDIA solution is named "triton" (it wraps fla's chunk_kda). B300 CI always has flashkda/cutedsl_kda installed so the label was never exercised there; on GB200 neither specialized package exists and the unmapped label raised NoKernelFoundError during prefill graph capture. Map it like the AMD shim does.

* Subgroup timeouts: dist.new_group() without a timeout falls back to torch's 1800s default, so --distributed-timeout-seconds never reached the attn/moe subgroups. K3's first boot on a device with no pre-swept MoE tactic table (tables ship for B300 ep8/tp8 only) autotunes for ~45 min and the subgroup gloo recv died at exactly 1800000 ms, twice. Inherit the init timeout in every new_group call.

Validated on GB200: full 93-layer K3 at attn-tp16/ep16 across 4 nodes (READY in 695 s warm, coherent greedy, 4-point concurrency sweep with zero failed requests, c16 = 569 tok/s with prefill graphs disabled) and a 24-layer real-weight variant at tp8 across 2 nodes, both with flashinfer-native SiTU MoE (896-expert cubins), flashkda and forced-fla KDA prefill, fp8 MLA KV.


nperrin-fr:
requires #860 to serve k3

nperrin-fr:
GB200 launch command used for validation (full Kimi-K3, 4 nodes × 4 GPUs, attn-TP16/EP16 — one `ts serve` per node):

```bash
# per-node env (aarch64/Grace specifics)
export CUDA_HOME="$CONDA_PREFIX"
export CPATH="$CONDA_PREFIX/targets/sbsa-linux/include${CPATH:+:$CPATH}"  # aarch64 CUDA headers for JIT
export NCCL_SOCKET_IFNAME=enP6p9s0np0
export GLOO_SOCKET_IFNAME=enP6p9s0np0   # gloo does not read NCCL_SOCKET_IFNAME

ts serve \
  --model moonshotai/Kimi-K3 \
  --served-model-name kimi-k3 \
  --engine-startup-timeout 604800 \
  --distributed-timeout-seconds 7200 \
  --attn-tp-size 16 \
  --ep-size 16 \
  --enable-allreduce-fusion \
  --nnodes 4 \
  --nprocs-per-node 4 \
  --world-size 16 \
  --node-rank $NODE_RANK \
  --dist-init-addr $MASTER_IP:29520 \
  --max-model-len 65536 \
  --max-num-seqs 16 \
  --max-prefill-tokens 8192 \
  --chunked-prefill-size 8192 \
  --gpu-memory-utilization 0.94 \
  --disable-cuda-graph-padding \
  --trust-remote-code \
  --moe-backend flashinfer_trtllm \
  --kv-cache-dtype fp8 \
  --mm-encoder-tp-mode data \
  --disable-kvstore \
  --enable-cache-report \
  --host 0.0.0.0 --port 8000
```

Notes:
- no `--quantization` flag: the checkpoint's compressed-tensors (mxfp4) config is auto-detected; passing `mxfp4` explicitly conflicts.
- `--mm-encoder-tp-mode data` now works multi-node (vision TP1 item-DP16): the item-DP output gather was already NCCL, and this PR ships the multimodal shm features over the CPU group to ranks on other hosts. Validated with an image request across all 4 nodes (correct description, zero engine exceptions); text throughput is unchanged with the vision tower loaded (c16 ≈ 596 vs 569 tok/s text-only). Text-only serving can still use `--language-model-only`.
- `--distributed-timeout-seconds 7200`: first boot on a device with no pre-swept MoE tactic table autotunes ~45 min (tables ship for B300 only); needs the subgroup-timeout fix in this PR to actually apply.
- scheduler built with `SKBUILD_CMAKE_DEFINE="TOKENSPEED_FLAT_KVCACHE=ON"`; deps: flashinfer 0.6.16.dev20260730 (+ aarch64 jit-cache wheel), flash-linear-attention, and tokenspeed-cutedsl-kda>=0.1.0.post20260731 (first aarch64 build) or tokenspeed-flashkda — all three KDA prefill tiers (cutedsl > flashkda > fla) validated on GB200.

Results: READY 695 s warm / ~50 min cold sweep, GPQA-diamond **0.9293** (198 Q, temp 1.0, 32k max_tokens, reasoning_effort high), c16 = 596 tok/s (vision loaded) / 569 tok/s (text-only), 0 failed requests.

nperrin-fr:
@chenht2022 regarding the cross-node dp mm

chenht2022 (dismissed):
Verified on 2 B300 nodes (16 GPUs total) with Kimi-K3: the cross-node multimodal DP path and E2E image inference both work correctly. There is still redundant cross-node transfer/H2D overhead because features are sent to more DP/TP ranks than necessary. I’ll follow up with an optimization.

nperrin-fr:
working on fixing PR Test NVIDIA ARM / unit-test / ut-runtime-1gpu (gb200-1gpu) (pull_request)

nperrin-fr:
> https://github.com/lightseekorg/tokenspeed/actions/runs/30862177497 @nperrin-fr qwen 35 pd failed

It doesn't feel like this PR will fail PD. I am taking a look on why PD is flaky.

PR #877, opened 2026-07-31
feat(amd): add initial GFX1250 Gluon MLA kernels

Yu-Zhewen:
## Summary

Add BF16 GFX1250 Gluon MLA kernels for prefill, cached extend, and decode, with runtime routing between absorbed and expanded attention.

The new kernels are:

- `gluon_mla_prefill_bf16_gfx1250`
- `gluon_mla_extend_bf16_gfx1250`
- `gluon_mla_decode_bf16_gfx1250`

## Absorbed and expanded MLA

MLA stores a compressed KV latent instead of full K/V tensors for every attention head.

- **Absorbed attention** applies the key projection to Q and performs attention directly against the compressed KV latent. The compressed result is expanded through the value projection afterward. This avoids materializing full per-head K/V tensors and is efficient for decode and short cached extends.
- **Expanded attention** materializes full per-head K/V tensors from the compressed latent and then runs standard attention. This adds projection and memory cost but provides better parallelism for prefill and long query chunks.

## Difference from AITER

The absorbed GFX1250 implementation is adapted from [ROCm/AITER at commit `4a1cc773`](https://github.com/ROCm/aiter/commit/4a1cc773f34cbfc74387259e51262556ee38edd0).

For the unshuffled BF16 path, AITER uses absorbed attention for prefill, cached extend, and decode:

- `mla_prefill_fwd` uses `_mla_prefill_fwd_kernel_non_pipelined` for both absorbed prefill and cached extend.
- `mla_decode_fwd` uses a dedicated `_mla_decode_fwd_kernel_non_pipelined` split-K decode implementation.

This PR instead provides three dedicated TokenSpeed kernels:

| Operation | Kernel | MLA path |
|---|---|---|
| Prefill | `gluon_mla_prefill_bf16_gfx1250` | Expanded |
| Cached extend | `gluon_mla_extend_bf16_gfx1250` | Absorbed |
| Decode | `gluon_mla_decode_bf16_gfx1250` | Absorbed split-K |

For cached extend, the kernel registry selects the implementation by query length:

- `q ≤ 256`: use `gluon_mla_extend_bf16_gfx1250`.
- `q > 256`: expand K/V and replay the prefix with `gluon_mla_prefill_bf16_gfx1250`.

This keeps the compressed-cache advantage for decode and short extends while using the more efficient dense expanded kernel for prefill and long extends.

## Performance

Speedup is Triton latency divided by Gluon latency. Values above `1.0×` favor Gluon.

Measurements use BF16, 12 query heads, page size 64, KV LoRA rank 512, RoPE dimension 64, and value-head dimension 128.

Each result is the median round p50 from three alternating rounds with 20 warmups and 100 synchronized GPU-event samples per round.

### Prefill

| Shape | Gluon vs Triton |
|---|---:|
| B1 · 1K | 1.00× |
| B1 · 4K | 2.34× |
| B1 · 8K | 2.84× |
| B4 · 1K | 1.50× |
| B4 · 4K | 3.55× |
| B4 · 8K | 4.30× |

### Decode

| Shape | Gluon vs Triton |
|---|---:|
| B1 · 1K context | 2.12× |
| B1 · 4K context | 6.47× |
| B1 · 8K context | 12.13× |
| B4 · 1K context | 2.15× |
| B4 · 4K context | 6.53× |
| B4 · 8K context | 12.33× |

### Cached extend

These are route-equivalent GPU timings. They include the applicable projections, K/V materialization, cache write, attention calls, and final value projection or state merge.

The table reports the path selected by the kernel registry: absorbed Gluon for `q ≤ 256` and expanded Gluon for longer queries.

| Shape | Gluon vs Triton |
|---|---:|
| B1 · q128 / p1K | 2.28× |
| B1 · q256 / p1K | 2.22× |
| B1 · q128 / p4K | 2.11× |
| B1 · q256 / p4K | 2.08× |
| B4 · q128 / p1K | 1.07× |
| B4 · q256 / p1K | 1.13× |
| B4 · q128 / p4K | 1.43× |
| B4 · q256 / p4K | 1.04× |
| B1 · q1K / p1K | 0.93× |
| B1 · q4K / p4K | 1.33× |
| B1 · q8K / p8K | 1.83× |
| B4 · q1K / p1K | 1.11× |
| B4 · q4K / p4K | 1.64× |
| B4 · q8K / p8K | 1.94× |


python/tokenspeed/runtime/models/deepseek_v3.py
```diff
@@ -731,14 +731,27 @@ def _attn(
...
-                positions[:num_prefill_tokens],
-                q[:num_prefill_tokens],
-                latent_cache[:num_prefill_tokens],
-                prefill_ctx,
-                out_cache_loc[:num_prefill_tokens],
-                attn_output[:num_prefill_tokens],
-            )
+            # Use absorbed attention for cached-prefix extend when supported by
+            # the backend and profitable for this query shape; otherwise use
+            # normal chunked prefill.
```
borontion, line 736: are you sure we need this extend kernel here? will it bring any benefit compared vs the chunked version?

borontion: also if that's the case. I think this diverge should happen in mla backend.

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/attention/gluon/__init__.py
borontion, line 1: given we need to maintain kernels across 2 hardware architectures now. should we have re-org this structure a little bit? like `tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/attention/gluon/gfx1250|gfx950`

cc @antiagainst

python/tokenspeed/runtime/models/deepseek_v3.py
borontion: > also if that's the case. I think this diverge should happen in mla backend.

we can do this separately, given need to consider nvidia's mla backend

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/attention/gluon/__init__.py
antiagainst: Yeah I think that makes sense. @borontion maybe you can send a pull request to update all existing kernels to unify the directory structure and filename first.

python/tokenspeed/runtime/models/deepseek_v3.py
Yu-Zhewen: I benchmarked the "absorbed" version is 1.04–2.16× faster than the "chunked" version for q ≤ 256. In AITER, I'm not sure why, but it always uses the "absorbed" version unconditionally for both prefill and extend. I think it makes sense to choose the kernel based on the query length here. FYI, the decode path also uses the "absorbed" approach.

borontion (approved):

PR #878, opened 2026-07-31
perf: Optimize Kimi K3 MoE input projections and low-token decode

panditsa:
## Summary

  - Fuse Kimi K3’s router, common routed-latent, and shared-expert input projections into one dispatch.
  - Add optimized gfx950 decode paths for small token counts, using warp-GEMV through M=4 and grouped MFMA from M=5 onward.
  - Tune the M=4 routed W2/combine path while retaining portable fallbacks for unsupported shapes.

### Overall pipeline

```text
Before Fusion:
3 input projections -> top-k -> routed experts + separate shared expert
                    -> joint reduce -> routed up-projection -> add3

After Fusion:
packed input projection fusion kernel -> top-k -> routed W2/shared-down fusion kernel
                        -> joint reduce -> routed up-projection + add3 epilogue fusion kernel
```

  ## Tests

  - Focused GPU test suite: 40 passed
  - 4K/1K benchmark on 8× MI350X, TP8/EP8:
    - M=1: +11.03% throughput
    - M=2: +15.01%
    - M=4: +14.06%
    - M=8: +7.27%
    - M=16: +2.62%


python/tokenspeed/runtime/layers/moe/latent.py
```diff
@@ -326,6 +358,22 @@ def forward(
...
 
+        # Fusing the three input projections into one GEMM serializes the
+        # shared branch against the routed one by construction, so it is only
+        # worth taking when the branches were not going to overlap anyway.
+        fused = None
+        if self.input_projections is not None and not overlap_shared and num_tokens > 0:
+            fused = self.input_projections(hidden_states)
+        fused_router, fused_routed, fused_shared = (
+            (None, None, None) if fused is None else fused
+        )
```
raikonenfnu: Thoughts on having an IS_MEGAKERNEL_FORWARD flag  and propagate down to make it more obvious?

python/tokenspeed/runtime/layers/moe/latent.py
```diff
@@ -407,14 +455,18 @@ def run_shared_branch() -> None:
...
+            _check_shape(routed_output, output_shape, "routed_up_proj")
             return routed_output
+        if not self.defer_routed_up_projection:
+            routed_output = _module_tensor_output(self.routed_up_proj, routed_latent)
+            _check_shape(routed_output, output_shape, "routed_up_proj")
         if self.shared_reduce is not None and not shared_reduction_applied:
             shared_output = self.shared_reduce(shared_output)
             _check_shape(shared_output, output_shape, "shared_reduce")
+        if self.defer_routed_up_projection:
+            return routed_latent, shared_output
```
raikonenfnu: Is this with megakernel/fused?

python/tokenspeed/runtime/layers/moe/latent.py
```diff
@@ -407,14 +455,18 @@ def run_shared_branch() -> None:
...
-        routed_output = _module_tensor_output(self.routed_up_proj, routed_latent)
-        _check_shape(routed_output, output_shape, "routed_up_proj")
-
         if shared_output is None:
+            routed_output = _module_tensor_output(self.routed_up_proj, routed_latent)
+            _check_shape(routed_output, output_shape, "routed_up_proj")
             return routed_output
+        if not self.defer_routed_up_projection:
+            routed_output = _module_tensor_output(self.routed_up_proj, routed_latent)
+            _check_shape(routed_output, output_shape, "routed_up_proj")
```
raikonenfnu, line 518: Thoughts on doing just `if shared_output is None or not self.defer_routed_up_projection:`?

python/tokenspeed/runtime/layers/moe/latent.py
```diff
@@ -337,7 +385,7 @@ def run_shared_branch() -> None:
             # both paths run serially on the primary stream.  The fork joins
             # before collectives, and this H-width result is added to the
             # routed result at the end of the layer.
-            shared_output = _module_tensor_output(self.shared_experts, hidden_states)
+            shared_output = project(fused_shared, self.shared_experts)
```
raikonenfnu: Thoughts on making it `shared_output = fused_shared or _module_tensor_output(self.shared_experts, hidden_states)` or adding a new arg to _`module_tensor_output` to take in fused values? Feels like having `project` on top of `_module_tensor_output` is quite a bit of abstraction.

python/tokenspeed/runtime/models/kimi_k3.py
```diff
@@ -1100,10 +1106,87 @@ def __init__(
...
         )
+        self._use_joint_decode = (
+            self.native_latent_moe is not None
+            and bool(
+                getattr(
+                    self.shared_experts,
+                    "_has_unquantized_shared_weights",
+                    False,
+                )
+            )
```
raikonenfnu: Remind me, why did we need to check for this?

python/tokenspeed/runtime/models/kimi_k3.py
```diff
@@ -1166,12 +1301,18 @@ def forward(
...
                 hidden_states,
                 num_global_tokens=num_global_tokens,
                 max_num_tokens_per_gpu=max_num_tokens_per_gpu,
             )
-            return add3(prefix_sum, routed_out, shared_out)
+            return self.routed_expert_up_proj.forward_add3(
+                routed_latent,
+                prefix_sum,
+                shared_out,
+            )
```
raikonenfnu, line 1309: Looks like forward_add3 internally calls `prefix_sum + routed_latent @ self.weight + shared_out`. But it used to be just `add3`. Which implies that we used to generate routed out use in a separate kernel as part of `native_latent_moe.forward`? 

Looks like because it was in latent.py but we are dropping `routed_output = _module_tensor_output(self.routed_up_proj, routed_latent)` for the "fused" case. 

Thoughts on moving this into part of latent_moe.forward as well? Just to have a centralized location where these mega/fused kernel is done?

python/tokenspeed/runtime/models/kimi_k3.py
```diff
@@ -1166,12 +1301,18 @@ def forward(
         the up-projection's store performs the accumulate in-kernel.
         """
         if self.native_latent_moe is not None:
-            routed_out, shared_out = self.native_latent_moe(
+            if self._use_joint_decode and hidden_states.shape[0] == 1:
+                return self._forward_native_joint_decode(hidden_states, prefix_sum)
```
raikonenfnu: Maybe we should come up with/agree on mega/fused kernel name decorum btw? 
CC: @qedawkins

python/tokenspeed/runtime/models/kimi_k3.py
```diff
@@ -1100,10 +1106,87 @@ def __init__(
...
+            )
+            and latent_moe_decode_pipeline_available(
+                self.gate.weight,
+                self.routed_expert_down_proj.weight,
+                self.shared_experts.gate_up_proj.weight,
+                self.shared_experts.down_proj.weight,
+                self.experts.plan,
+                joint_reduce=self.execution_plan.joint_moe_reduce,
+            )
+        )
```
raikonenfnu: Feels like quite verbose, do we need all these checks?

python/tokenspeed/runtime/models/kimi_k3.py
```diff
@@ -1153,6 +1236,58 @@ def _routed_experts(
                 out = all_reduce(out, self.mapping.moe.tp_group)
         return out
 
+    def _forward_native_joint_decode(
```
raikonenfnu: Maybe can name it something related to megakernel, maybe megakernel_forwrd_decode or something? That way in the future we can potentially fuse more stuff inside.

python/tokenspeed/runtime/models/kimi_k3.py
```diff
@@ -1153,6 +1236,58 @@ def _routed_experts(
...
+        self,
+        hidden_states: torch.Tensor,
+        prefix_sum: torch.Tensor,
+    ) -> torch.Tensor:
+        """Fuse K3 decode input projections and routed/shared expert execution."""
```
raikonenfnu: Thoughts on listing out the different components here, and which components are now fused?

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/latent_input.py
raikonenfnu, line 1: I think we'd need to rebase anyways, but would be good to have this file located somewhere more obvious/intuitive on what it does for MoE.

python/tokenspeed/runtime/layers/moe/latent.py
panditsa: Done, removed this unnecessary abstraction

python/tokenspeed/runtime/models/kimi_k3.py
panditsa: This is unecessary, it's only checking if shared_experts have a valid `.weight`

python/tokenspeed/runtime/models/kimi_k3.py
panditsa: Reduced to minimum numbers of checks needed

python/tokenspeed/runtime/layers/moe/latent.py
panditsa: This was originally tied to the specialized path for M = 1, but the defer flag has since been removed.

python/tokenspeed/runtime/models/kimi_k3.py
panditsa: I renamed it to ` _forward_fused_decode_pipeline` because the method coordinates multiple launches rather than calling one megakernel. But I am open to renaming these once we confirm on the naming convention.

test/runtime/layers/test_latent_moe.py
```diff
@@ -38,6 +38,21 @@ def forward(self, hidden_states: torch.Tensor):
...
+    def __init__(self, events: list[str]) -> None:
+        super().__init__(events, "up")
+
+    def forward_add3(
+        self,
+        routed_latent: torch.Tensor,
+        prefix_sum: torch.Tensor,
+        shared_output: torch.Tensor,
+    ) -> torch.Tensor:
+        self.events.append("up_add3")
```
raikonenfnu: Fancy, what does events do haha

python/tokenspeed/runtime/layers/moe/latent.py
```diff
@@ -234,6 +239,20 @@ def forward(self, hidden_states: torch.Tensor):
         )
         return output, None
 
+    def forward_add3(
+        self,
+        hidden_states: torch.Tensor,
+        addend_a: torch.Tensor,
+        addend_c: torch.Tensor,
+    ) -> torch.Tensor:
+        """Project routed latents and accumulate two full-width addends."""
```
raikonenfnu: Can we also add  an additional one line here here  `result = addend_a + hidden_states @ self.weight + addend_c` here.

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/gluon/latent_decode.py
raikonenfnu, line 1: Should we move the kernels to tokenspeed-kernels-amd within the moe directory?

test/runtime/layers/test_latent_moe.py
panditsa: This is to trace which kernels are executed and in what order. I don't think it's actually that helpful besides debug. I'll remove this mechanism altogether, given that traces we collect provide the same info.

test/runtime/test_trtllm_cache_groups.py
```diff
@@ -45,10 +45,10 @@ def _bare_backend(
         device="cpu",
         groups=None,
     ):
-        from tokenspeed.runtime.layers.attention.kv_cache.recipes.spec import (
-    PagedCacheGroupSpec,
-)
         from tokenspeed.runtime.layers.attention.kv_cache.base import CachePool
+        from tokenspeed.runtime.layers.attention.kv_cache.recipes.spec import (
+            PagedCacheGroupSpec,
+        )
```
raikonenfnu, line 51: Did we fix some upstream/main lint issue? :o

tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/kimi3.py
```diff
@@ -452,7 +453,12 @@ def kimi3_latent_projection_add3(
         and (k, n) in _KIMI3_SHAPES
     )
     if solution == "auto":
-        solution = "rowcta_gemv" if m == 1 and specialized else "composed"
+        if m == 1 and specialized:
+            solution = "rowcta_gemv"
+        elif Platform.get().is_cdna4 and m == 16 and specialized:
+            solution = "gluon_mfma_add3"
+        else:
+            solution = "composed"
```
raikonenfnu, line 461: No action required for now since pre-existing strucutre is already like this. 

But would be good to eventually refactor to have this properly use the kernel selector. :)

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/gluon/__init__.py
```diff
@@ -20,3 +20,8 @@
 
 import tokenspeed_kernel.ops.moe.gluon.bf16  # noqa: F401
 import tokenspeed_kernel.ops.moe.gluon.mxfp4  # noqa: F401
+from tokenspeed_kernel.platform import current_platform
+
+if current_platform().is_amd:
+    import tokenspeed_kernel_amd.ops.gfx950.moe.latent_decode  # noqa: F401
+    import tokenspeed_kernel_amd.ops.gfx950.moe.latent_input  # noqa: F401
```
raikonenfnu: QQ, how come other gluon kernels/functions do not need this but these guys do?

raikonenfnu (commented):
Overall looks good to me, just last few questions :)

test/runtime/test_trtllm_cache_groups.py
panditsa: yeah, it kept getting flagged for me, and it was only a lint fix so I just applied it.

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/gluon/__init__.py
panditsa: aah good catch, when moving the latent projection kernels from tokenspeed-kernels to tokenspeed-kernels-amd this also moved the registration causing this. Fixed.

tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/kimi3.py
panditsa: https://github.com/panditsa/tokenspeed/issues/1  opened an issue locally. I can address this as part of cleanup.

raikonenfnu (approved):
LGTM!

PR #889, opened 2026-08-01
kernel: remove gfx942 support for now

antiagainst:
We don't have immediate plans to support gfx942
and there are no CI machines for it. So it's effective
dead code; dropping for now.


borontion:
can we also drop this `PlatformInfo.fp8e4m3fn`? I think it is cdna3-specific as well. It should just be `torch.float8_e4m3fn`.

tokenspeed-kernel/test/test_platform.py
raikonenfnu, line 1: Do we need this test?

raikonenfnu (dismissed):
LGTM! just one quick question.

tokenspeed-kernel/test/test_platform.py
antiagainst: Yeah added by agent; its value is questionable. I'll remove it.

PR #891, opened 2026-08-01
refactor(runtime): make ForwardContext tensor-free 

rjzhb:
## Goal

`ForwardContext` documents itself as:

> Do not contain Tensor

However, five loose `torch.Tensor` fields had accumulated over time:

* `req_to_page`
* `gather_ids`
* `draft_seq_lens_buf`
* `accept_lengths`
* `dsa_swa_slot_mapping`

This PR makes `ForwardContext` tensor-free again. Each tensor now travels through one of the following mechanisms:

* An explicit forward argument
* A persistent buffer
* Attention metadata

Nothing loose is carried on the per-forward context anymore.

This is **not a bug fix**. These fields were either only read on eager paths or already backed by persistent buffers, so CUDA graph replay is unaffected. This is an invariant and code-hygiene cleanup intended to prevent `ForwardContext` from silently becoming a tensor grab bag.

## Status

All five tensor fields have been removed:

* [x] `draft_seq_lens_buf`

  * Removed.
  * Draft cache lengths now reuse the drafter's `seq_lens` buffer, which is passed explicitly.

* [x] `accept_lengths`

  * Passed explicitly as a forward keyword argument.
  * Only used during the drafter's first step.

* [x] `req_to_page`

  * Stored as a persistent page-table buffer.
  * All reads now receive it through explicit arguments.

* [x] `gather_ids`

  * Computed per forward in the executor as `_target_gather_ids`.
  * Passed explicitly to both the model and logits processor.

* [x] `dsa_swa_slot_mapping`

  * Folded into the DeepSeek V4 attention metadata.
  * Set and cleared within a single forward pass.

`ForwardContext` now carries no `torch.Tensor` fields.


rjzhb:
image
all ci passed

PR #914, opened 2026-08-03
refactor: rewrite triton moe kernel

borontion:
The goal is to provide a simple portable MoE triton implementation with minimal dependency, not for peak performance. The current PR rewrote the original triton MoE using a persistent grouped gemm style kernel. And it only provides support for on BF16/MXFP4 weights for target models, drop the support for FP8 weights.

PR #915, opened 2026-08-03
[Do not merge] feat(moe): MORI all-to-all expert-parallel MoE backend (AMD)

nirvedhmeshram:
## Summary (human)
PR introduces a MORI all2all backend, that exposes an API to dispatch tokens to ranks and then combine them after. To show this working we use it in Kimi-k2.5-mxfp4 where we have a `mori_ep_mxfp4_moe_apply` kernel that does dispatch+ grouped_gemm + combine . 

## Summary (AI)
Add an AMD-native expert-parallel MoE path selected by --enable-expert-parallel --all2all-backend mori, backed by MORI's v2 dispatch/combine op (github.com/ROCm/mori).

- ops/communication/mori_ep.py: MORI dispatch/combine adapter (shape-keyed, process-wide dispatcher; real all-to-all vs the masked-replicate `none` path).
- ops/moe/mori/mxfp4.py, ops/moe/mori/bf16.py: registered "moe"/"apply" kernels that run the per-expert grouped-GEMM over MORI's [E_local, cap, H] buffer -- the native MXFP4 path (no dequant) and a bf16 path -- selected via traits/capability.
- The MXFP4 kernel is HIP-graph capturable (sync-free gather + a capture-aware static row bound), so the MoE forward runs under CUDA/HIP graphs without --enforce-eager.
- deepseek_v3: consume the complete routed result via forward_alltoall for dp>1 (RSAG), and via the standard MoE path with /tp_ep_size compensation for dp=1 full-EP; run the idle forward inside inference_mode so DP ranks don't crash mid-collective and deadlock.
- moe/utils: add All2AllBackend.MORI and is_all_to_all().

## Test Plan

Tests: single-GPU kernel unit tests, a 2-GPU real all-to-all distributed test, backend-selection tests, and a Kimi-K2.5-MXFP4 aime25 accuracy eval through the MORI EP path with attention DP. The MORI CI jobs install the pre-built amd-mori wheel and are optional until the ROCm CI image ships MORI + flydsl.


tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/mori/mxfp4.py
```diff
@@ -0,0 +1,188 @@
...
+        ep_size = int(getattr(w, "ep_size", dist.get_world_size()))
+        ep_rank = int(getattr(w, "ep_rank", dist.get_rank()))
+        num_local = int(getattr(w, "num_local_experts"))
+        # Fixed MORI dispatch capacity (tokens/rank). MORI allocates symmetric buffers
+        # ~[num_local, ws*cap, H] at op-creation, so cap must upper-bound tokens/rank but
+        # stay memory-bounded. A prefill of more than cap tokens/rank overflows the buffer,
+        # so raise cap (or chunk prefill to <= cap) for prompts longer than the default.
+        import os as _os
+
+        cap = int(_os.environ.get("MORI_EP_MAX_TOKENS_PER_RANK", "2048"))
```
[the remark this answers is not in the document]
nirvedhmeshram: fixed

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/mori/mxfp4.py
```diff
@@ -0,0 +1,188 @@
...
+        plan: dict,
+        x: torch.Tensor,
+        w: torch.nn.Module,
+        router_logits: torch.Tensor,
+        topk_weights: torch.Tensor | None = None,
+        topk_ids: torch.Tensor | None = None,
+        num_tokens_global: int | None = None,
+        max_num_tokens_per_gpu: int | None = None,
+        do_finalize: bool = True,
+        enable_pdl: bool = False,
```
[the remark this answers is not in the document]
nirvedhmeshram: fixed

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/mori/mxfp4.py
```diff
@@ -0,0 +1,204 @@
...
+        weight_preprocessor=triton_mxfp4_moe_weights,
+        capability=CapabilityRequirement(
+            vendors=frozenset({"amd"}),
+            min_arch_version=ArchVersion(9, 5),
+            max_arch_version=ArchVersion(9, 5),
+        ),
+        signatures=format_signatures("x", "dense", {torch.bfloat16}),
+        traits={
+            "weight_dtype": frozenset({"mxfp4"}),
+            "activation": frozenset({"silu", "swiglu"}),
```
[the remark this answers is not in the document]
nirvedhmeshram: fixed

tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/mori_ep.py
```diff
@@ -0,0 +1,186 @@
...
+    objs = [uid]
+    dist.broadcast_object_list(objs, src=0)
+    uid = objs[0]
+    if per_rank_vmm is None:
+        per_rank_vmm = int(
+            os.environ.get("MORI_PER_RANK_VMM", str(4 << 30))
+        )  # 4 GiB default
+    _COMM = Communicator.init(
+        dist.get_world_size(), rank, uid, per_rank_vmm=per_rank_vmm
+    )
```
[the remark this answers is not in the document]
nirvedhmeshram: fixed

tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/mori_ep.py
```diff
@@ -0,0 +1,186 @@
...
+) -> "MoriEpDispatcher":
+    """Return a process-wide MoriEpDispatcher for this EP shape, creating it once."""
+    key = (
+        rank,
+        world_size,
+        hidden_dim,
+        num_local_experts,
+        num_experts_per_token,
+        max_num_inp_token_per_rank,
+        str(data_type),
```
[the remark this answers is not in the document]
nirvedhmeshram: fixed

python/tokenspeed/runtime/models/deepseek_v3.py
```diff
@@ -387,13 +387,87 @@ def forward(
...
+            topk_output = self.topk(hidden_states, router_logits)
+        else:
+            topk_output = self.topk.empty_topk_output(
+                hidden_states.device,
+                hidden_states=hidden_states,
+                router_logits=router_logits,
+            )
+
+        # routed experts: MORI all-to-all on local tokens -> complete, no framework reduce
+        routed = self.experts(
```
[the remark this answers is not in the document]
nirvedhmeshram: fixed

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/__init__.py
```diff
@@ -235,7 +236,11 @@ def moe_plan(
     # DeepEP does not pin a solution: the ``supports_all_to_all_ep`` trait plus
     # ``weight_dtype`` already narrow the candidates to the apply kernels that
     # own the dispatch/combine legs (nvfp4 cutedsl, block-scale fp8 DeepGEMM).
-    # Callers may still force one explicitly through ``solution``.
+    # MORI pins ``solution="mori"``: both MORI apply kernels share that solution and
+    # are disambiguated by ``weight_dtype`` (mxfp4 vs bf16). Callers may still force
+    # a solution explicitly.
+    if solution is None and a2a_backend == "mori":
+        solution = "mori"
```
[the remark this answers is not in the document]
nirvedhmeshram: added a opt in for MORI to fix this.

tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/mori_ep.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+def mori_available() -> bool:
+    """True if MORI v2 EP ops (and flydsl) can be imported."""
+    global _MORI_AVAILABLE
+    if _MORI_AVAILABLE is None:
+        try:
+            import flydsl  # noqa: F401
+            from mori.cco import Communicator  # noqa: F401
+            from mori.ops.dispatch_combine_v2 import (  # noqa: F401
```
borontion, line 49: let's add mori and fyldsl dependency to tokenspeed-kernel/python/requirements/rocm-thirdparty.txt (instead of only installing it in ci) and enable it if platform.is_amd

test/ci/ut/ut-mori-ep-2gpu.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - amd-mi35x-2gpu-test
+env:
+  CI: "true"
+  MORI_PER_RANK_VMM: "4294967296"
+install:
+  - bash test/ci_system/install_deps.sh
+  # MORI's .so links libgrpc++.so.1.51 (gRPC 1.51 C++ runtime), absent from the ROCm runner
+  # image; without it ``import mori`` fails and the test would skip via mori_available().
+  # Install the Ubuntu 24.04 (noble) t64 package so the test actually runs.
+  - sudo apt-get install -y libgrpc++1.51t64
```
borontion: can we add this to bash test/ci_system/install_deps.sh?

test/ci/eval/kimi-k2.5-mxfp4-aime25-mori-ep-amd.yaml
```diff
@@ -0,0 +1,82 @@
...
+  # MORI symmetric-buffer VMM per rank, and disable the flydsl disk cache: all EP ranks
+  # (world-size 4) compile the v2 device kernels concurrently and share ~/.flydsl/cache,
+  # which races into a corrupt binary on first use -- compile in-memory instead.
+  MORI_PER_RANK_VMM: "8589934592"
+  FLYDSL_RUNTIME_ENABLE_CACHE: "0"
+  # Per-rank MORI dispatch capacity (tokens/rank). Must upper-bound tokens/rank per step;
+  # --chunked-prefill-size below is set to the same value so prefill (and the autotune
+  # dummy prefill, which is chunked_prefill_size tokens) never exceeds the cap. Larger
+  # prefills need a bigger cap + VMM; a decode-cap-independent gather is a follow-on.
+  MORI_EP_MAX_TOKENS_PER_RANK: "2048"
```
borontion: can we add mori env as server args python/tokenspeed/runtime/utils/server_args.py like --mori-XXX 

also is this `FLYDSL_RUNTIME_ENABLE_CACHE` required?

borontion:
thanks @nirvedhmeshram. do you have any perf number so far compared to w/o using all2all?

nirvedhmeshram:
@borontion perf update — I ran an A/B on a single MI350X node (dp=4, EP=4, 4 GPUs), identical server config on both sides, only the backend toggled. `ts bench serve`, random 1024→256, 256 prompts, max-concurrency 64:

| Metric | `--all2all-backend mori` | `--all2all-backend none` (all-gather EP) |
|---|---|---|
| Output token throughput | 329 tok/s | **413 tok/s** |
| Total token throughput | 1647 tok/s | **2066 tok/s** |
| Mean TPOT | 131 ms | **118 ms** |
| Mean TTFT | 9985 ms | **6147 ms** |

So at this scale MORI a2a is currently **~20% slower** than the plain all-gather EP path — I want to be upfront about that rather than cherry-pick.

Cause: the MORI grouped-GEMM currently works over the full `[num_local_experts, cap, H]` capacity buffer every step, and `cap` has to cover the prefill chunk (2048 here), so decode (~16 tokens/rank) pays the full 2048-cap gather. The all-gather baseline only touches real tokens. The a2a win is expected at larger EP scale (more ranks/nodes, where all-gather bandwidth dominates), and a decode-cap-independent / O(#recv-tokens) gather is the planned follow-up to close this at small scale.

Correctness/accuracy are unaffected: aime25 `mean_acc = 0.75` (dp=4, GRAPH mode) on the current branch, matching the reference path.

test/ci/eval/kimi-k2.5-mxfp4-aime25-mori-ep-amd.yaml
nirvedhmeshram: moved to server arguments, about the cache yes its currently picking corrupted state so its necessary, however, I have moved that logic inline in the MORI EP setup.

tokenspeed-kernel/python/requirements/rocm-thirdparty.txt
```diff
@@ -1,2 +1,8 @@
 tokenspeed-triton-kernels==1.0.0.post20260510
 tokenspeed-iris==0.0.0.post20260511
+# AMD-native MORI all-to-all EP dispatch/combine backend (--all2all-backend mori). amd-mori
+# 1.2.2 requires the pre-0.3.0 flydsl API (0.3.0 renamed flydsl.expr.vector -> Vector), so pin
+# flydsl explicitly instead of amd-mori's unpinned [flydsl] extra. Usage is opt-in at runtime
+# and self-skips if the import fails (see ops/communication/mori_ep.py:mori_available).
+amd-mori==1.2.2
+flydsl==0.2.4
```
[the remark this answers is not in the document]
nirvedhmeshram: This conflicts with what @borontion wants so we are going to keep the this is requirements.

borontion (dismissed):
@qedawkins can you also take a look? thanks

python/tokenspeed/runtime/layers/moe/utils.py
```diff
@@ -48,6 +72,10 @@ class All2AllBackend(Enum):
     NONE = "none"
     DEEPEP = "deepep"
     FLASHINFER_NVLINK_ONE_SIDED = "flashinfer_nvlink_one_sided"
+    # AMD-native EP dispatch/combine via MORI (github.com/ROCm/mori). IntraNode
+    # (XGMI) + InterNode (RDMA) kernels; real all-to-all, unlike the masked-replicate
+    # `none` fallback.
+    MORI = "mori"
```
efric: are we introducing support for internode with this?

nirvedhmeshram: Yes because all the host setup is happening behind the MORI API, since MORI is already inter-node, this all2all implementation would work without any changes for inter-node.

qedawkins:
> Cause: the MORI grouped-GEMM currently works over the full `[num_local_experts, cap, H]` capacity buffer every step, and `cap` has to cover the prefill chunk (2048 here), so decode (~16 tokens/rank) pays the full 2048-cap gather. The all-gather baseline only touches real tokens. The a2a win is expected at larger EP scale (more ranks/nodes, where all-gather bandwidth dominates), and a decode-cap-independent / O(#recv-tokens) gather is the planned follow-up to close this at small scale.


If that is the hypothesis can we at least try ws=8 or just focus on larger prefill cases to demonstrate benefit? Or if just the internode support is the goal that's fine too.

nirvedhmeshram:
> > Cause: the MORI grouped-GEMM currently works over the full `[num_local_experts, cap, H]` capacity buffer every step, and `cap` has to cover the prefill chunk (2048 here), so decode (~16 tokens/rank) pays the full 2048-cap gather. The all-gather baseline only touches real tokens. The a2a win is expected at larger EP scale (more ranks/nodes, where all-gather bandwidth dominates), and a decode-cap-independent / O(#recv-tokens) gather is the planned follow-up to close this at small scale.
> 
> If that is the hypothesis can we at least try ws=8 or just focus on larger prefill cases to demonstrate benefit? Or if just the internode support is the goal that's fine too.

The main goal is to be feature complete in case a model needs this setup in a certain serving setup, its hard to say if Kimi2.5 is one of those models without optimizing, but I can spend some time to see which use case if any provides an advantage. I want to defer working on optimization in follow-up work though.

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/mori/bf16.py
```diff
@@ -0,0 +1,140 @@
...
+    E, cap, H = packed_x.shape
+    I = w13.shape[1] // 2
+    # zero padding rows (>= counts[e]) so padding never produces NaN/garbage that could leak
+    valid = torch.arange(cap, device=packed_x.device)[None, :] < counts[:, None].to(
+        torch.long
+    )
+    xin = (packed_x * valid[..., None]).to(torch.bfloat16)  # [E, cap, H]
+    gate_up = torch.bmm(xin, w13.transpose(1, 2))  # [E, cap, 2I]
+    inter = torch.nn.functional.silu(gate_up[..., :I]) * gate_up[..., I:]  # [E, cap, I]
+    out = torch.bmm(inter.to(torch.bfloat16), w2.transpose(1, 2))  # [E, cap, H]
```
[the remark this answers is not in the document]
nirvedhmeshram: yes this will be optimized in follow-up work. We are not using bf16 currently in load bearing way.

python/tokenspeed/runtime/models/deepseek_v3.py
```diff
@@ -387,13 +391,88 @@ def forward(
...
+            # shared expert is tp_ep-sharded (partial) and is completed by the same all_reduce,
+            # so it is NOT divided. (dp>1 uses forward_alltoall, no framework MoE reduce.)
+            # MORI only -- DeepEP keeps its upstream handling.
+            if (
+                get_all2all_backend().is_mori()
+                and self.mapping.attn.tp_size == self.mapping.moe.tp_ep_size
+                and self.mapping.moe.tp_ep_size > 1
+            ):
+                routed_expert_output = (
+                    routed_expert_output / self.mapping.moe.tp_ep_size
```
qedawkins, line 416: I'm a little confused here. Why are we still doing the all-reduce in this case? If we're returning the complete routed result shouldn't we skip the reduce here? Unless I'm misunderstanding this scaling step.

python/tokenspeed/runtime/layers/moe/utils.py
efric: AFAIK the vanilla `dispatch_combine_v2` API only supports intra node according to the documentation https://github.com/ROCm/mori/tree/main/python/mori/ops/dispatch_combine_v2 https://github.com/ROCm/mori/blob/main/python/mori/ops/dispatch_combine_v2/dispatch_combine_op.py, but correct me if it's outdated or I'm looking at the wrong place. If there is support can we test it out to see if it's stable?

python/tokenspeed/runtime/models/deepseek_v3.py
nirvedhmeshram: Yeah I was also wondering why we should do a reduce here but the reason we need it is in the dp=1 full-EP path because the shared expert is still RowParallel (partial per rank) and rides the same all-reduce — which we do need to complete the shared expert. So we divide the routed output by tp_ep_size (MORI's combine already returns the complete routed result on every rank unlike the non all2all path) so that all-reduce leaves it unchanged while still summing the shared expert to completion.

nirvedhmeshram: Note that dp=1 with all2all doesnt make much sense to begin with (since there is no need to route tokens anywhere) so no need to try optimizing this but just for completeness we still want it supported.

python/tokenspeed/runtime/layers/moe/utils.py
nirvedhmeshram: good catch, I was blindly trusting it to be inter-node but it isnt, so I have removed that claim from the comment.

python/tokenspeed/runtime/models/deepseek_v3.py
qedawkins: Ack it's not an expected path, but it is still surprising. Wouldn't we just skip reducing the routed outputs in this case and restrict the reduce to the shared expert?

qedawkins (dismissed):
LGTM otherwise

python/tokenspeed/runtime/models/deepseek_v3.py
nirvedhmeshram: So the reason to do the reduction for the routed experts is that if we don't do that then we wont have fusion, and the reduction is over the same bytes, the key point is we are doing the summation of ALL experts and then do the reduction over the summed tensor. here is a table of tradeoffs

| | divide-and-ride (current) | reduce-shared-only |
|---|---|---|
| routed + shared add | **before** the reduce (summed into one tensor, then reduced) | **after** the reduce (reduce `shared` alone, then add un-reduced `routed`) |
| extra elementwise divide | **yes** (routed / tp_ep_size) | no |
| all-reduce bytes | `[tokens, H]` | `[tokens, H]` (same) |
| separate collective for shared | no | yes |
| allreduce + RMSNorm fusion | **kept** | lost |

nirvedhmeshram:
Just want to say this in case some automation tries to merge this PR, I dont want this merged yet. I am noticing a further regression in the concurrency performance after some of the recent rebases I am still investigating.

tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/mori_ep.py
```diff
@@ -0,0 +1,238 @@
...
+    objs = [uid]
+    dist.broadcast_object_list(objs, src=0)
+    uid = objs[0]
+    if per_rank_vmm is None:
+        per_rank_vmm = int(
+            os.environ.get("MORI_PER_RANK_VMM", str(4 << 30))
+        )  # 4 GiB default
+    _COMM = Communicator.init(
+        dist.get_world_size(), rank, uid, per_rank_vmm=per_rank_vmm
+    )
```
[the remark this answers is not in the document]
nirvedhmeshram: A user should not use mori in such setup when it clearly says its for intra-node, so no need to handle this case.

nirvedhmeshram:
I wanted to give some final performance information after fixing issues after rebase (as I moved to a gluon kernel) before landing this PR. I couldnt access 8 GPUs at once to give world size 8 numbers though. However, here there is at least one use case where this work is pulling a bit ahead.

## TP4 vs DP4+EP4 (MORI a2a) — Kimi-K2.5-MXFP4

Single MI350X node, 4 GPUs, `ts bench serve`, random **1024→256**, `--ignore-eos`,
identical server config; only the parallelism/backend differs.

> **Note:** the `--all2all-backend none` (all-gather EP) baseline used in the earlier
> A/B is no longer testable — the EP-capable Triton MXFP4 kernel it relied on
> (`grouped_mxfp4_ffn`) was removed when the Triton MoE kernel was rewritten in #914.
> So **TP4** is used as the reference point here instead.

| Concurrency | Output tok/s (TP4) | Output tok/s (DP4+EP4 MORI) | MORI vs TP4 | Mean TTFT ms (TP4 / DP4) | TPOT ms (TP4 mean / DP4 median) |
|---|---|---|---|---|---|
| 16  | 157.5 | **183.0** | **+16%** | 2425 / 2237 | 83.3 / 82.6 |
| 32  | **312.2** | 236.5 | −24% | 3012 / 3929 | 79.6 / 112.7 |
| 64  | **436.9** | 337.2 | −23% | 4141 / 11105 | 115.0 / 123.4 |
| 128 | 420.2 | — (not run) | — | 21223 / — | 114.3 / — |

0 failed requests on every completed point, both configs.

**Takeaway:** DP4+EP4 with the MORI all-to-all is **faster at low concurrency**
(conc-16, +16% output throughput) but **does not scale as well** as tensor-parallel
as concurrency rises — TP4 pulls ahead by ~23–24% at conc-32 and conc-64, with
markedly lower TTFT at conc-64. Consistent with the known decode-cap tax on the
MORI path (the O(#recv-tokens) gather is the planned follow-up); I will push on this approach further to see if we can improve the scaling in follow up work.

nirvedhmeshram:
Since the CI has passed before and this PR is touching things that are mostly self contained I will clear this PR for merging as github actions are down. Will monitor the post submit for any issues that are related.

nirvedhmeshram:
@lightseek-bot please merge this PR

python/tokenspeed/runtime/models/deepseek_v3.py
qedawkins: Sorry, took me a while to understand what you mean here. I understand what is happening now, makes sense.

qedawkins (approved):

nirvedhmeshram:
Update on the PR after some internal discussion, we dont want the maintenance burden of the kimi-k2.5 kernels, so I am going to make this work on kimi-k3 and show a performant use case before we land this and remove all the kimi-k2.5 stuff, putting it back to do not merge state for now.

nirvedhmeshram:
closing this PR for now, as for kimi-k3 the dp configurations are not supported and its non trivial to support them because it breaks stride requirements of latent_kv. Moreover none of the other popular inference engines are using dp for this model. We can revive this work if an all-to-all backend becomes important for some use case.

PR #930, opened 2026-08-04
refactor(cache): unify all KV pools on LCM addressing path

FC-Li:
Migrate remaining cache pools onto the LCM planning/addressing path, removing the legacy allocator and dual-track req_to_page handling.


nirvedhmeshram:
Hi, I think the CI unit tests are broken after this PR and as a result the model tests are not running either?

PR #936, opened 2026-08-04
fix(prefill cudagraph): gate disable graph padding for prefill cudagraph

nperrin-fr:
## Summary

We currently capture graphs at log scale, but disabling cuda graph padding affected only the matching logic, resulting in prefill graph mismatch.

`--disable-cuda-graph-padding` should only take effect on decode graph.


nperrin-fr:
should be safe to merge.

PR #954, opened 2026-08-05
perf(kernel): fuse single-group MoE routing

Max191:
Use the dense Gluon route for single-group package MoE so sigmoid, bias, stable top-k selection, normalization, and scaling execute in one GPU launch for FP16, BF16, and FP32 router logits.

Encode each routing choice as a packed key containing ordered FP32 bits and the inverse expert index. Build the invariant key once before the unrolled top-k loop, apply the changing live mask per rank, and gather selected scores through integer bits to preserve deterministic exceptional-value behavior. Keep the defensive contiguous materialization at the Python API boundary, but skip it when callers already pass contiguous logits and bias tensors.

Add exact all-dtype routing, exceptional-value, graph-replay, public-dispatch, fused-input, strided-input, BF16 E=384, rounding-boundary, and CPU-device coverage.

Performance evidence:
- Compared ToM baseline with this change using three alternating Kimi 2.5 baseline/candidate pairs on 4 MI350 GPUs. Each run used attention TP4, MoE TP4, 8192 input and 8192 output tokens, and eight requests at concurrency eight.
- Median TTFT was 4504.873 -> 4550.146 ms (-1.005%), TPOT was 34.922 -> 25.135 ms (+28.028%), E2EL was 290788.499 -> 210427.049 ms (+27.636%), and output throughput was 221.458 -> 304.018 tok/s (+37.281%).
- Baseline/candidate ranges were 4295.430-4738.550/4522.990-4625.365 ms TTFT, 34.726-38.466/24.899-25.911 ms TPOT, 288732.526-319579.540/208577.163-216762.842 ms E2EL, and 201.872-223.042/294.840-306.691 tok/s output throughput.
- Every run completed 8/8 requests with exact generated text, the same 33016 response events, and the same 98.473467% speculative-work proxy. The all-dtype edge-semantics follow-up does not change Kimi's BF16 operation sequence.

Accuracy evidence:
- EvalScope 1.9.0 AIME25 scored 0.75 on four samples, meeting the 0.75 gate. The run used greedy decoding, an 8192-token output limit, attention TP4, MoE TP4, and the CI Eagle3 speculative-decoding configuration.


panditsa (approved):
LGTM!

PR #955, opened 2026-08-05
fix(eagle3): clear draft page table rows past the current batch

nperrin-fr:
The batch-ordered draft page table cleared columns past the table width but never rows past the current batch. A CUDA-graph replay reads padded_bs rows (cuda_graph_wrapper._init_replay_metadata hands the table straight to the drafter backend), so rows bs..padded_bs-1 kept real page ids left by an earlier, larger batch. The reserved req-pool sink row that shields pool-indexed tables from padding does not apply here: row i IS batch position i, so a padded row landed on another request's pages and EAGLE3's multi-step draft (steps >= 2 derives its write locations from this table) stored draft KV into them. The victim request then mispredicted permanently.

Observed as the MiniMax-M3 EAGLE3 CI regression: eval_accept_rate 0.665 -> 0.0015 (avg_accept_len 3.00 -> 1.005), score 0.90 -> 0.87, and the eval slowing from <20min to >30min. M3 is the only CI config on the vulnerable path: the K2.5 EAGLE3 evals run steps=1 (multi-step buffer is empty), the MTP drafters take write locations from the executor's out_cache_loc_buf instead of this table, and inkling/deepseek-v4 mask padded rows via uses_padded_decode_token_mask.

Verified on the CI-faithful local AIME25 gate: accept_len 3.040 (accept_rate 0.680, score 0.9, 1452 samples, min 2.370) against 1.005 broken / 3.00 healthy; at concurrency 8 four rounds hold 3.22-3.24 with 0 of 50 samples latched at 1.000 (was ~1.27 with a latched tail).


nperrin-fr:
should be safe to merge and spare some CI resource

Issue #845, opened 2026-07-29
K3 road to 70tps (on mi355)


raikonenfnu:
## Goal

Reach and sustain **70 output tokens/s** for Kimi K3 on an 8x gfx950/MI355-class TP8/EP8 deployment while preserving accuracy, portable fallbacks, and the `tokenspeed-kernel` dependency boundary.

The currently benchmarked stack reaches **67.105 output tok/s**, up from **41.852 tok/s on `main` (+60.34%)**. The remaining gap to the 70 tok/s output-throughput target is approximately **4.3%**. Its mean per-request decode rate is already 69.617 tok/s, so this issue deliberately tracks the stricter aggregate output-throughput metric.

Detailed experimental tracker: [raikonenfnu/tokenspeed#39](https://github.com/raikonenfnu/tokenspeed/issues/39).

## Sequential measured results

Each row is a cumulative, independently runnable branch in the `raikonenfnu/tokenspeed` fork.

| Split | Optimization introduced | Branch | Output throughput | Δ previous | Δ main | TPOT |
|---:|---|---|---:|---:|---:|---:|
| 0 | Current upstream baseline | `9193a3d` | 41.852 tok/s | — | — | 23.367 ms |
| 1 | Semantic MLA decode fusion | [`raikonen/k3-01-attention-ops`](https://github.com/raikonenfnu/tokenspeed/tree/raikonen/k3-01-attention-ops) | 44.007 tok/s | **+5.15%** | +5.15% | 22.181 ms |
| 2 | KDA norm/gate/output fusion | [`raikonen/k3-02-moe-ops`](https://github.com/raikonenfnu/tokenspeed/tree/raikonen/k3-02-moe-ops) | 43.994 tok/s | -0.03% | +5.12% | 22.159 ms |
| 3 | MoE input and expert decode fusion | [`raikonen/k3-03-gemm-fusions`](https://github.com/raikonenfnu/tokenspeed/tree/raikonen/k3-03-gemm-fusions) | 48.761 tok/s | **+10.83%** | +16.51% | 19.958 ms |
| 4 | MoE norm/up and collective staging | [`raikonen/k3-04-collective-lanes`](https://github.com/raikonenfnu/tokenspeed/tree/raikonen/k3-04-collective-lanes) | 52.318 tok/s | **+7.30%** | +25.01% | 18.574 ms |
[...]

PR #959, opened 2026-08-06
perf(kimi3): fuse AttnRes projection and collectives

qedawkins:
Summary:

Fuse single-token Kimi-K3 AttnRes work into its surrounding decode operations. Add a gfx950 Gluon projection that also computes two sets of AttnRes partials, and an Iris TP8 collective that all-reduces the attention projection, accumulates the residual, and completes the AttnRes mix and output RMSNorm in one launch. Wire the MLA and KDA paths through capability-gated APIs with existing fallbacks, and keep attn_res_fwd as the sole public AttnRes mixing API.

Testing Plan:

- Compare portable and gfx950 fused projection/partial results with their unfused composition, including fallback and strided-input coverage.
- Compare the TP8 Iris collective with separate all-reduce, residual, AttnRes, and RMSNorm operations across repeated launches and graph replay.
- Cover deterministic, cross-node, unsupported-shape, and ungated-MLA fallback dispatch.
- Verify public AttnRes dispatch with distinct scorer and output RMSNorm epsilons.

### DIFF
--- a/python/tokenspeed/runtime/distributed/comm_backend/base.py
+++ b/python/tokenspeed/runtime/distributed/comm_backend/base.py
@@ -78,32 +78,6 @@ class CommBackend(ABC):
             raise ValueError("all-reduce requires at least one output")
         return tuple(like.new_empty(shape) for shape in shapes)
 
-    def all_reduce_residual_attnres(
-        self,
-        partial: torch.Tensor,
-        residual: torch.Tensor,
-        score_weight: torch.Tensor,
-        output_weight: torch.Tensor,
-        scratch: tuple[torch.Tensor, torch.Tensor, torch.Tensor],
-        eps: float,
-        group: Group,
-        op=None,
-    ) -> tuple[torch.Tensor, torch.Tensor]:
-        """Reduce attention output, accumulate its residual, and finish AttnRes."""
-        from tokenspeed_kernel.ops.activation.triton import attnres_combine
-
-        reduced = self.all_reduce(partial, group, op=op)
-        residual_out = residual + reduced
-        hidden = attnres_combine(
-            residual_out,
-            score_weight,
-            output_weight,
-            eps,
-            scratch,
-            torch.empty_like(residual_out),
-        )
-        return hidden, residual_out
-
     @abstractmethod
     def all_gather(
         self, tensor: torch.Tensor, group: Group, dim: int = 0
--- a/python/tokenspeed/runtime/distributed/comm_backend/base.py
+++ b/python/tokenspeed/runtime/distributed/comm_backend/base.py
@@ -78,6 +78,32 @@ class CommBackend(ABC):
             raise ValueError("all-reduce requires at least one output")
         return tuple(like.new_empty(shape) for shape in shapes)
 
+    def all_reduce_residual_attnres(
+        self,
+        partial: torch.Tensor,
+        residual: torch.Tensor,
+        score_weight: torch.Tensor,
+        output_weight: torch.Tensor,
+        scratch: tuple[torch.Tensor, torch.Tensor, torch.Tensor],
+        eps: float,
+        group: Group,
+        op=None,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        """Reduce attention output, accumulate its residual, and finish AttnRes."""
+        from tokenspeed_kernel.ops.activation.triton import attnres_combine
+
+        reduced = self.all_reduce(partial, group, op=op)
+        residual_out = residual + reduced
+        hidden = attnres_combine(
+            residual_out,
+            score_weight,
+            output_weight,
+            eps,
+            scratch,
+            torch.empty_like(residual_out),
+        )
+        return hidden, residual_out
+
     @abstractmethod
     def all_gather(
         self, tensor: torch.Tensor, group: Group, dim: int = 0
--- a/python/tokenspeed/runtime/models/kimi_k3.py
+++ b/python/tokenspeed/runtime/models/kimi_k3.py
@@ -1769,6 +1769,8 @@ class KimiLinearDecoderLayer(nn.Module):
         attnres_partial_args = None
         if next_mix is not None and self._hoist_next_mlp and num_tokens == 1:
             next_layer, _ = next_mix
+            # This layer's attention projection writes both block partials
+            # hoisted for the next layer, which consumes their scratch slots.
             candidate_args = (
                 block_residual[:mlp_valid_blocks],
                 next_layer._mlp_wp,
@@ -1793,8 +1795,7 @@ class KimiLinearDecoderLayer(nn.Module):
             )
         )
         with self.attn_fork.scope(
-            enable=get_is_capture_mode()
-            and (attnres_partial_args is None or own_mlp)
+            enable=get_is_capture_mode() and (attnres_partial_args is None or own_mlp)
         ) as fork:
             with fork.branch():
                 if next_mix is not None:
@@ -1806,9 +1807,7 @@ class KimiLinearDecoderLayer(nn.Module):
                                 next_layer._mlp_wp,
                                 next_layer._attn_wp,
                                 self.mlp_res_norm.variance_epsilon,
-                                _sliced_scratch(
-                                    h, next_layer._mlp_slot, num_tokens
-                                ),
+                                _sliced_scratch(h, next_layer._mlp_slot, num_tokens),
                                 sc1,
                             )
                     else:
--- a/python/tokenspeed/runtime/models/kimi_k3.py
+++ b/python/tokenspeed/runtime/models/kimi_k3.py
@@ -98,7 +98,6 @@ from tokenspeed.runtime.configs.kimi_k3_config import KimiK3Config, KimiLinearCo
 from tokenspeed.runtime.distributed.comm_manager import CommManager
 from tokenspeed.runtime.distributed.comm_ops import (
     all_reduce,
-    all_reduce_residual_attnres,
     prepare_all_reduce_fusion,
     prepare_all_reduce_lane,
 )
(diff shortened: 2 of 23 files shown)


tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/linear_attnres_partials.py
```diff
@@ -0,0 +1,319 @@
...
+    )
+
+    if kernel is not None:
+        ShapeCapture.get().record(
+            "gemm",
+            "linear_attnres_partials",
+            kernel.name,
+            hidden_states.dtype,
+            traits,
+        )
```
raikonenfnu, line 272: Fancy! is this for future benchmarking or something?

qedawkins: The same thing happens in [mm](https://github.com/lightseekorg/tokenspeed/blob/b573644ef87da7e847a70abc91041f02a047e5a5/tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/__init__.py#L474-L490) and [bmm](https://github.com/lightseekorg/tokenspeed/blob/b573644ef87da7e847a70abc91041f02a047e5a5/tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/__init__.py#L623-L639). IIUC shape capture should be plumbed through all of the different kernel APIs to record the shapes and other metadata the kernel was setup with for replay and external tuning (I haven't tried any of that myself).

panditsa:
Would be good to have numbers here for M = 1, 2, 4, 8, 16 to see if we are seeing the perf gain originally mentioned in the issue https://github.com/lightseekorg/tokenspeed/issues/845

tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/linear_attnres_partials.py
```diff
@@ -0,0 +1,337 @@
...
+        if out_was_provided:
+            out.copy_(projected)
+        else:
+            out = projected
+    else:
+        torch.mm(hidden_states, weight.T, out=out)
+    triton_partial_eligible = (
+        blocks.is_cuda
+        and blocks.dtype == torch.bfloat16
+        and blocks.shape[2] == 7168
```
nirvedhmeshram: I think you may be able to use `KIMI3_HIDDEN_SIZE` here (`ops/gemm/kimi3.py:21`)?

tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/gluon.py
```diff
@@ -99,3 +109,56 @@ def gluon_bmm_a16w16_gfx950(
...
+                        score_weight_b=dense_tensor_format(torch.bfloat16),
+                        out=dense_tensor_format(torch.bfloat16),
+                    )
+                }
+            ),
+            priority=Priority.SPECIALIZED,
+            traits={
+                "tokens": frozenset({1}),
+                "input_size": frozenset({7168}),
+                "output_size": frozenset({3648, 6288}),
```
nirvedhmeshram: Two of these already have names in `ops/gemm/kimi3.py` — `KIMI3_HIDDEN_SIZE` (7168) and
`KIMI3_QKVFAB_SIZE` (6288) — and this package already imports from there. Worth using them?
(`3648` has no constant yet.)

python/tokenspeed/runtime/models/kimi_k3.py
```diff
@@ -1664,18 +1770,46 @@ def forward(
...
+                block_residual[:mlp_valid_blocks],
+                self._mlp_wp,
+                nxt_layer._attn_wp,
+                self.mlp_res_norm.variance_epsilon,
+                scratch,
+                sc1,
+            )
+            if self.self_attn.can_fuse_attnres_partials(h, candidate_args):
+                attnres_partial_args = candidate_args
+        reduce_consumes_scratch = (
```
nirvedhmeshram, line 1816: comment explaining who is writing and consuming the scratch here would be helpful.

nirvedhmeshram (commented):
Looks good to me aside from minor nits, lets see how the perf is looking before landing.

qedawkins:
> Would be good to have numbers here for M = 1, 2, 4, 8, 16 to see if we are seeing the perf gain originally mentioned in the issue #845

Will share fresh perf numbers once a machine opens up enough to try. This change is only for M = 1, I have a follow up for 2, 4, 8, 16, and 32.

qedawkins:
Ok I've been struggling to reproduce the `28%` improvement number cited in #845 but this is still giving a sizeable boost to tok/s

  Original mega_k3 A/B

```
   Revision           Output tok/s    Avg latency        TPOT
  ━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━  ━━━━━━━━━━━━━  ━━━━━━━━━━
   Parent ea4c2b30           60.09        68.16 s    16.58 ms
  ─────────────────  ──────────────  ─────────────  ──────────
   Commit f54a6578           70.89        57.78 s    14.05 ms
  ─────────────────  ──────────────  ─────────────  ──────────
   Change                   +18.0%         -15.2%      -15.3%
```

This revision

```
   Revision          Output tok/s    Avg latency
  ━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━  ━━━━━━━━━━━━━
   Base                     53.99        75.87 s
  ────────────────  ──────────────  ─────────────
   AttnRes commit           61.02        67.12 s
  ────────────────  ──────────────  ─────────────
   Change                  +13.0%         -11.5%
```

Doing a pure rebase of the old commit gives similar results to what I'm getting from this branch in terms of absolute throughput. Trying to compare direct kernel performance here is tricky since the unfused path used separate streams but this is an estimate.

```
   Output width    Projection alone    Partials alone    Unfused overlap bound    Fused kernel
  ━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━
   N=3648                  20.40 us          12.76 us           20.40–33.16 us        27.72 us
  ──────────────  ──────────────────  ────────────────  ───────────────────────  ──────────────
   N=6288                  31.40 us          12.76 us           31.40–44.16 us        26.44 us
```

These used 100 warmups and 1,000 repetitions. The unfused path used separate streams, so its real latency lies between perfect overlap and serial execution.

nirvedhmeshram (dismissed):
LGTM

python/tokenspeed/runtime/distributed/comm_backend/base.py
```diff
@@ -59,6 +59,32 @@ def prepare_all_reduce_lane(self, group: Group, hidden_dim: int) -> bool:
 
         return False
 
+    def all_reduce_residual_attnres(
```
panditsa: Would it be possible keep the Iris kernel specialized but expose it through a generic `all_reduce_with_epilogue `interface, with `attnres_combine `represented as the epilogue? The current API adds a Kimi-K3-specific operation, including its weights and scratch layout, to CommBackend and the distributed runtime. 

Here is quick an example of roughly what I was thinking about, generated with codex:

```
  ## Proposed file layout

   Responsibility                                       File                                                                                     Knows about AttnRes?    References fused Iris kernel?
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
   Generic epilogue protocol and dispatch               python/tokenspeed/runtime/distributed/comm_fusion.py                                                       No                               No
  ───────────────────────────────────────────────────  ───────────────────────────────────────────────────────────────────────────────────────  ──────────────────────  ───────────────────────────────
   AttnRes descriptor, result, and fallback             tokenspeed-kernel/python/tokenspeed_kernel/ops/attn_res/epilogue.py                                       Yes                               No
  ───────────────────────────────────────────────────  ───────────────────────────────────────────────────────────────────────────────────────  ──────────────────────  ───────────────────────────────
   Generic topology/backend selection                   python/tokenspeed/runtime/distributed/comm_backend/auto.py                                                 No                               No
  ───────────────────────────────────────────────────  ───────────────────────────────────────────────────────────────────────────────────────  ──────────────────────  ───────────────────────────────
   Adapter from typed epilogue to specialized kernel    python/tokenspeed/runtime/distributed/comm_backend/triton_allreduce.py                                    Yes                              Yes
  ───────────────────────────────────────────────────  ───────────────────────────────────────────────────────────────────────────────────────  ──────────────────────  ───────────────────────────────
   Fused Iris+AttnRes kernel and eligibility            tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/fusions/iris_attn_res.py                     Yes                       Defines it
  ───────────────────────────────────────────────────  ───────────────────────────────────────────────────────────────────────────────────────  ──────────────────────  ───────────────────────────────
   Generic Iris state and ordinary collectives          tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/iris.py                                       No                               No
  ───────────────────────────────────────────────────  ───────────────────────────────────────────────────────────────────────────────────────  ──────────────────────  ───────────────────────────────
   Constructs the epilogue request                      python/tokenspeed/runtime/models/kimi_k3.py                                                               Yes                               No

```

qedawkins: That layering sounds good, but I think we'd want to update more than just this use case to fall under a new generic epilogue fusion interface.

```
  - Residual + RMSNorm: all-reduce, add a residual, then RMSNorm. This is already exposed through fused_all_reduce.
  - Residual + RMSNorm + FP8 quantization: an existing variant of the same fusion.
  - Latent RMSNorm: all-reduce a combined routed/shared lane, then RMS-normalize its routed prefix.
  - AttnRes combine: all-reduce, residual accumulation, AttnRes combination, then output RMSNorm.
```

We'd also want to roll the nvidia side under the same interface as well. Instead I started with a narrower change that brings this PR's implementation more in line with other instances of fused comms kernels so we can add `all_reduce_with_epilogue` in a follow up. WDYT?

```
The narrow change aligns AMD with both existing patterns:

  - Specialized all-reduce epilogues remain in tokenspeed-kernel, outside CommBackend.
  - Kimi calls a specialized, capability-gated wrapper with an explicit unfused fallback.
  - NVIDIA uses the same model-to-specialized-wrapper structure for its AttnRes collective.
  - CommBackend remains limited to generic communication primitives.

The AMD and NVIDIA wrappers still differ internally because they use Iris and TRT-LLM respectively, but their layering now matches.
```

qedawkins: I drafted the API changes here: https://github.com/qedawkins/tokenspeed/tree/users/qedawkins/all-reduce-epilogue-api-draft and covers residual RMSNorm, latent-prefix RMSNorm, and AttnRes. I'd like to checkpoint the performance gains here to unblock follow up work if possible and send this out in parallel if that's ok with you.

panditsa: Yeah I was hoping for a common interface that can work for various epilogue fusion kernels. Your draft looks good!

Yeah I was running into the same issue where I had to decide whether the common interface should mean updating the kernels for  both Nvidia/AMD. I ended up keeping the default solution for Nvidia vs. picking the fused-approach for AMD.

panditsa (dismissed):
LGTM!

raikonenfnu (dismissed):

PR #968, opened 2026-08-07
perf(comm): add opt-in Triton AR+RMSNorm backend

Jeremy-wj:
## Summary (Human)

Add an alternative triton-native fused backend for AR+RMSNorm op using PyTorch symmetric memory. Results show improved performance across a majority of tested configs compared to Iris, the current default fused backend. Initial end-to-end results compared to unfused are mixed but show a promising improvement for the WS4 case. Benchmarking performed on GLM-5.2 and GPT-OSS-120B. Project effort has been paused, so defaults remain unchanged. Changes also made to unfused fallback behavior, AR+RMSNorm backend selection, and graph padding behavior.

Full project documentation, implementation history, and benchmarking results are available on companion branch [`jeremwan/triton-shmem-experiments`](https://github.com/Jeremy-wj/tokenspeed/tree/jeremwan/triton-shmem-experiments) in the source fork.

## Summary (AI)

Validated scope: contiguous BF16 `(tokens, hidden)` tensors and BF16 weights on
single-node MI350X (`gfx950`), TP 2/4/8. This is explicit opt-in only; no model
profile or default dispatch policy is added.

The full investigation, profiles, and campaign artifacts are in the
[finalized research snapshot](https://github.com/Jeremy-wj/tokenspeed/tree/508ccf879ba3c520ecd6a6dc5913f543fb244438).

## Performance evidence

The finalized research campaigns support an experimental backend, not default
promotion:

- GPT-OSS-120B WS4: +1.27% output throughput and -1.30% median TPOT in a
  five-pair screen; WS2 was inconclusive and WS8 lost.
- GLM-5.2-FP8: padded Triton beat Iris fused across the measured WS4/WS8
  captured-operator matrix on MI355X; full-model serving was not qualified.

These results were collected on the research snapshot, not rerun after the port
to upstream `c0f41e0`. Model-specific profiles remain reference material.

## Test Plan
From the repository root.

CPU/runtime integration:

```bash
PYTHONPATH=python:tokenspeed-kernel/python:tokenspeed-kernel-amd/python \
  python3 -m pytest test/runtime/test_server_args_attention_backends.py -q
# 20 passed

PYTHONPATH=python:tokenspeed-kernel/python:tokenspeed-kernel-amd/python \
  python3 -m pytest test/runtime/test_dp_sampling_routing_metadata.py -q
# 16 passed
```

MI350X communication and fallback:

```bash
HIP_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 \
PYTHONPATH=python:tokenspeed-kernel/python:tokenspeed-kernel-amd/python \
  python3 -m pytest \
  tokenspeed-kernel/test/ops/test_triton_shmem_communication.py -q
# 13 passed: WS 1/2/4/8, one-shot variants, two-shot, capture decline,
# and HIP graph replay

HIP_VISIBLE_DEVICES=0,1 \
PYTHONPATH=python:tokenspeed-kernel/python:tokenspeed-kernel-amd/python \
  python3 -m pytest \
  test/runtime/distributed/test_comm_ops.py::TestCommOps::test_declined_arrms_fusion_falls_back -q
# 1 passed
```

Two-rank benchmark smoke:

```bash
HIP_VISIBLE_DEVICES=0,1 \
PYTHONPATH=python:tokenspeed-kernel/python:tokenspeed-kernel-amd/python \
BENCH_WORLD_SIZES=2 BENCH_M_VALUES=1 BENCH_N_VALUES=2880 \
BENCH_BACKENDS=production_unfused,triton_shmem \
BENCH_N_WARMUP=3 BENCH_N_REPEAT=10 \
  python3 tokenspeed-kernel/test/ops/bench_triton_shmem_ar_rmsnorm.py
```

Both paths completed; unfused remained faster for this smoke row, so it is not
used as performance evidence.


Jeremy-wj:
Figures to support above claims in PR summary. Complete data available in [full companion branch](https://github.com/Jeremy-wj/tokenspeed/tree/508ccf879ba3c520ecd6a6dc5913f543fb244438).
GPT-OSS-120B on MI350, fused triton-shmem vs. unfused baseline
gpt-oss-120b-e2e-by-world-size

GLM-5.2 on MI355, fused triton-shmem vs. unfused baseline and default fused Iris
glm-5 2-fp8-ws4-ws8-micro-deltas

qedawkins:
@Jeremy-wj can you give a bit more detail behind choosing PyTorch symmetric memory over Iris? Mostly for my own edification since behind the scenes Iris does a lot of the IPC setup for us.

Jeremy-wj:
@qedawkins For IPC setup the new backend uses the existing pytorch symmetric memory API to set up symmetric memory, which already had functions set up in `triton.py`. HIP IPC is set up manually to ensure coarse-grained allocations (needed for performance) in `_coarse_shmem.py` and `hip_ipc.py`. All of this is enabled via setting the env vars `TS_ARNORM_BACKEND` and `TS_TRITON_SHMEM_COARSE` (default on).

If you were referring to motivation, the general idea is that pytorch symmetric memory is directly integrated as part of the torch+rocm stack and doesn't introduce external dependencies.

tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/_coarse_shmem.py
```diff
@@ -0,0 +1,196 @@
...
+    """Allocate a coarse-grained buffer and build its peer-pointer table via IPC.
+
+    ``_opened_cache`` lets a caller share one process-level dedup map across
+    several buffers (needed when two buffers may land in the same caching-allocator
+    segment, hence share one IPC handle). Raises on IPC failure so callers can
+    fall back to the fine-grained symm_mem path.
+    """
+    world_size = group.size()
+    rank = dist.get_rank(group)
+    tensor = torch.empty(shape, dtype=dtype, device=device)
```
qedawkins: The initial `torch.empty` allocation happens before the first cross-rank error exchange. If it fails on only one rank, that rank enters the fine-grained fallback while the other ranks block in `all_gather_object`, leaving the ranks in different collectives and deadlocking. Could we include allocation failure in the collective status protocol so every rank decides on the fallback together? A test injecting a rank-local allocation failure would also cover this path.

python/tokenspeed/runtime/utils/server_args.py
```diff
@@ -2020,6 +2056,11 @@ def add_cli_args(parser: argparse.ArgumentParser):
             action="store_true",
             help="Enable allreduce fusion for improved decode performance. Auto-enabled on supported single-node TP configurations.",
         )
+        parser.add_argument(
+            "--disable-allreduce-fusion",
```
qedawkins: Could this be represented as a single tri-state option, such as `--allreduce-fusion={auto,on,off}`, instead of two independent booleans? The current representation has three valid states plus the invalid case where both flags are set. A tri-state would model the intended behavior directly and match the existing `auto`-mode configuration pattern in this file.

python/tokenspeed/runtime/utils/server_args.py
```diff
@@ -733,10 +734,45 @@ def resolve_speculative_decoding(self):
             )
 
     def resolve_communication(self):
+        if self.enable_allreduce_fusion and self.disable_allreduce_fusion:
+            raise ValueError(
+                "--enable-allreduce-fusion and --disable-allreduce-fusion "
+                "are mutually exclusive"
+            )
+
+        arnorm_backend = os.environ.get("TS_ARNORM_BACKEND", "auto").strip().lower()
```
qedawkins: Could we avoid duplicating ARNorm backend selection and `triton_shmem` policy in `ServerArgs`? `tokenspeed-kernel` already owns `_arnorm_backend`, backend dispatch, state eligibility, and per-call capability checks, while ordinary communication uses the existing `AutoBackend`/`can_run` pattern for this responsibility. The missing information appears to be runtime scheduling context such as DP or speculative execution guarantees. Could that context be passed to an ARNorm backend preflight API instead of parsing `TS_ARNORM_BACKEND` and hard-coding one backend's policy here? This would keep backend names and capabilities behind the backend boundary and avoid changing `ServerArgs` for every implementation.

qedawkins (commented):
Sorry for the late reply, a few codex driven comments below. There was some offline discussion about motivation behind a new comms backend so I'll defer to those folks if they want to add details here

efric:
Addressed comments and refactored a bit to remove extraneous bits like experimental env knobs, simplify call paths, remove trivial wrappers. Also removed the custom hip ipc backend; in microbenchmarks against the default pytorch symmetric memory for the same kernel I did not observe the speed up. Also use Pytorch native barrier; performance is 1% better than custom barrier kernel. The core algorithm(s) is preserved and as advertised as in the original PR. 

I found the triton_shmem reference a bit misleading since this is also actually pytorch symmetric memory approach; will do a benchmark against existing symm_mem fused kernel to see if this can just subsume that

PR #996, opened 2026-08-08
refactor(inkling): ring sconv state with in-kernel persistence and speculative checkpoint publish

syuoni:
## What

Rework Inkling's sconv (short causal conv) state handling under MTP speculative
decoding, in four steps (one commit each):

1. **Batch checkpoint publication** — replace the eager per-destination publish
   loop (1,948 kernels, 48.8 ms/round) with a precomputed scatter (65 kernels,
   1.22 ms).
2. **One unified sconv kernel** — merge the decode kernel into the varlen
   kernel (decode = T=1); scratch write and boundary-checkpoint publish move
   in-kernel, speculative and accept-independent (full-attn KV pattern: later
   rounds overwrite rejected content). Kills the Python publish path entirely.
3. **Delete the dead columns/as-swa sconv mode** (unreachable since the pool
   refactor; −1.1k lines).
4. **Ring-addressed working state** — `conv_state` becomes `[L, slots, R, D]`;
   ring row of absolute position `p` is `p % R`, positions derive from the
   through-chunk `seq_lens`. No stored cursor: commit = the scheduler advancing
   `seq_lens`. Kills the verify stash, the post-verify select hook, the decode
   `cache_update`, and the draft lag buffer. `R = (W−1) + K + max(K−2, 0)`
   (lookback depth), computed in the registry, mirrored in the recipe's
   workspace plan, enforced by the startup byte check. Long extend chunks
   persist via a small epilogue kernel (in-kernel writes are race-free only for
   `chunk_len ≤ R−(W−1)` — the decode/verify/catch-up shapes).

## Why

Profiling bs1 MTP decode showed ~85 ms/step of eager host work between graph
replays, dominated by sconv state bookkeeping (publish loop, select, copies).
After this series the decode hot path is the conv compute kernels alone —
1 launch per conv call, in-graph, publication included.

## Notes for reviewers

- Spans `tokenspeed-kernel` and runtime; the kernel API changed shape
  (`inkling_ring_sconv` / `inkling_ring_sconv_update`), so they land together.
- Checkpoint density is dormant until the scheduler's `kMambaState` alignment
  gate is relaxed (Phase B, C++, deliberately deferred — larger blast radius;
  correctness here does not depend on it).
- Follow-ups: decode-shape re-profile; non-spec configs (R=4) covered by unit
  tests, not an e2e run.


syuoni:
> Hi @syuoni The latest main branch has undergone significant refactoring recently. Could you merge the latest main branch? Thanks.

Rebased to latest main.

syuoni:
All passed except two Qwen stages:
* https://github.com/lightseekorg/tokenspeed/actions/runs/31253096909/job/93092393977?pr=996
* https://github.com/lightseekorg/tokenspeed/actions/runs/31253096907/job/93094461006?pr=996

This PR touches no Qwen-related files. Merging.

PR #1000, opened 2026-08-08
fix(kernel): add the missing PDL wait to the MLA rope-quantize kernels

nperrin-fr:
`_fp8_quantize_kernel` and `_mla_nope_quantize_fp8_kernel` in ops/embedding/triton.py are launched with `launch_pdl=True` but never executed `griddepcontrol.wait`. A PDL-launched kernel starts while its producer is still running, and the producer's stores are only guaranteed visible after the wait — so both kernels could read half-written q/k from the projection GEMM. On Blackwell under prefill concurrency (abort storms) this intermittently produced NaN prefill logits; argmax over a NaN row picks token 0, and the input-buffer clamp then locks the request onto vocab_size - 1 ([PAD]) forever: 25/192 storm requests on 8x B300.

Both kernels sit behind apply_rope_mla and the prefill rope-quant path, so the corruption entered through two model-level call sites at once — which is why disabling PDL on either attention subgroup alone (28/192, 30/192) never fixed it and only disabling both was clean.

Add `ENABLE_PDL: tl.constexpr`, `gdc_wait()` before the first load, and `gdc_launch_dependents()` after the last store, matching the pattern the other PDL kernels in this repo already follow (ops/quantization/triton.py, mla_kv_pack_quantize_fp8). Constexpr-False branches fold away, so the AMD build is unaffected.

With this fix and PDL fully enabled everywhere (including the sites gated upstream by 1ada3424): 0/192 storm requests, 0/96 sequential.


nperrin-fr:
@HarryWu99 @dongjiyingdjy Thanks for narrowing to PDL around MLA, when you have time, can you take a look?

PR #1011, opened 2026-08-09
fix(k3): restore DSpark cache view on unified arena

jhinpan:
## Summary
- restore Kimi-K3 + DSpark startup after the unified draft-cache change
- keep one scheduler-owned arena while binding the FP8 target and BF16 MLA draft through separate typed compute views
- cover global continuation mapping, target isolation, shared ownership, and sleep/wake clearing with regression tests

## Root cause

Kimi-K3 stores target MLA KV in FP8, while `Inferact/Kimi-K3-DSpark` requires BF16 KV. After #974, `HybridKDATokenToKVPool` tried to bind all 93 target layers plus five draft continuation layers using the target pool's FP8 dtype. Startup therefore failed on `layer.93.latent_kv` with `dtype itemsize does not match plan`. Treating that field as BF16 inside the target pool is also insufficient because the homogeneous MLA accessors reinterpret it with target FP8 semantics.

This change slices the merged spec into target and draft compute views. The draft view maps local layers 0-4 to merged fields 93-97 and inherits the target's arena and runtime contract without allocating a second buffer.

## Validation

- `180 passed, 36 subtests passed` across cache planning, target/draft wiring, DSpark, and K3 model tests
- Black 26.3.1 and isort 8.0.1 checks pass on all changed files
- 8x MI355X, latest main `0f8a300c`, `Inferact/Kimi-K3-DSpark@cf6b8244`, official AIME26 YAML at `gpu-memory-utilization=0.92`:
  - DSpark: 28/30 (93.33%), eval wall 954.07 s, TTFT 1814.66 ms, TPOT 29.04 ms, 35.62 output tok/s/request, mean output 6437.7 tokens, sampled accept rate 0.26063
  - no-spec control: 21/30 (70.00%), eval wall 469.65 s, TTFT 1785.19 ms, TPOT 63.61 ms, 15.85 output tok/s/request, mean output 3130.2 tokens
  - DSpark improves TPOT by 2.19x and per-request output rate by 2.25x. Aggregate output throughput is effectively flat (~202 vs ~200 tok/s) because this run generated 2.06x more tokens and had a longer tail.

Accuracy caveat: the current AMD AIME YAML selects `--sampling-backend greedy`, which ignores the request's `temperature=1.0` and always uses argmax. The one-pass score difference is therefore a batch/kernel-numerics observation, not evidence that speculation improves model quality; repeated sampling parity should use a stochastic backend.

## Test plan
- [x] focused K3 cache-view tests
- [x] broader cache and DSpark regression suite
- [x] full AIME26 DSpark run on 8x MI355X
- [x] paired no-spec AIME26 control

### DIFF
--- a/python/tokenspeed/runtime/layers/attention/configs/mla.py
+++ b/python/tokenspeed/runtime/layers/attention/configs/mla.py
@@ -40,8 +40,9 @@ def resolve_mla_kv_cache_dtype(
 
     The public K3 DSpark checkpoint has no FP8 KV scales and its reference vLLM
     launch uses the default BF16 cache. TokenSpeed's K3 target currently requires
-    FP8 LCM storage, but the draft owns a separate pool, so keep only that pool
-    in BF16. Other MLA drafts continue to honor the global cache setting.
+    FP8 LCM storage, so the unified arena exposes a separate BF16 compute view
+    for the draft continuation fields. Other MLA drafts continue to honor the
+    global cache setting.
     """
     hf_config = getattr(model_config, "hf_config", None)
     if (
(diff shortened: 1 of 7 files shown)


jhinpan:
Superseded by #1012, which carries the identical patch with the required DCO sign-off.

PR #1012, opened 2026-08-09
fix(k3): restore DSpark cache view on unified arena

jhinpan:
## Summary
- restore Kimi-K3 + DSpark startup after the unified draft-cache change
- keep one scheduler-owned arena while binding the FP8 target and BF16 MLA draft through separate typed compute views
- cover global continuation mapping, target isolation, shared ownership, and sleep/wake clearing with regression tests

## Root cause

Kimi-K3 stores target MLA KV in FP8, while `Inferact/Kimi-K3-DSpark` requires BF16 KV. After #974, `HybridKDATokenToKVPool` tried to bind all 93 target layers plus five draft continuation layers using the target pool's FP8 dtype. Startup therefore failed on `layer.93.latent_kv` with `dtype itemsize does not match plan`. Treating that field as BF16 inside the target pool is also insufficient because the homogeneous MLA accessors reinterpret it with target FP8 semantics.

This change slices the merged spec into target and draft compute views. The draft view maps local layers 0-4 to merged fields 93-97 and inherits the target's arena and runtime contract without allocating a second buffer.

## Validation

- `180 passed, 36 subtests passed` across cache planning, target/draft wiring, DSpark, and K3 model tests
- Black 26.3.1 and isort 8.0.1 checks pass on all changed files
- 8x MI355X, latest main `0f8a300c`, `Inferact/Kimi-K3-DSpark@cf6b8244`, official AIME26 YAML at `gpu-memory-utilization=0.92`:
  - DSpark: 28/30 (93.33%), eval wall 954.07 s, TTFT 1814.66 ms, TPOT 29.04 ms, 35.62 output tok/s/request, mean output 6437.7 tokens, sampled accept rate 0.26063
  - no-spec control: 21/30 (70.00%), eval wall 469.65 s, TTFT 1785.19 ms, TPOT 63.61 ms, 15.85 output tok/s/request, mean output 3130.2 tokens
  - DSpark improves TPOT by 2.19x and per-request output rate by 2.25x. Aggregate output throughput is effectively flat (~202 vs ~200 tok/s) because this run generated 2.06x more tokens and had a longer tail.

Accuracy caveat: the current AMD AIME YAML selects `--sampling-backend greedy`, which ignores the request's `temperature=1.0` and always uses argmax. The one-pass score difference is therefore a batch/kernel-numerics observation, not evidence that speculation improves model quality; repeated sampling parity should use a stochastic backend.

## Test plan
- [x] focused K3 cache-view tests
- [x] broader cache and DSpark regression suite
- [x] full AIME26 DSpark run on 8x MI355X
- [x] paired no-spec AIME26 control


jhinpan:
CI methodology follow-up after the paired AIME26 run:

- The two K3 AIME jobs currently run exactly one pass (`repeats: 1`); unlike the GPQA job, neither command sets `--repeats`.
- Both servers select `--sampling-backend greedy`. That backend ignores EvalScope's `temperature=1.0` / `do_sample=true` and always uses argmax.
- The jobs run independently and compare each score only to a fixed 0.90 floor; CI does not compute a no-spec/DSpark accuracy delta or confidence interval.

Therefore the 21/30 vs 28/30 observation is not an accuracy-equivalence result. A separate manual parity protocol should first validate the ROCm stochastic backend, then run both arms with matched sampling (`temperature=1.0`, `top_p=0.95`), shared seeds, and at least four repeats, reporting mean pass@1 + SEM / paired delta. Only after that gate should wall time be interpreted; TPOT and output-token throughput should remain primary because completion-length drift makes raw suite wall time incomparable.

jhinpan:
Additional paired stochastic smoke on 8x MI355X (`sampling-backend=triton`, `temperature=1.0`, `top_p=0.95`, explicit seed 42):

- no-spec completed 30/30: 20 correct (66.67%), eval wall 1051.14 s, TTFT 1343.52 ms, TPOT 60.95 ms, 18.97 tok/s/request, mean output 4958.8 tokens.
- DSpark completed 29/30 within the 30-minute shared-node lease: 26/29 correct (89.66% over completed samples); one long-tail request was stopped at the lease boundary, so there is no valid full-run score or aggregate perf report.

This validates that the Triton stochastic path runs, but one paired seed still does not establish accuracy equivalence. It also shows why the current one-pass greedy CI cannot be used for that claim. A practical parity gate needs at least four matched-seed repeats, split across leases (or run on a dedicated CI node), and should compare paired mean pass@1/SEM before interpreting speed.

jhinpan:
Exact latest-main A/B after merging `bcb3afd7` into this PR (`14071607`), using the same official greedy AIME26 protocol on the same 8x MI355X node:

- A — `origin/main@bcb3afd7`, no spec: 22/30 (73.33%); eval wall 478.17 s; mean latency 192.77 s; TTFT 1469.63 ms; TPOT 63.03 ms; 16.33 output tok/s/request; 94,458 output tokens (mean 3,148.6).
- B — PR #1012, DSpark: 28/30 (93.33%); eval wall 884.43 s; mean latency 158.21 s; TTFT 1592.73 ms; TPOT 28.75 ms; 32.49 output tok/s/request; 154,204 output tokens (mean 5,140.1).

Derived:
- no accuracy regression in this pass (DSpark is +6 questions), but greedy batch/kernel numerics mean this is not an equivalence proof.
- TPOT improves 2.19x and mean per-request output rate 1.99x; mean request latency falls 17.9% despite 63.3% longer completions.
- high-concurrency aggregate output throughput is 174.4 vs 197.5 tok/s (DSpark 0.88x), and suite wall is 1.85x longer because of output-length/tail drift. Therefore this run shows a latency win, not an AIME high-throughput/wall-clock win.

The first latest-main launch exposed a stale locally installed scheduler ABI after #1002 (`Scheduler.clear_cache` missing); rebuilding `tokenspeed-scheduler` from `bcb3afd7` resolved it. The DSpark report completed successfully; the outer process later hung in teardown, after report generation.

jhinpan:
@torchspec-bot I reproduced the #1015 behavior on latest main `bcb3afd7` by making the K3 draft follow the target's FP8 KV dtype, then ran the same official greedy AIME26 job on the same 8x MI355X node.

Compared with this PR's BF16 draft view:

- cache parent bytes: 30,081,024 -> 25,657,344 (-14.7%)
- admitted token capacity: 2,965,120 -> 3,562,624 (+20.2%)
- sampled accept length: roughly 2.6-2.8 -> 1.79 (`accept_rate=0.112391`)
- TPOT: 28.75 -> 33.10 ms (15.1% slower)
- per-request output rate: 32.49 -> 26.20 tok/s (19.4% lower)
- one-pass AIME: 28/30 -> 21/30

The accuracy delta is not an attribution-quality result because this CI is greedy and batch/kernel numerics change trajectories, but the acceptance and normalized per-token regressions are direct. The FP8 run also emits the expected warning that the checkpoint provides no KV scales and scale 1.0 is being used. PR #1015's checks do not include the manual K3+DSpark model gate.

My recommendation: a separate user-facing draft KV dtype is a good API, and #1012's typed MLA backing view makes it possible. Keep checkpoint-faithful BF16 as K3's `auto` default; expose FP8 as an explicit capacity-vs-acceptance opt-in rather than forcing the draft to follow the target.

chenht2022 (approved):

PR #1017, opened 2026-08-09
fix(iris): close all-reduce payload lifetime race

raikonenfnu:
## Summary

Fix a cross-rank lifetime race in IRIS one-shot all-reduce when consecutive calls reuse the same symmetric input buffer.
To fix we do:

1. Track per block completion - add a `done_flag`, so other rank knows it is safe to overwrite.
2. Wait at the next overwrite, not at the current return - Wait to be sure it is safe before overwriting data.
3. Functionalize the symmetric-memory epoch spin-wait - Make the iris spin lock into function so easy to re-use + use tl.load instead of iris.load for vectorization/better perf. 

## Visualization

```text
rank A: write payload N -> ready[N] -> finish its own reduction -> overwrite with N+1
rank B:                         sees ready[N] -> still reading payload N -----^
```

`ready[N]` answers “has rank A produced N?” It does not answer “has rank B consumed N?” Local stream ordering cannot protect storage read remotely by another rank.

The corrected two-rank timeline adds the matching consumption proof:

```text
rank A: wait peer done[N-1] -> write A[N] -> ready_A[N] -> read B[N] -> done_A[N]
rank B: wait peer done[N-1] -> write B[N] -> ready_B[N] -> read A[N] -> done_B[N]

rank A may overwrite A[N] with A[N+1] only after observing done_B[N]
rank B may overwrite B[N] with B[N+1] only after observing done_A[N]
```

## Validation

Validated on commit `5a1802ef` after rebasing onto current main:

```bash
PYTHONPATH="$PWD/tokenspeed-kernel-amd/python:$PWD/tokenspeed-kernel/python:$PWD/tokenspeed-kernel/test" \
pytest -q -s \
  tokenspeed-kernel/test/ops/test_iris_communication.py::test_iris_all_reduce_correctness_world2 \
  tokenspeed-kernel/test/ops/test_iris_communication.py::test_iris_all_reduce_correctness_world4 \
  tokenspeed-kernel/test/ops/test_iris_communication.py::test_iris_all_reduce_buffer_reuse_and_graph_replay_world8
# 3 passed

pre-commit run --all-files
# all hooks passed
```

The world-8 stress case also passed three independent runs at 256 eager epochs plus 256 graph replays per production shape.

### DIFF
--- a/tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/iris.py
+++ b/tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/iris.py
@@ -513,6 +513,36 @@ class IrisAllReduce(object):
 
 
 @triton.jit
+def _iris_wait_for_peers_triton(
+    flags,
+    block_id,
+    epoch,
+    heap_bases,
+    RANK: tl.constexpr,
+    WORLD_SIZE: tl.constexpr,
+):
+    """Acquire every peer's published epoch."""
+    local_heap = tl.load(heap_bases + RANK)
+    flags_heap_offset = tl.cast(flags, tl.uint64) - local_heap
+    for peer in tl.static_range(0, WORLD_SIZE):
+        if peer != RANK:
+            peer_heap = tl.load(heap_bases + peer)
+            peer_flag = tl.cast(
+                peer_heap + flags_heap_offset,
+                tl.pointer_type(tl.int32),
+            )
+            peer_flag += block_id * WORLD_SIZE + peer
+            seen = tl.full((), 0, dtype=tl.int32)
+            while seen < epoch:
+                seen = tl.atomic_add(
+                    peer_flag,
+                    0,
+                    sem="acquire",
+                    scope="sys",
+                )
+
+
+@triton.jit
 def iris_stage_one_shot_allreduce_kernel(
     input_ptr,
     input_sym_ptr,
@@ -537,18 +567,14 @@ def iris_stage_one_shot_allreduce_kernel(
     epoch = tl.load(local_ready).to(tl.int32) + 1
     tl.atomic_xchg(local_ready, epoch, sem="release", scope="sys")
 
-    for peer in tl.static_range(0, WORLD_SIZE):
-        if peer != RANK:
-            seen = tl.full((), 0, dtype=tl.int32)
-            while seen < epoch:
-                seen = iris.load(
-                    ready_flags + flag_offset + peer,
-                    RANK,
-                    peer,
-                    heap_bases,
-                    cache_modifier=".cv",
-                    volatile=True,
-                )
+    _iris_wait_for_peers_triton(
+        ready_flags,
+        block_id,
+        epoch,
+        heap_bases,
+        RANK,
+        WORLD_SIZE,
+    )
 
     acc = local.to(tl.float32)
     for peer in tl.static_range(0, WORLD_SIZE):
@@ -596,7 +622,7 @@ def _iris_heap_base(
 
 
 @gluon.jit
-def _iris_wait_for_rank_epoch(
+def _iris_wait_for_peers_gluon(
     ready_flags,
     block_id,
     epoch,
@@ -613,7 +639,7 @@ def _iris_wait_for_rank_epoch(
     WORLD_SIZE: gl.constexpr,
     NUM_WARPS: gl.constexpr,
 ):
-    """Poll rank epochs with a layout matching the specialized launch."""
+    """Acquire every peer's published epoch."""
     ready_layout: gl.constexpr = gl.BlockedLayout([1], [64], [NUM_WARPS], [0])
     peer_ids = gl.arange(0, WORLD_SIZE, layout=ready_layout)
     peer_heaps = gl.where(peer_ids == 0, heap_base_0, heap_base_7)
@@ -632,12 +658,12 @@ def _iris_wait_for_rank_epoch(
     peer_mask = peer_ids != RANK
     seen = gl.full([WORLD_SIZE], 0, gl.int32, layout=ready_layout)
     while gl.min(gl.where(peer_mask, seen, epoch), axis=0) < epoch:
-        seen = gl.load(
+        seen = gl.atomic_add(
             peer_flags,
+            0,
             mask=peer_mask,
-            other=epoch,
-            cache_modifier=".cv",
-            volatile=True,
+            sem="acquire",
+            scope="sys",
         )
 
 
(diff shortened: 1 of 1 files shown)


raikonenfnu:
Moved from the main PR description for easier focused review.

## Key changes

### 1. Track per-block consumer completion

[`IrisAllReduce.__init__`](https://github.com/raikonenfnu/tokenspeed/blob/5a1802ef/tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/iris.py#L352-L367) allocates `done_flags` with the same `(block, rank)` layout as `ready_flags`:

```python
self._ready_flags = self._ctx.zeros(
    (self._max_blocks, self.world_size), dtype=torch.int32
)
self._done_flags = self._ctx.zeros(
    (self._max_blocks, self.world_size), dtype=torch.int32
)
```

The extra memory is one `int32[max_blocks, world_size]` symmetric array (from `done_flags`). Both [`all_reduce`](https://github.com/raikonenfnu/tokenspeed/blob/5a1802ef/tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/iris.py#L382-L423) and [`all_reduce_two`](https://github.com/raikonenfnu/tokenspeed/blob/5a1802ef/tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/iris.py#L425-L520) pass it to their kernels.

### 2. Wait at the next overwrite, not at the current return

The one-shot protocol is now:

```text
before writing epoch N:
    wait until every peer reports done[N-1]

produce epoch N:
    write local payload with .wt
    publish ready[N] with system-scope release
    wait until every peer reports ready[N]

consume epoch N:
    read peer payloads
    publish done[N] after those reads complete
    store/return the local result without waiting for peer done[N]
```

The key ordering in the [single-tensor Triton kernel](https://github.com/raikonenfnu/tokenspeed/blob/5a1802ef/tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/iris.py#L556-L626) is:

```python
epoch = tl.load(local_ready).to(tl.int32) + 1

# Protect the old contents only when this rank is about to overwrite them.
_iris_wait_for_peers_triton(done_flags, block_id, epoch - 1, ...)

tl.store(input_sym_ptr + offsets, local, cache_modifier=".wt", ...)
tl.debug_barrier()
tl.atomic_xchg(local_ready, epoch, sem="release", scope="sys")

_iris_wait_for_peers_triton(ready_flags, block_id, epoch, ...)
acc += iris.load(input_sym_ptr + offsets, ..., cache_modifier=".cg")

# Tell producers this consumer has finished; do not wait here.
tl.debug_barrier()
tl.atomic_xchg(local_done, epoch, sem="release", scope="sys")
tl.store(output_ptr + offsets, ...)
```

This restores the invariant while allowing the current output to progress. A peer completion wait is paid only by a later collective that actually wants to reuse the same block.

The same ordering is implemented in:

- [`iris_stage_one_shot_allreduce_two_kernel`](https://github.com/raikonenfnu/tokenspeed/blob/5a1802ef/tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/iris.py#L886-L975), the generic two-tensor Triton path.
- [`iris_stage_one_shot_allreduce_two_gluon_kernel`](https://github.com/raikonenfnu/tokenspeed/blob/5a1802ef/tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/iris.py#L728-L883), the world-8 gfx950 K3-specialized path.

### 3. Functionalize the symmetric-memory epoch spin-wait

The original generic kernels embedded the same serial spin-wait directly in each ready path:

```python
for peer in tl.static_range(0, WORLD_SIZE):
    if peer != RANK:
        seen = 0
        while seen < epoch:
            seen = iris.load(ready_flags + flag_offset + peer, RANK, peer, ...)
```

This is epoch polling rather than a mutual-exclusion lock: a consumer repeatedly observes monotonic peer epochs until every required publication is visible. The PR extracts that behavior into `_iris_wait_for_peers_triton(flags, block_id, epoch, ...)`, so the identical operation can protect both sides of the lifetime handshake:

```python
_iris_wait_for_peers_triton(done_flags, ..., epoch - 1, ...)  # safe to overwrite
_iris_wait_for_peers_triton(ready_flags, ..., epoch, ...)     # safe to consume
```

Once expressed as a reusable operation, the helper loads all ready/done epochs as one small vector, masking the local rank and padding up to the next power of two:

```python
peer_ids = tl.arange(0, PEER_BLOCK)
peer_mask = (peer_ids < WORLD_SIZE) & (peer_ids != RANK)
...
while tl.min(tl.where(peer_mask, seen, epoch), axis=0) < epoch:
    seen = tl.load(
        peer_flags,
        mask=peer_mask,
        other=epoch,
        cache_modifier=".cv",
        volatile=True,
    )
```

See [`_iris_wait_for_peers_triton`](https://github.com/raikonenfnu/tokenspeed/blob/5a1802ef/tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/iris.py#L523-L553) and the equivalent [`_iris_wait_for_peers`](https://github.com/raikonenfnu/tokenspeed/blob/5a1802ef/tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/iris.py#L658-L701) Gluon helper.

Only the changing epoch flags use volatile coherent loads. Payload loads retain the existing streaming `.cg` policy.

tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/iris.py
```diff
@@ -508,43 +520,89 @@ def all_reduce_two(
...
+
+    seen = tl.zeros((PEER_BLOCK,), dtype=tl.int32)
+    while tl.min(tl.where(peer_mask, seen, epoch), axis=0) < epoch:
+        seen = tl.load(
+            peer_flags,
+            mask=peer_mask,
+            other=epoch,
+            cache_modifier=".cv",
+            volatile=True,
+        )
```
efric: This might be a good place to use a `tl.atomic_poll` with `acquire` semantics. I don't see any matching `acquire` currently, so although unlikely, I wonder if it may be possible for us to e.g load stale values as a consumer even though we observed the producer has publicized it's done

efric (commented):
LGTM, just one comment

tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/iris.py
raikonenfnu: Makes sense, done!

raikonenfnu:
Closing as superseded by #1038, implemented by @panditsa. That PR replaced K3’s staged `all_reduce_two` path with producer-direct writes into Iris-owned symmetric views, and `iris_reduce_symmetric_gluon_kernel` now publishes a ready epoch before peer reads plus a completion epoch after those reads, waiting for every peer before returning so the next producer cannot overwrite a payload that is still being consumed. This resolves the same symmetric-buffer lifetime hazard in the new design and removes the old path where this PR’s separate `done_flags` and wait-before-overwrite protocol applied; thank you @panditsa for carrying the fix into the productionized fused path.

PR #1019, opened 2026-08-09
perf(gfx950): stop using FP32 atomics for the SiTU MoE stage-2 combine

antiagainst:
## Summary

The gfx950 A16W4 SiTU MoE stage-2 combine selected the FP32 atomic epilogue for
`num_tokens <= 16`, on the premise that atomics win when few routes contend.
They don't — the cost tracks the *padded* route list, not the token count, and
under sparse EP the padding is nearly all of it. `GROUPED_ATOMIC_COMBINE_MAX_TOKENS`
moves to 0.

Tightend `atol` for SiTU MoE checks along the way.

## Why atomics lose

The kernel doesn't walk the list of real (token, expert) pairs. It walks a padded
version of that list: `moe_align` rounds every expert up to a whole 64-row block,
so an expert holding a single pair still costs 64 rows. With 112 experts on this
GPU that's ~7168 rows, and it stops growing past 7 tokens — from there on, adding
tokens adds real work but no rows.

That would be harmless if the padding rows were skipped, and on the scratch-buffer
path they are: the store is wrapped in a branch, so a block that's entirely padding
jumps over it. The atomic path can't do that. It hides the padding by pointing the
write at an address the hardware throws away, which means the instruction still
issues on every lane. So the number of atomic adds sent to memory is
`padded rows × output width`, no matter how much of it is real.

At 16 tokens that's 25.7 million atomic adds issued to perform 0.92 million useful
ones — about 4% useful.

## Numbers (Kimi K3 shape, 3584×512, 112 experts, 16 experts per token, MI350X)

End-to-end time, atomics ÷ scratch buffer: **0.89× at 1 token, 1.21× at 2, 2.20×
at 4, and 2.66–3.06× from 8 through 128.** So atomics are behind from 2 tokens up.
They do win at 1 token, but only because the scratch path pays for a second kernel
launch (~28 µs of it exposed), and a separate kernel handles single-token decode
anyway.

Two controlled runs show it's the padding and not the real work. Each varies one
and pins the other (kernel time; scratch = both kernels summed):

| run | real (token, expert) pairs | experts | padded rows | atomics | scratch |
|---|---:|---:|---:|---:|---:|
| **A** — same work, less padding | 256 (16 tok) | 112 | 7168 | 458 µs | 74 µs |
| | 256 (16 tok) | 16 | 1024 | 41 µs | 30 µs |
| | 256 (16 tok) | 4 | 256 | 30 µs | 27 µs |
| **B** — same padding, more work | 112 (7 tok) | 112 | 7168 | 456 µs | 73 µs |
| | 2048 (128 tok) | 112 | 7168 | 503 µs | 112 µs |

Cutting padding 7× with the work unchanged makes atomics 11× faster. Raising the
work 18× with the padding unchanged costs them 10%. The cost is essentially all
padding.

The threshold constant stays instead of deleting the atomic path, so this is easy
to re-measure if the padding ever stops dominating. A new test runs the atomic
path (now unreachable in normal use) against the scratch path and checks they
agree.

### DIFF
--- a/tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/moe/mxfp4/situ_grouped.py
+++ b/tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/moe/mxfp4/situ_grouped.py
@@ -52,11 +52,6 @@ GROUPED_FUSED_ALIGN_MAX_ROUTES = 128
 # not skip the instruction.  A 16-token batch therefore issues 25.7M atomics to
 # perform 0.92M useful ones.
 #
-# Measured on the K3 shape (3584x512, 112 experts, top-k 16), atomics are
-# slower than the scratch buffer from 2 tokens up: 1.2x at 2 tokens, 2.2x at
-# 4, and 2.6-3.0x from 8 through 128.  They only come out ahead at 1 token,
-# and a different kernel handles that batch anyway.  Hence 0.
-#
 # The constant stays instead of deleting the atomic path outright, so the
 # comparison is easy to redo if the padding ever stops dominating.
 GROUPED_ATOMIC_COMBINE_MAX_TOKENS = 0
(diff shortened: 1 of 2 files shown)


panditsa (approved):
lgtm!

PR #1020, opened 2026-08-10
refactor(amd): split gfx950 mxfp4 decode kernels

antiagainst:
Split decode_kernels.py by ownership: move the direct MFMA stage kernels and invokers into decode_stage1.py and decode_stage2.py, shared addressing, load, and MFMA primitives into decode_common.py, and dense top-k routing into routing.py. Update SiTU decode to use the shared scale helper and remove the old implementation module after all in-tree callers have moved.

Consolidate the routing dtype and power-of-two helpers shared by decode and prefill. Add direct reference coverage for both MFMA stages and update routing tests to patch the owning module.

The moved function bodies and launch configurations are unchanged.


panditsa (approved):
LGTM! The new test adds the missing stage1/stage2 specific tests, and I updated one of the existing test that cover E2E run with non-zero weights.

PR #1021, opened 2026-08-10
cleanup(amd): drop dead gfx950 mxfp4 decode entry point and warp-GEMV path

antiagainst:
Two unreferenced paths in the gfx950 A4W4 MoE package:

* ``moe.py``'s ``gluon_mxfp4_moe_decode`` was only reachable through the package ``__init__`` re-export.  Production dispatch calls ``_maybe_precomputed_mxfp4_direct_mfma_decode`` in ``fused.py``, which duplicates the same quantize / stage1 / stage2 sequence.
* The BF16-activation ``_stage{1,2}_mxfp4_warp_gemv_gluon`` kernels, their ``invoke_stage{1,2}_warp_decode_gluon`` launchers, and the ``_legacy_a16w4_*`` drivers that called them.  Their docstring claimed they were retained for microbenchmarks through ``mxfp4_warp_decode_gfx950``, a name that does not exist in the tree; the identically named helpers the fp16 package imports live in ``fp16/warp_decode_gluon_kernel.py``.  Removing them also orphans ``_e2m1_to_f32`` and ``_LANES``, their only users.

Deletion only: no surviving kernel or launcher is modified.  ``ops/moe/`` passes 128 tests with 31 skips.


raikonenfnu (approved):
LGTM!

panditsa (approved):

PR #1038, opened 2026-08-10
perf(kimi3): fuse MoE norm projection and collectives

panditsa:
### Packed Kimi K3 decode pipeline

Kimi K3 projects the same hidden state into router logits, the common routed latent, and the shared-expert gate/up input. This PR packs those three weights so the decode path can compute the projections together, then runs routed W13/W2 beside the shared down projection. The specialized M=1 epilogue folds the optional routed RMSNorm, the 3584→7168 projection, `prefix_sum`, and the shared output into the final projection; warp decode remains selected through M=16.

### Producer-direct multi-tensor all-reduce

`all_reduce` now accepts either one tensor or a tuple of independent tensors. An eligible AMD producer can first call `acquire_all_reduce_outputs`, write routed and shared results directly into the returned symmetric-memory views, and pass that tuple to the ordinary `all_reduce` entry point:

```python
outputs = acquire_all_reduce_outputs((routed_shape, shared_shape), like, group)
producer(..., routed_out=outputs[0], shared_out=outputs[1])
routed_out, shared_out = all_reduce(outputs, group)
```

Iris recognizes its owned views and reduces all outputs in one launch, avoiding an intermediate concatenation or copy. Ineligible AMD requests use grouped RCCL, while NVIDIA retains the existing NCCL/TRT-LLM behavior and does not probe the Iris path.

### Generalized AMD Iris path

The producer-direct reducer now supports world sizes 2, 4, and 8, including non-contiguous subgroups, with capacity expressed in bytes rather than a Kimi-specific element count. Packing derives from the dtype: BF16/FP16 use four elements per 64-bit word and FP32 uses two. Unsupported dtypes, operations, alignments, and capacities fall back to the existing collective backend.

### CUDA-graph and AttnRes coverage

The gfx950 Attention-Residual eligibility and workspace limits now cover the larger Kimi K3 graph workloads used by M=8 and M=16. Platform-specific admission remains behind the kernel selector, so the existing non-CDNA4 path is unchanged.

## Test Plan

- `python -m ruff check --select F401,F841,F821,F823,F811,F541,UP035,UP006,UP007,UP045 python/tokenspeed/runtime`: passed.
- Runtime/backend/latent/Kimi policy tests: 87 passed.
- gfx950 projection, norm-projection, MoE reference, and CUDA-graph tests: 58 passed.
- Iris communication suite: 35 passed across world sizes 2/4/8, BF16/FP16/FP32, graph replay, and non-contiguous subgroups.
- TP8/EP8 4K/1K sweep: all 93 measured requests completed with exact input/output lengths.
- TP8/EP8 coherency canary: passed; short and long output hashes match the previously validated producer-direct build.

## Performance

Kimi K3 TP8/EP8, 4K input / 1K output, CUDA graphs, median of three:

| M | Output throughput | TTFT | TPOT |
|---:|---:|---:|---:|
| 1 | 50.46 tok/s | 465.85 ms | 19.38 ms |
| 2 | 68.27 tok/s | 760.98 ms | 28.58 ms |
| 4 | 122.57 tok/s | 1374.42 ms | 31.32 ms |
| 8 | 198.05 tok/s | 2799.15 ms | 37.69 ms |
| 16 | 299.42 tok/s | 5887.06 ms | 47.72 ms |


python/tokenspeed/runtime/distributed/comm_backend/auto.py
```diff
@@ -153,6 +153,68 @@ def all_reduce_two(
     def prepare_all_reduce_lane(self, group: Group, hidden_dim: int) -> bool:
         return self._trtllm_ar.ensure_group_lane(group, hidden_dim)
 
+    def prepare_all_reduce_two(
```
panditsa: change the name, add docstring

python/tokenspeed/runtime/distributed/comm_backend/base.py
```diff
@@ -59,6 +59,46 @@ def prepare_all_reduce_lane(self, group: Group, hidden_dim: int) -> bool:
 
         return False
 
+    def prepare_all_reduce_two(
```
panditsa: ditto

python/tokenspeed/runtime/distributed/comm_backend/triton_allreduce.py
```diff
@@ -113,6 +116,86 @@ def all_reduce_two(
             return all_reduce_two(state, first, second, op=op)
         return super().all_reduce_two(first, second, group, op=op)
 
+    def prepare_all_reduce_two(
```
panditsa: why do we need this?

python/tokenspeed/runtime/layers/moe/latent.py
```diff
@@ -531,15 +548,17 @@ def run_shared_branch() -> None:
         elif self.latent_reduce is not None:
             routed_latent = self.latent_reduce(routed_latent)
             _check_shape(routed_latent, latent_shape, "latent_reduce")
-        if self.routed_norm is not None:
-            routed_latent = _module_tensor_output(self.routed_norm, routed_latent)
-            _check_shape(routed_latent, latent_shape, "routed_norm")
-
         if shared_output is None:
+            if self.routed_norm is not None:
```
panditsa, line 628: repeated code, add docstring

tokenspeed-kernel/python/tokenspeed_kernel/ops/attn_res/__init__.py
```diff
@@ -20,18 +20,18 @@
...
 
-# Blackwell kernel coverage. This must stay a subset of the authoritative
-# TVM_FFI_ICHECK bounds in csrc/attn_res_binding.cu; this gate only decides when
-# to fall back to torch.
+# Common specialized-kernel coverage. The larger gfx950 bound is restricted to
+# the Kimi K3 shape and its fused output RMSNorm below.
 _SUPPORTED_H = frozenset({4096, 5120, 6144, 7168, 8192})
-_MAX_T = 16384
+_MAX_TOKENS = 16384
+_MAX_GFX950_TOKENS = 65536
```
panditsa, line 33: validate this range

tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/iris.py
```diff
@@ -355,8 +356,11 @@ def __init__(
         self._input_buf = self._ctx.zeros((max_numel,), dtype=dtype)
         self._block_size = 2048
         self._max_blocks = triton.cdiv(max_numel, self._block_size)
+        # Private epochs are reserved for exact K3 specializations so ordinary
+        # all-reduce launches cannot advance their staging counters. The K3
+        # two-segment reduction uses 21 half-width staging blocks.
         self._ready_flags = self._ctx.zeros(
-            (self._max_blocks, self.world_size), dtype=torch.int32
+            (self._max_blocks + 64, self.world_size), dtype=torch.int32
```
panditsa: why?

tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/triton.py
```diff
@@ -2021,6 +2024,129 @@ def all_reduce_two(
...
+    output_weight: torch.Tensor,
+    scratch: tuple[torch.Tensor, torch.Tensor, torch.Tensor],
+    op=None,
+) -> bool:
+    """Return whether Iris can finish the exact K3 attention prefix mix."""
+    if op is None:
+        op = torch.distributed.ReduceOp.SUM
+    m, s_, acc = scratch
+    platform = current_platform()
+    return (
```
panditsa: do we need all these checks?

python/tokenspeed/runtime/distributed/comm_backend/triton_allreduce.py
```diff
@@ -76,42 +91,64 @@ def can_run(self, tensor: torch.Tensor, group: Group, op=None) -> bool:
...
+        op=None,
+    ) -> tuple[torch.Tensor, ...]:
+        """Acquire symmetric outputs when Iris supports the request."""
+        if not self.can_acquire_outputs(shapes, like, group, op=op):
+            return super().acquire_all_reduce_outputs(shapes, like, group, op=op)
+
+        try:
+            state = self._get_or_create(group)
+            return acquire_symm_outputs(state, shapes, like.dtype)
+        except Exception:
```
qedawkins: [P1] **Producer-direct fallback can deadlock the group.**

This fallback decision is rank-local. If acquisition fails on one rank, that rank returns ordinary tensors while peers may retain Iris-owned buffers. The later all-reduce then selects different collective protocols across ranks and can hang. Please either treat acquisition failure as fatal, or coordinate a group-wide success flag outside graph capture so every rank falls back together.

python/tokenspeed/runtime/distributed/comm_ops.py
```diff
@@ -126,41 +126,17 @@ class FusionParams:
 
 
 def all_reduce(
-    tensor: torch.Tensor,
+    tensor: torch.Tensor | tuple[torch.Tensor, ...],
```
qedawkins, line 129: [P2] **Collection `all_reduce` is not implemented by every advertised backend.**

This API now accepts either a single `torch.Tensor` or a tuple of tensors, but `NcclBackend.all_reduce` and `TrtllmAllReduceBackend.all_reduce` still implement the single-tensor case only. A direct backend override therefore passes a tuple into NCCL or calls `.numel()` on a tuple. Please update those overrides to detect collections and delegate them to `super().all_reduce(...)`, while preserving their existing optimized single-tensor paths.

python/tokenspeed/runtime/distributed/comm_backend/triton_allreduce.py
```diff
@@ -23,28 +23,42 @@
...
+        fallback: CommBackend,
+        producer_direct_max_bytes: int = _DEFAULT_PRODUCER_DIRECT_MAX_BYTES,
+    ):
         self._fallback = fallback
         self._instances = {}
-        self._max_bytes = 512 * 1024
+        self._producer_direct_max_bytes = producer_direct_max_bytes
         self._max_numel = (
-            self._max_bytes // torch.empty((), dtype=torch.bfloat16).element_size()
+            producer_direct_max_bytes
```
qedawkins: [P2] **Producer-direct capacity also changes ordinary all-reduce dispatch.**

This couples the new 1 MiB producer-direct capacity to ordinary single-tensor Iris eligibility, silently moving 512 KiB-1 MiB AMD all-reduces from NCCL to Iris. Please keep a separate 512 KiB ordinary all-reduce admission threshold while retaining the larger backing capacity for producer-direct collections, unless the wider single-tensor range has comparative benchmarks supporting the dispatch change.

tokenspeed-kernel/test/ops/moe/test_gluon_mxfp4_situ_gfx950.py
```diff
@@ -532,7 +532,7 @@ def test_mxfp4_situ_virtual_ep_sum_matches_global_reference_gfx950(
     torch.testing.assert_close(actual, expected, atol=3e-4, rtol=3e-2)
 
 
-@pytest.mark.parametrize("num_tokens", [1, 2, 4, 8])
+@pytest.mark.parametrize("num_tokens", [1, 2, 4, 8, 16])
```
qedawkins: [P3] **Grouped MoE caller-owned output is not directly covered.**

Every case here has `M <= 16`, so all of them select the route-direct implementation. The newly added grouped `out=` path begins at `M=17` and remains untested directly. Please add an `M=17` or `M=32` case that verifies grouped-path selection, output pointer identity, numerics, and graph replay.

qedawkins (commented):
Overall looks good, trying some Codex driven feedback below

qedawkins (approved):

PR #1053, opened 2026-08-11
perf(kernel): accelerate GLM FP8 DSA attention

Max191:
GLM 5.2 uses dense FP8 query and key/value data for selected-key attention. The existing AMD tiled path supported BF16 but did not cover the production FP8, rank-512 shape, leaving long-prompt prefill and decode on a slower generic implementation.

Extend the tiled kernel to FP8 with data-type-specific matrix instruction shapes, vector widths, and shared-memory layouts. Register it only for the supported dense FP8 shape so packed sparse and other layouts retain their existing implementations. Bound the tile loop by the valid-entry count when the visible context is shorter than the fixed-width selection buffer, while retaining one tile so zero-entry rows still produce a defined zero output.

Verification on an AMD Instinct MI355X against ToM:
- Eighteen DSA kernel-selection tests passed, and four production-shape GPU tests passed for decode and prefill.
- Decode comparisons covered 0, 1, 31, 32, 33, and 2,048 valid entries. All outputs were finite, the empty case returned zero, and the worst maximum absolute difference from the upstream implementation was 0.03515625.
- The 8,192-query prefill comparison had maximum absolute difference 0.0234375 and mean absolute difference 0.00115871.

Isolated GPU timing after two warmups:
- Decode with 4 valid entries took 21.240 us median over 50 iterations, versus 69.860 us upstream, a 3.29x speedup.
- Decode with 2,048 valid entries took 120.541 us median over 50 iterations, versus 314.823 us upstream, a 2.61x speedup.
- Prefill with 8,192 queries and 2,048 selected entries took 3.815 ms median over 5 iterations, versus 494.874 ms upstream, a 129.72x speedup.

### DIFF
--- a/tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/attention/dsa/attention.py
+++ b/tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/attention/dsa/attention.py
@@ -31,6 +31,7 @@ __all__ = [
 ]
 
 _REGISTERED_TOPK_WIDTHS = (512, 1024, 2048)
+_FP8_DTYPES = (torch.float8_e4m3fn, torch.float8_e5m2)
 
 
 @gluon.constexpr_function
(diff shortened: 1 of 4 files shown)


tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/attention/dsa/attention.py
```diff
@@ -858,6 +883,8 @@ def _run_dense_kv(
         topk_slots.shape[1],
         D_V=kv_lora_rank,
         D_ROPE=qk_rope_head_dim,
+        IS_FP8=q.dtype == torch.float8_e4m3fn,
```
Yu-Zhewen: Nit: Maybe have both E4M3 and E5M2. I don’t see E5M2 being used currently, but it may be worth keeping both.

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/gluon/__init__.py
```diff
@@ -924,6 +924,39 @@ def gluon_dsa_prefill_topk_fp8_gfx950(*args, **kwargs):
...
+            min_arch_version=ArchVersion(9, 5),
+            max_arch_version=ArchVersion(9, 5),
+            vendors=frozenset({"amd"}),
+        ),
+        signatures=frozenset(
+            {
+                format_signature(q=dense_tensor_format(torch.float8_e4m3fn)),
+            }
+        ),
+        priority=Priority.SPECIALIZED + 1,
```
Yu-Zhewen: I wonder any particular reason to use `Priority.SPECIALIZED + 1` instead of `Priority.SPECIALIZED`?

Yu-Zhewen: Also, should we rename the existing kernels to "_bf16_"?

Max191: The existing kernel covers fp8 with 2 different kv cache layouts and also 2 different `kv_lora_rank`, but this one only  supports dense kv cache with 512 kv_lora_rank, so it partially overlaps with the other kernel. We need to add 1 so it gets chosen in the overlapping case. For prefill, the existing kernel doesn't support FP8 I think, so I can rename that one to bf16

Max191: Actually, looking at it again, the new registrations are no different from the old registrations. I don't think they are necessary at all.

Yu-Zhewen (approved):
LGTM, thanks!

PR #1062, opened 2026-08-12
perf(kimi3): fold the MoE finalize into the multicast latent tail (+ extract the K3 comm layer)

dongjiyingdjy:
## What

Two commits:

1. **`refactor(kimi3): extract the K3 communication layer into kimi_k3_comm`** — pure refactor, behavior-equivalent to main. K3's attention-AR strategy dispatch and the MoE-tail tier table (TAIL_FUSION / MULTIMEM_AR / FUSED_LANE_AR / SEPARATE_REDUCE) move behind `K3AttnComm` / `K3MoeTailComm` with one MIN-vote per process instead of one per layer; `kimi_k3.py` states semantics only. Tier bodies are taken verbatim from current main (post-#1038 sharded variants, post-#1102 workspace pooling, post-#1060 ctx plumbing). Also folds in singleton-state parameter guards and a ctx-forwarding fix in `_forward_fused_attnres_graph` (pre-existing gap: the cross-DP-EP token gather requires a ForwardContext).

2. **`perf(kimi3): fold the MoE finalize into the multicast latent tail`** — a `FINALIZE_INPUT` variant of the tail's publish phase consumes the experts kernel's deferred-finalize triple (permuted gemm2 rows, bf16 expert scales, expanded-to-permuted indices with `-1` padding) and performs gather + fp32 ascending-k weighted accumulation + a single bf16 round in-register before the multicast store. One fewer kernel launch and no `[M, latent]` round-trip per MoE layer per step. Pure local CuTe-DSL: no new dependencies, no env switches, no topology limits beyond the tail's own; capability-negotiated with automatic fallback to the materialized-finalize path (Marlin / native / eager unaffected).

## Validation (4x GB200, TP16 cross-node NVLS)

- Numerical equivalence suite: 18 cases x 2 PDL modes (default and `--disable-pdl`), against an independent fp32 torch reference, with NaN-poisoned padding rows to prove the `-1`-skip logic never reads a dropped slot.
- Tier truth table: 19 cases.
- E2E serving: negotiation log shows `deferred_finalize=True` engaged; GSM8K within the established baseline band.
- Two rounds of multi-lens adversarial review; all confirmed findings addressed.

## Known items

- Upstream (unrelated, also fails on main): `test_latent_tail_accepts_sharded_weight[1]` fails at world=16; will be reported separately.
- Decode-TPOT gain measurement on 8x B300 is follow-up.
- pre-commit hooks could not run in the validation environment (config pins python3.12); black/isort of the same major version report clean.

## Note

This replaces the earlier MNNVL CuTe-DSL finalize-tail approach on this branch: the in-tail fusion covers the same work without the flashinfer-nightly dependency, the world==8 restriction, or a new tier.

### DIFF
--- a/python/tokenspeed/runtime/layers/moe/latent.py
+++ b/python/tokenspeed/runtime/layers/moe/latent.py
@@ -24,7 +24,6 @@ from __future__ import annotations
 
 from collections.abc import Callable
 from dataclasses import dataclass, replace
-from enum import IntEnum
 from functools import partial
 
 import tokenspeed_kernel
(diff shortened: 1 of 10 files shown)


nperrin-fr:
~~how about we merge to #957 first~~

PR #1080, opened 2026-08-13
feat(spec): frontier-anchored multi-layered MTP

syuoni:
## Summary

Makes the frontier-anchored decode window the only multi-depth MTP drafting
mechanism and cleans the path end to end.

### Core change

Every draft depth `0..steps-1` runs the **same k rows per request, ending at
the committed frontier** (`frontier = vc + accept`). Rows re-run committed
positions, so KV/sconv slots written last round from since-rejected drafts are
rewritten in place with exact values — no repair passes, no per-depth metadata
advance. All round geometry (positions, cache write locations, seq_lens,
sample gather) is `frontier ± const`, recomputed per CUDA-graph replay from
`accept` alone with pure tensor math; no host syncs.

Supporting pieces:
- **Cross-round drafter stash** (last k−1 committed tokens + target hiddens,
  keyed by req pool slot) reconstructs the window's pre-verify history; it is
  the tail `[:, 1:]` of the depth-0 window, so the roll is two slice copies.
- **Rolling per-depth ids**: depth d+1 = depth d's window shifted one left
  with its draft appended (one `cat`); only depth 0 is accept-dependent.
- **Backend hook** `update_draft_forward_metadata(frontier)` re-anchors conv
  metadata and grouped write locs (the per-round absolute counterpart of
  `advance_draft_forward_metadata`).
- The superseded designs (verify-anchored window, lookback variant) and their
  configure/arm handshakes are removed; EXTEND rounds always prime the stash.

### Cleanups riding along

- MM pad substitution hoisted from the drafters to `fill_input_buffers`
  (once per chunk on the drafter-only shift-1 buffer): drafters only ever see
  in-vocab ids; per-depth substitute+clamp deleted from MTP and EAGLE;
  unconfigured multimodal targets now fail loudly at startup.
- Never-binding defenses removed (accept-length clamps from #574 — their
  producer died in #974 — redundant casts/slices, dead `is_decode` conv
  metadata field, dead `bs` params, single-use locals).
- Token-value tensors unified to int32; docstrings/comments trimmed to
  constraints the code can't state.

### DIFF
--- a/python/tokenspeed/runtime/configs/inkling_config.py
+++ b/python/tokenspeed/runtime/configs/inkling_config.py
@@ -38,7 +38,6 @@ from __future__ import annotations
 
 import copy
 import enum
-import os
 from typing import Any, Literal
 
 from transformers.configuration_utils import PretrainedConfig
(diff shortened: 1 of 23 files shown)


syuoni:
@lightseek-bot   Looks like `tokenspeed-smg==1.9.0.post20260814` is broken in aarch64 platform: 
* https://github.com/lightseekorg/tokenspeed/actions/runs/31802542102/job/94773720695?pr=1080
* https://github.com/lightseekorg/tokenspeed/actions/runs/31802542102/job/94773720587?pr=1080

Locally hit the same error, downgrading to `tokenspeed-smg==1.9.0.post20260813` can WAR.

Failures are unrelated to this PR. Merging.

PR #1087, opened 2026-08-13
refactor(runtime): split the KDA attention backend into a subclass

nperrin-fr:
MambaAttnBackend served both the GDN/mamba and the KDA kernel families, discriminated by an ``is_kda`` constructor flag threaded through the two big forward methods. Hoist the family-divergent regions of ``forward_decode`` / ``forward_extend`` into five named seam methods and move the KDA implementations to a new ``KdaAttnBackend`` subclass, so the shared bodies read as one linear flow with hook points and the base class no longer knows KDA exists.

No behavior change: same kernels, launch order, buffers and messages. Verified bitwise against the pre-split module on target verify (fused and recurrent), prefill and single-token decode for both families.


nperrin-fr:
https://github.com/lightseekorg/tokenspeed/actions/runs/31743163028

PR #1108, opened 2026-08-15
refactor(pd): clarify cache transfer contract ownership

OftenDream:
Centralize PD topology and recipe-owned transfer semantics so cache-block manifests remain model-neutral across transfer modes.

## Summary

- Introduce typed `PDParallelTopology` as the single Prefill/Decode TP/CP/DP owner, and fail fast when CachePD is asked for unsupported CP.
- Move model-specific field partition / transfer schema into the cache recipe layer, so `cache_protocol` stays a model-neutral wire and validation envelope.
- Rename request manifests from page-centric naming to CacheBlock terminology (`block_ids`), so history KV and state/checkpoint groups share one generic PD contract across transfer modes.

### DIFF
--- a/python/tokenspeed/runtime/pd/topology.py
+++ b/python/tokenspeed/runtime/pd/topology.py
@@ -1,0 +1,108 @@
+# Copyright (c) 2026 LightSeek Foundation
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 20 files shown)


python/tokenspeed/runtime/pd/topology.py
chenht2022, line 1: Is there a particular reason for introducing _AttentionParallelMapping, _PDMapping, and PDParallelTopology here instead of reusing AttentionLayerMapping directly? They seem to contain largely the same topology information.

PR #1121, opened 2026-08-17
ci(kimi-k3): exercise DSpark in the AIME26 gate

raikonenfnu:
## Problem

AMD per-commit CI currently exercises only the speculator-free Kimi-K3 AIME26 configuration, while enabling the existing DSpark gate exposed an Iris cache bug: ordinary all-reduce could create a 512 KiB state under the same key later used by the 1 MiB producer-direct path, causing CUDA graph capture to fail with `Iris symmetric outputs exceed the input buffer`.

## Key changes

- Promote `test/ci/eval/kimi-k3-dspark-mxfp4-tp8ep8-evalscope-aime26-amd.yaml` to `per-commit`, making Kimi-K3 + DSpark the regular AMD AIME26 correctness gate.
- Move `test/ci/eval/kimi-k3-mxfp4-tp8ep8-evalscope-aime26-amd.yaml` to manual-only, retaining the speculator-free configuration as a comparable control without duplicating the expensive per-commit run.
- In `tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/triton.py`, create every cached Iris state with the larger of the ordinary all-reduce capacity and the advertised byte-backed producer-direct capacity, so whichever operation initializes the shared cache first can safely serve both paths while legacy states with `max_bytes=0` retain their original capacity.
- In `tokenspeed-kernel/test/ops/test_iris_communication.py`, cover both the enlarged producer backing and the legacy element-only fallback, including cache reuse.

## Validation

- Local 8x MI350X launch with the CI TP8/EP8 K3 + DSpark server arguments captured decode graphs for batches `1, 2, 4, 8, 16`, completed warmup and a real request, and exercised DSpark with an observed accept length of 4.2.
- Exact CI EvalScope AIME26 command: `0.9333` (28/30), clearing the `0.90` gate; 11m43s evaluation time, 42.8 ms average TPOT.
- `python -m pytest -q tokenspeed-kernel/test/ops/test_iris_communication.py` (`37 passed`)
- `python -m pytest -q test/ci_system/test_pipeline.py` (`78 passed`)

### DIFF
--- a/test/ci/eval/kimi-k3-dspark-mxfp4-tp8ep8-evalscope-aime26-amd.yaml
+++ b/test/ci/eval/kimi-k3-dspark-mxfp4-tp8ep8-evalscope-aime26-amd.yaml
@@ -1,11 +1,11 @@
 api_version: ci.tokenspeed.io/v1
-# Kimi-K3 + DSpark on the same AIME26 gate as the speculator-free run, so the
-# two scores are directly comparable. Separate job from
-# eval-kimi-k3-mxfp4-tp8ep8-aime26-amd, which tracks the pure target.
+# Kimi-K3 + DSpark is the per-commit AIME26 gate. The matching speculator-free
+# configuration remains available as a manual control for direct comparison.
 name: eval-kimi-k3-dspark-mxfp4-tp8ep8-aime26-amd
 type: eval
 workflow_stage: model-test
 triggers:
+  - per-commit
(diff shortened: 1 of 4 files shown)


efric (dismissed):
LGTM! ran into the same observation elsewhere; thanks

efric (approved):

PR #1129, opened 2026-08-18
perf(kimi3): avoid packed KDA QKV copies

qedawkins:
Keep the merged KDA projection output packed and pass its QKV view directly to the stride-aware fused decode kernel, eliminating one device copy per KDA layer for multi-request decode. Materialize compact QKV only for multi-token and fallback paths that require it.

- Cover packed projection aliasing and numerical parity for single- and multi-row outputs.
- Exercise gfx950 fused KDA decode with packed row strides at graph batch sizes 1, 2, 4, 8, 16, and 32.

Benchmark (8x AMD Instinct MI355X, Kimi-K3 TP8/EP8, FP8 KV): EvalScope random workload with fixed 4096-token inputs and 1024-token outputs. Graph-mode decode without speculative decoding
- B1/C1: 71.03 -> 70.93 output tok/s (-0.1%)
- B2/C2: 88.14 -> 89.22 output tok/s (+1.2%)
- B4/C4: 164.07 -> 165.60 output tok/s (+0.9%)
- B8/C8: 240.53 -> 242.74 output tok/s (+0.9%)
- B16/C16: 344.29 -> 346.79 output tok/s (+0.7%)

### DIFF
--- a/python/tokenspeed/runtime/layers/attention/backends/hybrid_linear_attn.py
+++ b/python/tokenspeed/runtime/layers/attention/backends/hybrid_linear_attn.py
@@ -1397,6 +1397,9 @@ class MambaAttnBackend(AttentionBackend):
         if fused_out is not None:
             return fused_out
 
+        # Stride-aware fused decoders consume packed projection views directly.
+        # Preserve the shared fallback's established compact input layout.
+        mixed_qkv = mixed_qkv.contiguous()
         mixed_qkv = causal_conv1d_update(
             mixed_qkv,
             conv_states,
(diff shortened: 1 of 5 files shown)


Yu-Zhewen (approved):
Not for this PR, but could we make all KDA kernel variants support strided `mixed_qkv` to simplify runtime `.contiguous()` handling?

qedawkins:
> Not for this PR, but could we make all KDA kernel variants support strided `mixed_qkv` to simplify runtime `.contiguous()` handling?

Makes sense to me. I don't think that extends only to KDA, we should try to avoid the copies that come with `.contiguous()` everywhere.

qedawkins:
@lightseek-bot can you please merge this

PR #1133, opened 2026-08-19
perf(kda): consume strided decode projections

panditsa:
## Summary

- Let gfx950 KDA decode consume token-strided Q/K/V/gate projections directly.
- Let the gated RMSNorm epilogue consume a token-strided output gate.
- Remove the corresponding model-side contiguous copies.

### Current revisions

- Base: `978ed2cf` (current `upstream/main`)
- Head: `0e1c7a72` (merge of current main into the PR branch)

### Correctness and copy removal

- Targeted current-head fused/strided KDA coverage: 8 passed, 16 deselected.
- Parent and head isolated output checksums match exactly.
- Parent profile: one logical `contiguous -> clone -> copy_` sequence.
- Head profile: no contiguous/clone/copy event.

### Isolated KDA performance

Production fused KDA decode at M=2 under CUDA-graph replay, median of seven trials:

| Revision | Median latency | Change |
|---|---:|---:|
| Parent | 23.147 µs | |
| Head | 19.550 µs | -15.5% |

This removes 3.597 µs per KDA layer. Across 69 KDA layers, that predicts about 0.248 ms saved per decode step.

### Current end-to-end performance

MI350X, TP8/EP8, M=2, 4K input / 1K output, CUDA graph capture size 2, median of three:

| Metric | Parent | Head | Change |
|---|---:|---:|---:|
| Output throughput | 82.06 tok/s | 82.05 tok/s | -0.01% |
| TPOT | 23.663 ms | 23.672 ms | +0.04% |

The local copy removal remains measurable and correct, but after merging current main its effect is no longer distinguishable in the full-model benchmark. All 12 measured requests per boundary completed with valid lengths.

## Stack

- Base: current `main`

### DIFF
--- a/python/tokenspeed/runtime/models/kimi_k3.py
+++ b/python/tokenspeed/runtime/models/kimi_k3.py
@@ -1257,7 +1257,7 @@ class KimiLinearKDA(nn.Module):
             # per-head norm implementation.
             core_out = rmsnorm_gated_sigmoid(
                 core_out.contiguous(),
-                out_gate.contiguous(),
+                out_gate,
                 self.o_norm.weight,
                 self.o_norm.variance_epsilon,
                 hn,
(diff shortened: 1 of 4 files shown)


qedawkins (approved):
LGTM, some overlap with this though: https://github.com/lightseekorg/tokenspeed/pull/1129

raikonenfnu (approved):

PR #1138, opened 2026-08-19
feat(amd): add gfx950 A8W4 SiTU core

panditsa:
Summary

- add dynamic FP8 activation quantization for the TP-sharded A8W4 path
- add route-direct SiTU W13/W2 decode with optional shared-down fusion
- use the package-prefill path above the small-M regime

Performance

Direct TP8/EP1 K3-shaped CUDA-graph microbenchmark (`E=896, H=3584, I/TP=384, topk=16`):

| M | Existing grouped A16W4 | New A8W4 core | Reduction |
|---:|---:|---:|---:|
| 1 | 84.80 µs | 26.51 µs | 68.7% |
| 2 | 85.53 µs | 27.32 µs | 68.1% |
| 4 | 85.12 µs | 30.26 µs | 64.4% |
| 8 | 89.84 µs | 39.20 µs | 56.4% |
| 16 | 157.37 µs | 65.03 µs | 58.7% |

This PR exposes the core but does not register it with the model-facing selector, so there is intentionally no standalone end-to-end claim here. Registration and E2E selection land in #1139.

Tests

- direct A8W4 core validated in eager mode and CUDA-graph replay for M=1/2/4/8/16
- explicit output-buffer reuse verified at every tested M
- numerical reference and selector coverage are provided by #1139

Stack

- depends on #1137


tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/moe/mxfp4/fused/_layouts.py
```diff
@@ -343,3 +362,96 @@ def _moe_partial_reduce(
...
+            ).to(gl.float32)
+            shared_input = gl.convert_layout(shared_input[None, :], shared_layout)
+            shared_acc += gl.sum(shared_weight * shared_input, axis=1)
+        gl.store(
+            SharedOut + shared_token * stride_som + shared_n * stride_son,
+            shared_acc.to(SharedOut.dtype.element_ty),
+        )
+
+
+def _route_small_m(logits, topk, dtype):
```
antiagainst, line 454: What's this for? Is it to be used later?

antiagainst (approved):

PR #1141, opened 2026-08-19
perf(comm): add two-stage producer-direct reduction

panditsa:
Summary

- keep one-stage Iris reduction for small producer-direct payloads
- switch world-size 4/8 payloads above the dtype-aware threshold to reduce-scatter plus all-gather
- retain BF16, FP16, and FP32 packing behind the existing symmetric-output API

Performance

Exact parent `2ebd8f81` → head `60babeb9`, world-size 8 BF16 producer-direct reduction under CUDA-graph replay, critical-rank p50:

| Payload | M | Parent | Head | Reduction |
|---|---:|---:|---:|---:|
| AttnRes [M,7168] | 2 | 9.26 µs | 7.96 µs | 14.0% |
| AttnRes [M,7168] | 4 | 10.68 µs | 8.70 µs | 18.6% |
| AttnRes [M,7168] | 8 | 14.52 µs | 10.25 µs | 29.4% |
| AttnRes [M,7168] | 16 | 21.04 µs | 10.47 µs | 50.2% |
| Joint MoE [M,3584]+[M,7168] | 2 | 9.31 µs | 8.30 µs | 10.8% |
| Joint MoE [M,3584]+[M,7168] | 4 | 11.93 µs | 9.68 µs | 18.9% |
| Joint MoE [M,3584]+[M,7168] | 8 | 16.88 µs | 10.41 µs | 38.3% |
| Joint MoE [M,3584]+[M,7168] | 16 | 26.99 µs | 10.75 µs | 60.2% |

M=2/4 remain on the one-stage path; M=8/16 cross the world-size-8 BF16 two-stage threshold.

Tests

- threshold, world-size 4/8, noncontiguous subgroup, large-payload, dtype, and repeated graph-replay coverage: 7 passed
- benchmark validates exact reduced values before timing each eager/captured operation

Stack

- depends on #1140

Issue #1132, opened 2026-08-18
[gfx950][k3][dspark] Improve long context decoed cost on low acceptance


raikonenfnu:
# Summary

On 8×MI350X, Kimi-K3 DSpark improves 4k/1k TP8/EP8 throughput by 39–106%, but at 131,072-token input its average accepted length collapses to exactly `1.0`, turning the extra 7 draft positions plus 8-wide target verification into overhead and making decode 28–41% slower than ordinary TokenSpeed.

The first-principles performance target is: **make one DSpark verification round a first-class vectorized M=8 execution path whose cost approaches one ordinary decode round, independent of how many candidates are eventually accepted.** Acceptance determines how many output tokens amortize that round, but inefficient infrastructure currently inflates the numerator by 2.39×; that underlying cost should be fixed before relying on an acceptance-based fallback.

## Reproduction contract

- TokenSpeed: `8911ce4cd7bd9f4fcea33bc069a5d0f27cebfa95`
- Hardware: 8×AMD Instinct MI350X (`gfx950`)
- Model: Kimi-K3 DSpark checkpoint
- Topology: TP8/EP8
- Runtime: FP8 KV cache, graph-compiled decode, eager/chunked prefill, greedy sampling
- Workload: identical random-token prompts, exact ISL/OSL, prefix caching explicitly disabled
- “132K” below means ISL=131,072 and OSL=1,024 (132,096 total tokens/request)

## Current end-to-end behavior

| Shape | Non-DSpark output tok/s | DSpark output tok/s | Delta | Non-DSpark TPOT | DSpark TPOT | DSpark / base TPOT |
|---|---:|---:|---:|---:|---:|---:|
| 4K/1K B1/C1 | 66.30 | **136.28** | **+105.6%** | 14.62 ms | **6.83 ms** | 0.47× |
| 4K/1K B8/C8 | 223.59 | **348.55** | **+55.9%** | 32.97 ms | **18.74 ms** | 0.57× |
| 4K/1K B16/C16 | 331.84 | **459.90** | **+38.6%** | 42.76 ms | **27.64 ms** | 0.65× |
| 132K/1K B1/C1 | **30.45** | 17.88 | **-41.3%** | **18.04 ms** | 40.69 ms | **2.26×** |
| 132K/1K B2/C2 | **36.55** | 23.65 | **-35.3%** | **24.96 ms** | 54.03 ms | **2.16×** |
[...]

PR #1146, opened 2026-08-19
(WIP) perf(kimi3): enable batched AttnRes runtime fusion

panditsa:
Summary

- allow the backend-selected fused AttnRes combine for supported multi-token decode batches
- hoist next-layer MLP/attention partials whenever the projection kernel reports support
- join the producer stream before gfx950 M<16 reductions that consume AttnRes scratch

Validation result

This WIP is not ready to land: activating the batched fusion regresses end-to-end decode even though the isolated kernel in #1142 is faster.

4K/1K TP8/EP1 CUDA graphs, median of three:

| M | Equivalent parent tok/s | Head tok/s | Change | Parent TPOT | Head TPOT |
|---:|---:|---:|---:|---:|---:|
| 2 | 103.64 | 93.99 | -9.3% | 18.621 ms | 20.308 ms |
| 4 | 185.64 | 172.06 | -7.3% | 20.380 ms | 21.459 ms |

The parent figures are the exact #1144 head. #1145 only selects its sharded tail at M>=8, so M=2/4 are unchanged at the #1146 parent boundary.

The likely issue is orchestration rather than fused-kernel latency: joining the producer stream and consuming AttnRes scratch removes overlap that the prior split path retained. Runtime activation should remain gated until the dependency/stream schedule is redesigned and reprofiled.

Tests

- batched runtime-dispatch test passes and confirms the fused Iris combine is selected when supported
- #1142 covers world-size-8 M=1/2/4/8/16 numerical correctness and CUDA graphs
- all 18 measured head requests completed with valid lengths

Stack

- depends on #1145

WIP blocker

- restore the runtime gate or recover the lost producer/consumer overlap before landing

### DIFF
--- a/python/tokenspeed/runtime/models/kimi_k3.py
+++ b/python/tokenspeed/runtime/models/kimi_k3.py
@@ -2316,10 +2316,11 @@ class KimiLinearDecoderLayer(nn.Module):
             else None
         )
         attnres_partial_args = None
-        if next_mix is not None and self._hoist_next_mlp and num_tokens == 1:
+        if next_mix is not None and self._hoist_next_mlp:
             next_layer, _ = next_mix
             # This layer's attention projection writes both block partials
             # hoisted for the next layer, which consumes their scratch slots.
+            # Kernel availability below is the token-count capability gate.
             candidate_args = (
                 block_residual[:mlp_valid_blocks],
                 next_layer._mlp_wp,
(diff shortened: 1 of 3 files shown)


jhinpan:
The exact #1132 M=8/EP8 point differs from the published M=2/4 EP1 blocker. On current main (`9196fa27`), #1141 + #1142 + this runtime activation gives:

- forced 131072/1024 TP8/EP8: 36.25 ms TPOT and 19.9703 tok/s (3 runs), versus 36.80 ms / 19.7647 baseline: -1.49% TPOT, +1.04% throughput
- normal 4K/1K DSpark: 155.078 tok/s versus 152.949 baseline (+1.39%)
- exact token identity versus ordinary greedy passes at 4K/1K and 131K/128 under forced rejection

So the activation is beneficial at M=8/EP8, but still regresses the already-reported M=2/4 EP1 points and only moves the #1132 round ratio from 2.157x to ~2.125x. This supports redesigning/gating the orchestration by shape/topology rather than landing the current activation globally.

PR #1153, opened 2026-08-19
feat(loader): add the InstantTensor weight loader, opt-in only

nperrin-fr:
(the motivation is to accelerate local dev)

InstantTensor reads safetensors shards directly onto the GPU, sharding the reads across ranks. On MiniMax-M3 over four GPUs it cuts the weight-load phase from 34-43s to a steady 21s.

This re-lands PR #418, which was reverted for CI failures. The revert was correct: the original also switched CI jobs and image builds over to the new format. Only the loader comes back here, reachable solely through an explicit `--load-format instanttensor`; no default and no job changes.

It is an optional extra rather than a core dependency, because the project publishes x86_64 wheels only -- as a core dependency it would break `pip install` on ARM, including this repo's own ARM CI, whether or not the format is ever requested.

Two things the loader's GPU-direct behavior exposes elsewhere:

A model's `load_weights` must consume the iterator lazily. Buffering it is free when tensors stage through host RAM but holds the whole checkpoint on the device here, so gpt-oss's MXFP4 path streams its expert tensors instead of collecting them into a list. That in turn removes the whole-iterator `any()` probe it used to pick an expert layout -- a probe would consume the stream -- so each tensor now routes on its own name, which is unambiguous either way and avoids silently dropping every tensor of the layout that lost the latch.

Sub-byte safetensors dtypes are rejected up front. InstantTensor applies a header's logical element count to a torch dtype whose element is a packed pair, so an `F4` tensor declared [2048] over 1024 bytes arrives at shape (2048,) -- twice its own storage -- and nothing raises. NVFP4 and MXFP4 checkpoints are unaffected: they store packed 4-bit values as byte-aligned `U8`, and load bit-identically to the safetensors loader (verified over 321 tensors of MiniMax-M3-NVFP4).

### DIFF
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -37,7 +37,8 @@ export default defineConfig({
         text: "Guides",
         items: [
           { text: "Getting Started", link: "/guides/getting-started" },
-          { text: "Launching a Server", link: "/guides/launching" }
+          { text: "Launching a Server", link: "/guides/launching" },
+          { text: "InstantTensor Loading", link: "/guides/instanttensor" }
         ]
       },
       {
--- a/docs/configuration/server.md
+++ b/docs/configuration/server.md
@@ -16,7 +16,7 @@ For a compact compatibility table, see
 | `--tokenizer` | Tokenizer path when it differs from the model path. |
 | `--tokenizer-mode` | Select tokenizer behavior. `auto` uses fast tokenizers and model-specific hooks when available. |
 | `--skip-tokenizer-init` | Skip tokenizer initialization for input-ID-only serving paths. |
-| `--load-format` | Weight loading format: `auto`, `pt`, `safetensors`, `npcache`, `dummy`, or `extensible`. |
+| `--load-format` | Weight loading format: `auto`, `pt`, `safetensors`, `instanttensor`, `npcache`, `dummy`, or `extensible`. See [InstantTensor](/guides/instanttensor) for the accelerated NVIDIA loader. |
 | `--trust-remote-code` | Allow custom model code from the model repository. |
 | `--revision` | Model branch, tag, or commit. |
 | `--download-dir` | Hugging Face download/cache directory. |
--- a/docs/guides/instanttensor.md
+++ b/docs/guides/instanttensor.md
@@ -12,11 +12,25 @@ default safetensors loader, so model accuracy is unaffected.
 
 ## Installation
 
-InstantTensor is an optional extra, because it publishes x86_64 wheels only:
+InstantTensor is an optional extra, because PyPI carries x86_64 wheels only:
 
 ```bash
 pip install "tokenspeed[instanttensor]"
 ```
+
+On aarch64 hosts (GB200, GB300) there is no published wheel, but the sdist
+builds from source in about a minute. The extension reaches CUDA, cuFile, and
+NCCL through `dlopen` and vendors its own Boost, libaio, and liburing, so it
+needs a C++ toolchain and `make` -- not `nvcc`, and not a matching CUDA
+toolkit:
+
+```bash
+pip install --no-binary instanttensor "tokenspeed[instanttensor]"
+```
+
+This is why the dependency is an extra rather than a core one: as a core
+dependency every ARM install would compile it, whether or not the format is
+ever requested.
 
 It is imported lazily, so it is loaded only when you select
 `--load-format instanttensor`.
@@ -37,6 +51,19 @@ tokenspeed serve deepseek-ai/DeepSeek-R1 \
   --tensor-parallel-size 8 \
   --enable-expert-parallel
 ```
+
+## Measured effect
+
+Weight-load phase for Kimi-K3 (1.42 TiB over 96 shards, 497k tensors) on two
+GB300 nodes with `--tensor-parallel-size 8`, checkpoint on shared storage:
+
+| `--load-format` | weight-load phase |
+| --- | --- |
+| `auto` (safetensors) | 869-1106 s |
+| `instanttensor` | 231-234 s |
+
+The ranges are run-to-run variance on the shared filesystem. Both formats
+serve identical output, as the loaders are bit-for-bit equivalent.
 
 ## Memory considerations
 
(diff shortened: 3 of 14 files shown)


nperrin-fr:
## B200 validation: 9 models, 4 speculative-decoding configs

Ran the loader across every checkpoint that fits on one 8×B200 node, in two stages:

- **Bit-identity** — iterate `instanttensor_weights_iterator` and `safetensors_weights_iterator` over **every shard** of each checkpoint and compare dtype, shape and contents tensor by tensor. No engine involved.
- **End to end** — serve the model with `--load-format instanttensor` and answer three fixed prompts (`17*23`, `3x+7=25`, capital of Japan) at `temperature=0`.

### Bit-identity: 1,084,387 tensors, zero differences

| model | shards | tensors | differing |
| --- | ---: | ---: | ---: |
| gpt-oss-20b (MXFP4) | 3/3 | 459 | 0 |
| gpt-oss-120b (MXFP4) | 15/15 | 687 | 0 |
| Inkling-Small-NVFP4 | 10/10 | 1,360 | 0 |
| Inkling-NVFP4 | 34/34 | 2,056 | 0 |
| MiniMax-M3-NVFP4 | 88/88 | 89,614 | 0 |
| Qwen3.5-122B-A10B-NVFP4 | 9/9 | 149,309 | 0 |
| MiniMax-M2.7-NVFP4 | 15/15 | 191,087 | 0 |
| Kimi-K2.5-NVFP4 | 119/119 | 278,341 | 0 |
| Qwen3.5-397B-A17B-NVFP4 | 11/11 | 371,474 | 0 |
| **total** | **304** | **1,084,387** | **0** |

### End to end: every config serves and answers correctly

| config | TP | serve → ready | answers |
| --- | ---: | ---: | --- |
| gpt-oss-20b `--load-format auto` | 1 | 231 s | ✅ |
| gpt-oss-20b `instanttensor` | 1 | 230 s | ✅ |
| gpt-oss-120b `--load-format auto` | 4 | 290 s | ✅ |
| gpt-oss-120b `instanttensor` | 4 | 320 s | ✅ |
| Qwen3.5-122B-A10B-NVFP4 | 4 | 350 s | ✅ |
| MiniMax-M2.7-NVFP4 | 4 | 350 s | ✅ |
| Inkling-Small-NVFP4 | 4 | 350 s | ✅ |
| MiniMax-M3-NVFP4 | 4 | 450 s | ✅ |
| Qwen3.5-397B-A17B-NVFP4 | 8 | 330 s | ✅ |
| Kimi-K2.5-NVFP4 | 8 | 350 s | ✅ |
| Inkling-NVFP4 | 8 | 321 s | ✅ |
| MiniMax-M3-NVFP4 + EAGLE3 | 4 | 421 s | ✅ |
| MiniMax-M2.7-NVFP4 + EAGLE3 | 4 | 340 s | ✅ |
| Kimi-K2.5-NVFP4 + EAGLE3 | 8 | 310 s | ✅ |
| Inkling-NVFP4 + MTP | 8 | 320 s | ✅ |
| Qwen3.5-397B-A17B-NVFP4 + MTP | 8 | 320 s | ✅ |

Both gpt-oss arms were run because this PR rewrites that model's `load_weights`. Their answers are **character-for-character identical** between `auto` and `instanttensor` at both sizes. The other models rely on the bit-identity stage for equivalence and were run on the new loader only.

`serve → ready` is dominated by kernel autotuning (~250–320 s of it) and gateway warmup, both loader-independent; weight loading itself is a small share. It is reported for completeness, not as a loader benchmark.

### One real bug found, fixed in 976b515c5

MiniMax-M2.7 + EAGLE3 could not start under the new loader:

```
RuntimeError: Expected all tensors to be on the same device, but found at least
two devices, cuda:1 and cpu!
  File ".../models/llama_eagle3.py", line 692, in load_weights
```

`self.hot_token_id = loaded_weight + torch.arange(n)` builds the arange on the CPU. Every CPU-staging loader also hands over a CPU `loaded_weight`, so the addition happened to work and the assumption was never exercised — a GPU-direct loader yields the weight already on the device and the line raises. **Before the fix, `--load-format instanttensor` could not serve any EAGLE3 draft at all.**

An AST scan for tensor construction without `device=` inside any `load_weights` found exactly two sites, both the same `d2t` vocabulary map: `llama_eagle3.py:692` and `deepseek_v3.py:2322`. The DeepSeek one is not reachable from this model but carries the identical defect and is fixed too. A follow-up scan for other ways a CPU tensor enters `load_weights` (`from_numpy`, `as_tensor`, `.cpu()`, `torch.cat` across sources) turned up only two `torch.cat` calls whose operands both come from the same loader-produced cache — not a hazard.

The regression test is mutation-checked: removing `device=` from either site turns both cases red.

### Not covered

- **LongCat-2.0** — 3.3 TiB against 1,466 GiB of node memory. Needs multiple nodes.
- **Kimi-K3** — verified separately on GB300 (1.42 TiB, 96 shards, 497k tensors), not in this run.

### Failures that were environment or harness, not the loader

Recorded so the run is auditable rather than implying a clean first pass:

- `kvstore` is on by default (`--disable-kvstore` unset ⇒ enabled), and at `ratio=2.0` tried to pin ~259 GB of host memory. Fixed by capping `--kvstore-size`.
- Speculative decoding plus `kvstore` fails the L2 transfer-layout check, because a draft that shares layer numbering with the target collides on KV fields — `layer.60.k/v` for MTP, `layer.61.latent_kv` for EAGLE3-MLA. All spec configs above were run with `--disable-kvstore`, matching the repo's own CI specs for MTP. Pre-existing and unrelated to this PR.
- Kimi-K2.5 needs `--attention-backend tokenspeed_mla` and enough per-GPU headroom for the MLA cache budget (TP8, not TP4, at 550 GiB).
- `tokenspeed_mla` cannot supply paged cache metadata for the prefill-graph dummy batch, so K2.5 logs a warning and falls back to eager prefill. The engine labels this a known gap; serving is unaffected.
- Harness defects of my own: a probe that did not use the gateway's model id, leaked server process groups holding GPUs and port 8400, and a readiness check that treated a warning-level traceback as a startup failure.

PR #1156, opened 2026-08-20
feat(kda): eager replay commit for speculative target verify

nperrin-fr:
Replace the per-draft-position state scratch with an eager replay: verify captures the compact projection payload (qkv / f_a / beta rows) and runs non-destructively from the committed pages -- the fused paged verify gains a no-store mode so no state tape is written -- and commit replays only the accepted prefix from the committed state, inside the same iteration. No state ever survives a round boundary, so no deferred-window lifecycle, no ownership inference, and no engine-side hooks exist (contrast PR #1058; design per its review discussion).

The commit is a batched two-launch sequence over all 69 KDA layers (one all-layer gate precompute, one all-layer conv/recurrent commit) driven by a device descriptor table built once at startup, and it is captured INLINE in the one-round decode graph (verify -> accept -> commit -> draft), deliberately sequential -- its cost stays visible in the graph timeline (38 us stream time for all layers at bs1). Per-round host cost is zero. Eager execution launches the same batched path directly; platforms without the replay kernel keep the per-layer loop, then the scratch path.

Measured on 4-node TP16 Kimi-K3 (EAGLE3, bs1, paired arms in one allocation):

  main   978ed2cf: KV 29.75 GiB, act+graphs 5.27 GiB, 186.9 tok/s
  eager  d478429b: KV 33.87 GiB, act+graphs 1.18 GiB, 203.1 tok/s

i.e. +4.1 GiB/rank of KV (the verify tape eliminated; matches the planner-vs-allocator byte accounting exactly) and +8.7% bs1 decode throughput (tape-write bandwidth out of the graph; commit host time 20.5 ms -> 0). Numerics: committed conv is bit-identical to the scratch path; gate/recurrent fp32 agree to 1e-6 abs (reduction-order shift in the batched gate); multi-round equivalence, CUDA-graph replay, padded lanes, and the double-commit guard are each pinned by tests with negative controls.


syuoni (approved):
I did some cleanup -- this PR is ready to merge.

We should revisit the whole design of replay-SMM implementation, unifying different models and platforms and co-designing with the scheduler.

Currently the implementation is a bit ad-hoc, while I expect this feature should be well-scoped.

@nperrin-fr @zhyncs @tuanzhangCS

PR #1162, opened 2026-08-20
feat(mla): add decode context parallelism

qimcis:
## Summary

Adding decode context parallelism (DCP), building on top of the fp8 MLA kernel introduced in #364 , i subdivide each attention tensor-parallel group and cyclically place the sequence history across its ranks. tested and implemented with dsv4 model family and kimi k3 in mind

```text
Logical KV history
  ↕ owner(position) = position % DCP, local_row = position // DCP
Rank-local MLA history (1 / DCP rows)
  ↕ query-head all-gather
Local decode over 1 / DCP context
  ↕ exact LSE merge
Original TP-local output heads
```

for sequence-sharded MLA history, the per-rank physical storage is:

```text
DCP1: 100.0% of baseline  (0.0% saved)
DCP2:  50.0% of baseline (50.0% saved)
DCP4:  25.0% of baseline (75.0% saved)
DCP8:  12.5% of baseline (87.5% saved)
```

thought, it's importnant that we note that this is the saving for the eligible dcp history, not necessarily the entire model cache. dsv4 for example will keep SWA, the FP4 sparse indexer, and compressor state sequence-replicated, and models like kimi k3 keep kda state sequence-replicated because it is already partitioned by TP heads. 

When testing with dsv4-flash with a fixed global 8,192-token capacity, dcp2 halved the compressed history rows and reduced the complete rank-local arena from 1,574,012,160 to 1,513,935,360 bytes (3.82%, about 57.3 MiB per rank).

## Test Plan

dsv4 flash official checkpoint (`deepseek-ai/DeepSeek-V4-Flash`) on 2×B200:

- Baseline TP2, TP2/DCP2 `ag_rs`, and TP2/DCP2 `a2a` end-to-end serving.
- CUDA graphs at batch sizes 1 and 2; plumbing coverage also exercised sizes 1, 2, 4, 7, and 8.
- 30K-token chunked prefill, prefix-cache replay, and two concurrent requests.
- Matched 30K prompt / 128-token decode benchmark:
  - Baseline TP2: 76.018 tok/s; 334.765 ms short-context median.
  - DCP2 AG/RS: 55.464 tok/s; 426.305 ms short-context median.
  - DCP2 A2A: 65.117 tok/s; 378.743 ms short-context median.

MODEL=deepseek-ai/DeepSeek-V4-Flash
REVISION=60d8d70770c6776ff598c94bb586a859a38244f1

```
COMMON_ARGS=(
  --revision "$REVISION"
  --tensor-parallel-size 2
  --enable-expert-parallel
  --kv-cache-dtype fp8_e4m3
  --moe-backend mega_moe
  --attention-use-fp4-indexer-cache
  --max-model-len 32768
  --max-total-tokens 65536
  --max-num-seqs 2
  --max-cudagraph-capture-size 2
  --chunked-prefill-size 2048
  --disable-prefill-graph
  --gpu-memory-utilization 0.9
  --disable-kvstore
  --host 127.0.0.1
  --port 18000
)
```

Baseline TP2:
```tokenspeed serve "$MODEL" "${COMMON_ARGS[@]}"```

TP2/DCP2 with AG/RS:
```
tokenspeed serve "$MODEL" "${COMMON_ARGS[@]}" \
  --decode-context-parallel-size 2 \
  --dcp-comm-backend ag_rs
```

TP2/DCP2 with A2A:
```
tokenspeed serve "$MODEL" "${COMMON_ARGS[@]}" \
  --decode-context-parallel-size 2 \
  --dcp-comm-backend a2a
```

image

trace above shows dcp's decode overhead is primarily from comms, with A2A reducing the per-layer collective path from three operations and 378 µs to two operations and 292 µs.

A2A improved decode throughput by 17.4% over AG/RS, while remaining 14.3% below the no dcp baseline. dcp preserves the per-rank attention work product (DCP× gathered heads over 1/DCP context), so its primary benefit is kv capacity as the collectives add latency

kimi k3 on 8xb300, using the official checkpoint:

- TP8 baseline, 4K input / 1K output, concurrency 1: 43.78 output tok/s; 870.0 ms median TTFT; 21.52 ms median TPOT.
- TP8 baseline, 16K input / 256 output, concurrency 8: 67.50 output tok/s; 12.46 s median TTFT; 70.01 ms median TPOT.
- TP8 + DCP2 (A2A), 4K input / 1K output, concurrency 1: 41.34 output tok/s; 837.2 ms median TTFT; 23.25 ms median TPOT.
- TP8 + DCP2 (A2A), 16K input / 256 output, concurrency 8: 69.29 output tok/s; 12.01 s median TTFT; 65.92 ms median TPOT.
- DCP4, 4K input / 1K output, concurrency 1: 41.29 output tok/s; 850.8 ms median TTFT; 23.35 ms median TPOT.
- DCP4, 16K input / 256 output, concurrency 8: 67.97 output tok/s; 11.62 s median TTFT; 72.58 ms median TPOT.
- DCP8, 4K input / 1K output, concurrency 1: 58.67 output tok/s; 695.8 ms median TTFT; 16.40 ms median TPOT.
- DCP8, 16K input / 256 output, concurrency 8: 116.75 output tok/s; 6.95 s median TTFT; 41.50 ms median TPOT.

At a fixed 65,536-token capacity:  
DCP2: 1,189,085,184 → 732,561,408 bytes/rank, saving 38.39% (435.4 MiB/rank).  
DCP4: 1,189,085,184 → 508,723,200 bytes/rank, saving 57.22% (648.8 MiB/rank).  
DCP8: 1,189,085,184 → 386,629,632 bytes/rank, saving 67.49% (765.3 MiB/rank).

The MLA history scales as 1/dcp, while about 263.3 MiB/rank of kda state remains replicated, which produces diminishing memory savings at higher DCP degrees.

### CODE REVIEW

File: python/tokenspeed/runtime/layers/attention/registry.py [hunk 1]
```diff
@@ -820,6 +823,21 @@ def create_attn_components(
         if is_hybrid_linear
         else config.backend_name
     )
+    if server_args.mapping.attn.dcp_size > 1:
+        if type(config) is not MLAConfig:
+            raise ValueError(
+                "decode context parallelism is implemented only for dense MLA "
```
kangguangli: I see that the draft configuration forces dcp_size=1, but could enabling DCP for the target still affect the draft model through the shared cache layout? I did not find an explicit restriction for DCP with draft/MTP models.

python/tokenspeed/runtime/models/deepseek_v3.py
```diff
@@ -523,9 +523,18 @@ def __init__(
         self.cli_factor = getattr(config, "cli_factor", 1)
         self.prefix = prefix
 
-        # modification to rope_scaling must be done early enough, b/c e.g. Indexer needs it
+        # Transformers may normalize an absent rope_scaling configuration to a
+        # default RoPE dictionary.  That is not a YaRN configuration and does
+        # not contain the YaRN-specific factor fields.
         if rope_scaling:
-            rope_scaling["rope_type"] = "deepseek_yarn"
+            rope_scaling = dict(rope_scaling)
```
kangguangli: Is there any issue with leaving it unchanged? Setting to `deepseek_yarn' for dsv3 seems reasonable.

python/tokenspeed/runtime/layers/attention/backends/deepseek_v4.py
```diff
@@ -1421,22 +1513,93 @@ def forward_deepseek_v4_decode(
...
+                local_lse,
+                group=self.dcp_group,
+                lse_base=2.718281828459045,
+            )
+        return reconstruct_and_reduce_scatter(
+            out,
+            local_lse,
+            dcp_rank=self.dcp_rank,
+            group=self.dcp_group,
+            lse_base=2.718281828459045,
```
kangguangli: suggest `math.e`

kangguangli (commented):
The overall design of virtually expanding the page size looks good, as it allows all ranks to share the same logical block-table structure. My concerns are:
1. The SKIP_ZERO design is worth discussing. It requires every KV writer that may receive location 0 to skip writing to the reserved null page, while page 0 is frequently used as a placeholder in other paths.
I am concerned that this implicit contract may create maintenance issues in future development. I suggest passing an explicit write mask instead of relying on default writer behavior. For example, Kimi-K3 reuses DeepseekV3AttentionMLA and may invoke apply_rope_mla_set_kv() to update the cache, but this fused path does not currently support SKIP_ZERO.
Alternatively, page 0 could be writable, but in that case read and reconstruction paths should not rely on it always remaining zero, since that invariant is easy to break.
2. During MLA prefill, each rank materializes a full-length cached-prefix tensor, for example [KV0, 0, KV2, 0, KV4, 0] on rank 0. This may introduce redundant memory loads in get_mla_kv_buffer() and extra communication in reconstruct_prefix_kv(), especially when the cached prefix is long. Could we read only the rank-local rows and reconstruct them afterward?
3. Enabling DCP results in a 14.3% throughput regression compared with TP. This seems relatively large considering the communication payload and the NVLink bandwidth of Blackwell. Could you provide a short analysis or profile breakdown of the main sources of this regression?
4.  I suggest adding accuracy tests like evaluation on GSM8K or GPQA. A direct logit-parity comparison against the TP baseline would be even better.

kangguangli:
hello, any update?

qimcis:
> The overall design of virtually expanding the page size looks good, as it allows all ranks to share the same logical block-table structure. My concerns are:
> 
> 1. The SKIP_ZERO design is worth discussing. It requires every KV writer that may receive location 0 to skip writing to the reserved null page, while page 0 is frequently used as a placeholder in other paths.
>    I am concerned that this implicit contract may create maintenance issues in future development. I suggest passing an explicit write mask instead of relying on default writer behavior. For example, Kimi-K3 reuses DeepseekV3AttentionMLA and may invoke apply_rope_mla_set_kv() to update the cache, but this fused path does not currently support SKIP_ZERO.
>    Alternatively, page 0 could be writable, but in that case read and reconstruction paths should not rely on it always remaining zero, since that invariant is easy to break.
> 2. During MLA prefill, each rank materializes a full-length cached-prefix tensor, for example [KV0, 0, KV2, 0, KV4, 0] on rank 0. This may introduce redundant memory loads in get_mla_kv_buffer() and extra communication in reconstruct_prefix_kv(), especially when the cached prefix is long. Could we read only the rank-local rows and reconstruct them afterward?
> 3. Enabling DCP results in a 14.3% throughput regression compared with TP. This seems relatively large considering the communication payload and the NVLink bandwidth of Blackwell. Could you provide a short analysis or profile breakdown of the main sources of this regression?
> 4. I suggest adding accuracy tests like evaluation on GSM8K or GPQA. A direct logit-parity comparison against the TP baseline would be even better.

1. i looked into this a little more - I believe a fused writer path could bypass the null page protection, thus I kept page 0 reserved and moved the owner handling into the backend and now pass a write mask through both the fused and split MLA writers (which keeps the sentinel interpretation out of individual writers)

2. there were indeed redundant memory loads, i've fixed this now (changed it to read only the rank local prefix rows, then it gathers the compact buffers, then interleaves them into global token order)

3. the original regression was mainly because of overhead from collectives plus packing/merging, and index prep and can definitely be optimized, working on this now

4. added a logit comparison test!

+ addressed other comments

python/tokenspeed/runtime/layers/attention/configs/mla.py
```diff
@@ -87,6 +87,59 @@ def generate(
         draft_block_decode = bool(
             is_draft and server_args.speculative_algorithm in ("DFLASH", "DSPARK")
         )
+        mapping = server_args.mapping.attn
+        dcp_size = 1 if is_draft else mapping.dcp_size
```
kangguangli, line 91: DCP with MTP is currently rejected during argument validation, so this branch is unreachable for that combination.
I suggest removing this.

python/tokenspeed/runtime/utils/server_args.py
```diff
@@ -746,6 +769,17 @@ def resolve_speculative_decoding(self):
                 "(topk=1) is wired end-to-end."
             )
 
+        if self.decode_context_parallel_size > 1 and (
```
kangguangli, line 772: Why can’t we support MTP directly? It might be sufficient for the draft model to use the same parallel strategy as the target model.

kangguangli (commented):
Thanks for the update. The compact prefill reconstruction addresses an important part of the original page-0 concern. I still have several questions and suggestions.

  ## 1. Cache placement

  CachePlacementContract currently has two effective consumers:

  - The MLA cache pool reads layer_placements to build _cyclic_layers, which only controls the SKIP_ZERO writer behavior. With the new explicit write_mask, and with compact reconstruction no longer depending on page 0 remaining zero, this path appears redundant.

  - The DSV4 cache pool reads only dcp_size and dcp_rank; it does not consume DSV4’s per-layer placement values. These topology fields could be passed directly from the attention configuration or, preferably, through a small shared DcpContext.

  The contract does not drive cache-group geometry, field shapes, memory planning, addressing, or backend communication. It is constructed after the physical layout has already been derived from dcp_size, and its layer-level representation cannot describe a DSV4 layer containing sharded compressed KV alongside replicated SWA, indexer KV, and states. The sliced target/draft contracts are also not the runtime source used by the pools. Therefore, the current abstraction appears redundant and potentially misleading.

  I suggest removing it for now. It can be reintroduced when there is a concrete need, with CachePlacementContract serving as the source of truth for placement consistency across cache planning and runtime behavior.

  ## 2. Page-0 dependency and the new write mask

  To clarify, my original concern about SKIP_ZERO was mainly about removing the read-side assumption that page 0 must always remain zero, since it may be difficult to guarantee that every current and future KV writer preserves this invariant. The new compact reconstruction in prefill addresses this concern well.

  Now that the read paths no longer depend on the contents of page 0, masked and unmasked page-0 writes should be equivalent from the cache-semantics perspective. The write mask may still provide a performance benefit by avoiding unnecessary memory operations and concurrent writes to the same page-0 entry.

  However, the indices and launch shape remain full-length, and there is not yet profiling evidence showing whether this benefit is significant. Removing the mask would make the implementation simpler. What do you think?

  ## 3. SKIP_ZERO appears redundant

  Regardless of whether the explicit write_mask is retained, I think the SKIP_ZERO parameter can be removed. The split writer currently applies both write_mask and SKIP_ZERO, while for DCP they represent the same loc != 0 condition.

  At most one of these mechanisms should remain; otherwise, future writers will need to understand and maintain two overlapping contracts.

  ## 4. DeepSeek-V4-Flash KV-cache memory breakdown

  Let codex calculate the following logical cache-payload breakdown for an 8,192-token context, using the current field layouts and including 44 cache layers:

| Cache                              | Logical payload | Percentage |
|------------------------------------|----------------|------------|
| SWA KV, 44 layers, 128-token window | 3.14 MiB       | 7.13%      |
| Ratio-4 compressed KV              | 25.10 MiB      | 56.95%     |
| Ratio-128 compressed KV            | 1.41 MiB       | 3.19%      |
| **All compressed KV**                  | 26.51 MiB      | 60.14%     |
| Ratio-4 MXFP4 indexer KV           | 2.79 MiB       | 6.33%      |
| Ratio-4 compressor state           | 1.31 MiB       | 2.98%      |
| Ratio-128 compressor state         | 10.00 MiB      | 22.69%     |
| Indexer state                      | 0.33 MiB       | 0.74%      |
| **All state**                          | 11.64 MiB      | 26.41%     |
| Total logical payload              | 44.08 MiB      | 100%       |

  This is a logical payload calculation and does not include arena packing, alignment padding, per-group capacity requirements.

  Since compressed KV accounts for approximately 60% of the logical payload, while the PR reports only a 3.82% reduction in the complete arena at DCP2, could you provide a breakdown of the actual arena bytes by cache group, including packing and padding? This would help explain where the difference comes from and how much memory each replicated group contributes.

  It would also be useful to confirm whether the reported 8,192-token result includes the NextN draft cache. If it is target-only, the payload calculation should use the 43 target layers rather than all 44 cache-layer entries.

  ## 5. Accuracy tests and the logit-dump interface

  The previously discussed tests do not appear to be included in the latest push. The PR also adds the TOKENSPEED_TEST_LOGIT_DUMP_DIR runtime hook, which copies and saves the full pre-sampling logits. I could not find the corresponding comparison script or automated test in the current diff.

  A direct logit comparison against a TP/DCP1 baseline would be an excellent correctness check. I recommend including the results in the PR description.
  I also suggest adding a DCP CI evaluation based on:
  test/ci/eval/deepseek-v4-flash-mtp-evalscope-gsm8k.yaml

  This would provide an end-to-end gate covering DSV4 Flash, MTP, cache placement, and DCP communication.

python/tokenspeed/runtime/layers/attention/configs/mla.py
qimcis: will do

python/tokenspeed/runtime/utils/server_args.py
qimcis: we can, however i believe it will take some additional engineering work, ill look into it after fixing up the other problems

qimcis:
> 2. Page-0 dependency and the new write mask
To clarify, my original concern about SKIP_ZERO was mainly about removing the read-side assumption that page 0 must always remain zero, since it may be difficult to guarantee that every current and future KV writer preserves this invariant. The new compact reconstruction in prefill addresses this concern well.

> Now that the read paths no longer depend on the contents of page 0, masked and unmasked page-0 writes should be equivalent from the cache-semantics perspective. The write mask may still provide a performance benefit by avoiding unnecessary memory operations and concurrent writes to the same page-0 entry.

> However, the indices and launch shape remain full-length, and there is not yet profiling evidence showing whether this benefit is significant. Removing the mask would make the implementation simpler. What do you think?

It is true that the read paths do not depend on the contents of page 0 (so masked and unmasked p0 writes are technically equivalent) but i believe that the mask should still be kept as in the current implmentation DCP maps every non-owner row to slot 0, thus removing the mask could cause multiple writers to write to the same cache row. 

I think the best solution here is probably to keep one ownership mask, generate it once with metadata from forward, and reuse it across all writers ( including chunked prefill). Otherwise, ill remove SKIP_ZERO and address your other points (which i agree with) @kangguangli

kangguangli:
> > 2. Page-0 dependency and the new write mask
> >    To clarify, my original concern about SKIP_ZERO was mainly about removing the read-side assumption that page 0 must always remain zero, since it may be difficult to guarantee that every current and future KV writer preserves this invariant. The new compact reconstruction in prefill addresses this concern well.
> 
> > Now that the read paths no longer depend on the contents of page 0, masked and unmasked page-0 writes should be equivalent from the cache-semantics perspective. The write mask may still provide a performance benefit by avoiding unnecessary memory operations and concurrent writes to the same page-0 entry.
> 
> > However, the indices and launch shape remain full-length, and there is not yet profiling evidence showing whether this benefit is significant. Removing the mask would make the implementation simpler. What do you think?
> 
> It is true that the read paths do not depend on the contents of page 0 (so masked and unmasked p0 writes are technically equivalent) but i believe that the mask should still be kept as in the current implmentation DCP maps every non-owner row to slot 0, thus removing the mask could cause multiple writers to write to the same cache row.
> 
> I think the best solution here is probably to keep one ownership mask, generate it once with metadata from forward, and reuse it across all writers ( including chunked prefill). Otherwise, ill remove SKIP_ZERO and address your other points (which i agree with) @kangguangli

Sure. I'm just wondering how much performance benefit we get from the mask. If the gain is so small that we cannot measure it in tests, I think keeping the implementation simple would be better.

I'd like to write a test for this and post the results later. So you can address the other points first, and we can discuss this again after I post the results.

PR #1167, opened 2026-08-20
fix(cache): route null decode pages to dummy slot

jhinpan:
## Summary
- route null or negative page IDs to the fixed dummy slot in both fused target decode prep and DFlash draft prep
- add boundary, overflow, negative-page, and CUDA-graph replay coverage for the shared cache-location contract

This is an independently landable correctness fix found while investigating #1132. At exact 131K it only changed average accepted length from 1.0000 to 1.0365, so it is not presented as the full acceptance-collapse fix.

## Test plan
- `pytest -q test/runtime/test_decode_cache_loc.py test/runtime/test_sliding_cache_loc.py` (`11 passed` on gfx950)
- `isort --check-only` and `black --check` on the changed Python files
- exact 131072/1024 TP8/EP8 DSpark A/B with the draft guard enabled


jhinpan:
I have read the DCO document and I hereby sign off on the commits contained in this pull request.

jhinpan:
Superseded by #1169 with an identical tree and a commit-level DCO sign-off. Closing this branch rather than rewriting its published history.

PR #1168, opened 2026-08-20
feat(dspark): expose forced-rejection diagnostics

jhinpan:
## Summary
- add `TOKENSPEED_FORCE_SINGLE_TOKEN_VERIFY=1` to keep the full proposal/target-verify round while committing exactly one target token
- reuse the existing per-row force mask, preserve remote-recovery semantics, and keep the default local decode path unchanged
- document the diagnostic and cover CPU, mixed-row, default-off, and CUDA-graph behavior

This supplies the acceptance-independent gate requested by #1132. On exact 131072/1024 TP8/EP8 it measured 36.80 ms per speculative round versus 17.06 ms ordinary decode (`2.157x`). Forced DSpark remained token-identical to ordinary greedy for a 4K/1K repeated prompt and an exact 131K/128 natural prompt.

## Test plan
- focused DSpark/runtime suite: `84 passed` on gfx950
- `isort --check-only`, `black --check`, `git diff --check`
- exact 131072/1024, B1/C1, TP8/EP8, FP8 KV, greedy, decode graphs, prefix caching off: three unprofiled runs per arm
- separate all-rank first-64 trace for attribution

Local `pre-commit` could not create its configured Python 3.12 hook environment because this ROCm container only ships Python 3.10; the equivalent Python formatting/syntax checks above passed.


jhinpan:
I have read the DCO document and I hereby sign off on the commits contained in this pull request.

jhinpan:
Superseded by #1170 with an identical tree and a commit-level DCO sign-off. Closing this branch rather than rewriting its published history.

PR #1172, opened 2026-08-20
perf(k3): fuse the multi-token MoE front

chenht2022:
## Summary

- fuse the K3 router, routed-down, and shared gate/up projections for captured multi-token decode batches on SM100
- consume the fused FP32 strided views directly in route/top-k/MXFP8 preparation and allow the shared SiTU path to write BF16 without an extra materialization kernel
- keep the existing singleton GEMV path for `M == 1`; use the fused front for captured `M >= 2` batches

## Dispatch A/B

Three paired serving repetitions per CUDA graph bucket, 256 input / 256 output tokens, TP8/EP8 on 8x B200:

| M | Output throughput | TPOT |
|---:|---:|---:|
| 1 | -0.62% | -0.72% |
| 2 | +2.80% | +3.04% |
| 4 | +5.05% | +5.02% |
| 8 | +2.44% | +2.72% |
| 12 | +5.79% | +3.63% |
| 16 | +0.21% | +0.10% |

Positive values are improvements. The all-M candidate regressed at `M == 1` in all three TPOT repetitions, so the final dispatch retains GEMV only for that singleton case. The old exact-`M == 16` allowlist was otherwise too restrictive.

## Testing

- `pre-commit run --files` on all changed Python files
- complete activation and focused FlashInfer TRT-LLM SiTU MoE tests: 100 passed
- CUDA graph capture and warmup generation at `M = 1, 2, 4, 8, 12, 16`
- 36 serving benchmark cases, zero failed requests

### DIFF
--- a/python/tokenspeed/runtime/layers/moe/expert.py
+++ b/python/tokenspeed/runtime/layers/moe/expert.py
@@ -291,6 +291,7 @@ class MoELayer(torch.nn.Module):
         num_global_tokens: int,
         max_num_tokens_per_gpu: int,
         do_finalize: bool = True,
+        prepared_input: tuple[torch.Tensor, torch.Tensor, torch.Tensor] | None = None,
         low_latency: bool | None = None,
         overlap_fn: Callable[[], None] | None = None,
     ):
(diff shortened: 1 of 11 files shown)


chenht2022:
Verified not useful on latest main. Closing this PR.

PR #1221, opened 2026-08-24
feat(kernel): write AttnRes block snapshots

nperrin-fr:
The CUDA AttnRes kernel refused block_write_idx >= 0, so the eight block-write layers failed the availability check. That check is per layer, so a failing attention-side call dragged the layer's mlp-side call with it and both ran the split partial/combine fallback instead.

The consumer loop already holds the updated prefix row in registers and already stores it to the layer residual; the snapshot is one more store to the target block row, so the constraint was never intrinsic. Lifting the n == prefix_n test out of the HAS_DELTA guard also covers block-write layers, which pass no delta.

Measured at bs=8 on GB300, per decode step: the split fallback disappears entirely (partial 15, dual 8, combine 15 -> 0), the fused kernel absorbs them (171 -> 187 launches), and AttnRes device time drops 1159.3 -> 977.7us. Median step time improves 36us (95% CI [29, 42], 16 boots) on a tree that carries the decode-projection route table, and 95us (95% CI [90, 101], 8 band-matched comparisons over 16 boots) on one that does not; the two trees differ in both that table and build freshness, so the magnitude is quoted per tree rather than averaged. The rest of the device-time saving was already hidden under other work. Boots were paired by CUDA-graph capture band, since a single unpaired boot carries ~250us of layout variance. Build cost is 47s and 114KB of .so.


nperrin-fr:
Both review findings are addressed in a651b123.

**P1 (widen the snapshot row offset)** — confirmed. `block_write_idx * block_stride_r` multiplied in 32-bit at both vector-store branches while the neighbouring `tb * block_stride_m` term was already widened. Both are now cast, matching `residual_addr`. Not reachable at the current `max_prefill_tokens` (8192 tokens over 8 blocks is 411M, under INT_MAX) but reachable at ~43k-token chunks, so it was latent rather than active.

**P2 (preserve snapshots for strided block storage)** — confirmed, and it turned out to be half the defect. The kernel writes through **two** parameters: the block snapshot, and under `delta` the accumulated prefix via `layer_res`. The adapter passed `.contiguous()` on both, so a strided caller lost either write into the temporary. The second one is load-bearing — `_forward_fused_attnres_graph` relies on `prefix_sum` being updated in place before handing it to the MoE as the residual, so a copy would have fed the MoE a pre-delta residual while the returned mix looked correct. Both writes are now copied back only when a copy was actually taken; contiguous callers are unaffected.

Neither is reachable from the K3 caller today — `kimi_k3.py` passes the full contiguous tensor when `block_write_idx >= 0` and a leading-dim slice otherwise — but nothing enforced that, so the guard is worth having.

Two regression tests, each confirmed to fail without its fix (7132/7168 and 14186/14336 elements mismatched). Both use T > 1: with T == 1 a size-1 dimension satisfies contiguity regardless of stride, which made the first version of the test vacuous.

Verified the fixes are free: per-call AttnRes cost is 5.15us before and after, and device time 963.3 -> 956.2us/step. Step time is band-confounded here, so the per-kernel number is the one to read.

PR #1223, opened 2026-08-24
feat(runtime): prefill chunk pipeline + control-plane/data-plane split

FC-Li:
## Summary

Two things ship together here: the **prefill chunk pipeline** (same-request chunk-level pipelining across PP stages on the prefill role) and the **event-loop cohesion refactor** that made room for it. The later commits finish that second half — the control plane no longer touches CUDA, and the pieces that were living in the executor for historical reasons moved to where their data already is.

Commits, oldest first:

**`551b55a6` feat(runtime): prefill chunk pipeline + event-loop cohesion refactor**
Per-stage layer windows (including non-uniform `--pp-layer-partition`), boundary AttnRes handoff, PD transfer plans narrowed to each stage's layers; the unified loop runs the pipeline as in-flight depth `pp_size`. EventLoop becomes a coordinator: pause/resume, EPD admission, PD transfer events and L2 cache-op tracking each move into their own module and enter the loop as one-line hooks. `advance_scheduler` becomes the only caller of `scheduler.advance`, invoked explicitly from the loop body — helpers return events. Design principles in `docs/design/event-loop.md`.

**`7aa4309b` feat(kernel): make the KDA prefill host boundary hint a required int64 tensor**
Every KDA paged-prefill solution plans its chunk indices on the host. With boundaries only on device, that plan costs a stream-synchronizing D2H per KDA layer per prefill chunk, which stalls the launch thread behind all queued GPU work and serializes the chunk pipeline's stages. `cu_seqlens_cpu` becomes a required host int64 tensor, so no solution can silently fall back to re-reading device boundaries.

**`a6b5b583` fix(runtime): keep runtime profiling usable on graph-mode servers**
A profiler that first attaches *after* a CUDA graph is captured invalidates it — every later replay dies with `cudaErrorLaunchFailure`. One empty profiler session warms CUPTI before any capture. Separately, CUPTI GPU-activity tracing of graph-captured NCCL kernels is unstable on torch 2.11 / NCCL 2.28, so a multi-rank graph-mode profile now drops the GPU activity and says so instead of crashing the engine (`--enforce-eager` for a full GPU timeline; `TOKENSPEED_PROFILE_GPU_WITH_GRAPHS=1` to override).

**`28ac2eef` refactor(runtime): finish the control-plane / data-plane split** — no behaviour changes
- `PendingExecution` replaces the raw result in the in-flight queue. It owns the one place the control plane waits for the GPU (join the forward-thread future, then the D2H copy event) and makes that sync exactly-once *by construction*: the future is private, so host tensors are unreachable without going through `result()`; `result()` memoizes; `sync()` raises on a second call — which is what surfaced a redundant sync on the commit path.
- Batch logging moves to `BatchLogger` on the loop. Those lines report scheduler quantities the loop already samples, so page/queue counts were being threaded down through dispatch and across the forward thread just to reach a `logger.info`, and the decode counters were written on the control plane but read on the forward thread. The executor loses two now-dead config fields, and the `_with_log` wrapper folds back into a single `execute_forward_op`.
- Load reporting moves to `LoadReporter` next to its sinks. The dirty-tracking latch goes away entirely: the sinks already dedup unchanged tuples and a frozen round only comes around once per idle sleep, so the latch bought three scheduler getters per millisecond in exchange for "whoever forgets to mark dirty publishes a stale tuple".
- Per-role dispatch moves to `forward_dispatch.py` — the role is fixed at startup, so it resolves once instead of an `isinstance` chain per round.
- `DpForwardMetadata` passes as one object instead of four unpacked fields; several always-supplied parameters lose their defaults; the decode `#pages(active/cached/total)` total now reports usable pages, the same total the load snapshot and the Prometheus gauge publish.

**`0bf4e87d` fix(runtime): enforce structured output on the PD decode node**
Prefill and decode each compile their own matcher, and the decode node never samples the one token the prefill node produced — so its matcher was always one token behind and the decode path dropped grammar inputs rather than mask from a stale state. The failure was silent: on the capturable path the mask is all-ones and the matcher never advances, so a `json_schema` request on a PD deployment returns free-form text. The bootstrap token is now accepted into the decode matcher where it already lands in `output_ids`, before `check_finished`; with the matcher in sync the decode path masks like every other role. If the prefill side could not supply that token, the request's grammar is dropped and the lost enforcement is logged, rather than masking from a state nothing can resynchronize.

`28ac2eef` deliberately carries the *old* grammar behaviour so the refactor is verifiably behaviour-preserving; the fix lands separately and can be reverted or cherry-picked on its own.

## Test Plan

- `pre-commit run --all-files` clean; the CI-lint ruff selectors (`F401,F841,F821,F823,F811,F541,UP*` over `python/tokenspeed/runtime`, plus `tokenspeed-scheduler`) clean.
- New CPU-only tests: `test_forward_thread.py` (FIFO order, exception relay, non-blocking submit, and the exactly-once sync contract), `test_batch_log.py`, `test_forward_dispatch.py` (one case per role, including the decode node's receive ordering and the prefill node's token hook), plus load-reporter and PD-matcher cases in the existing suites.
- Targeted runs over every touched module pass (124 tests across `test_forward_dispatch`, `test_forward_thread`, `test_batch_log`, `test_load_snapshot`, `test_generation_output_processor`, `test_retract_readmit`, `test_dspark_config`, `test_model_executor_cache_state`, `test_draft_page_table_scrub`, `test_event_loop_*`, `test_pause_controller`, `test_zmq_msgpack`).
- The full `test/runtime` sweep was **not** completed locally. Its remaining failures all sit in `test/runtime/distributed/`: `test_auto_backend.py` fails identically at `551b55a6` (pre-existing), and the PD/EPD quality tests error because their checkpoint is not available offline. `test_comm_manager.py` and `test_draft_moe_capture_global_bs.py` (8-GPU NCCL) fail in a fixed set that did not move across the session and share no code path with these changes — noted as inference, not verification. CI is the arbiter for those.
- Not re-validated locally: PD/PP end-to-end quality and the chunk-pipeline throughput numbers from `551b55a6`.

### DIFF
--- a/docs/design/event-loop.md
+++ b/docs/design/event-loop.md
@@ -1,0 +1,138 @@
+# Event Loop: Design Principles
+
+This document records the design principles of the scheduler event loop
+(`python/tokenspeed/runtime/engine/event_loop.py`). It is the reference for
+where new logic belongs, how components feed results back into the scheduler,
+and what the loop body itself is allowed to contain. The rules below were
+established deliberately; treat deviations as bugs in review.
+
+## Principle 1: the event loop is a coordinator, nothing more
+
+`EventLoop.event_loop` sequences components; it does not implement them.
+Domain logic — pause/resume semantics, EPD admission, PD transfer handling,
+L2 cache-op tracking, wire handshakes, multimodal batch assembly — lives in
+its own module and enters the loop as a **single-line hook**. The loop body
+should read, top to bottom, as the schedule of one scheduling round, with no
+feature's internals inlined into it.
+
+Consequences:
+
+* Low-frequency or optional features (pause/resume control, EPD, SMG
+  transport, kvstore) must never make the *normal* scheduling path harder to
+  read. If understanding decode throughput requires skipping over your
+  feature's code, the feature is in the wrong place.
+* When a feature needs several collaborators of the loop, give it a hooks
+  class (see below) instead of weaving branches through the loop and its
+  helpers.
+
+## Principle 2: scheduler feedback is explicit and centralized
+
+`advance_scheduler` (`scheduler_utils.py`) is the **only** caller of
+`scheduler.advance`, and it is invoked **only explicitly and directly in the
+`event_loop` body — never from helpers**. Helpers RETURN their events; the
+loop applies them. Reading the loop body alone must reveal every point where
+the scheduler's state advances, and why.
+
+There are exactly two call sites, each with a documented reason:
+
+* **Head of the round** — completed L2 cache-op events
+  (`_cache_hooks.poll_ready_events()`). These must advance *before*
+  `next_execution_plan`, otherwise cache-gated admissions are delayed by a
+  full round.
+* **Tail of the round** — forward results and PD transfer events, funneled
+  through the single `request_changes` list. These can only exist after
+  dispatch/commit, and must advance before the *next* round plans.
+
+Anything that produces scheduler events (a new transfer backend, a new async
+op kind) either returns events into one of these two points or adds a new
+explicit call site in the loop body with a comment stating why the existing
+points don't fit. It must not call `advance_scheduler` itself.
+
+## Principle 3: correctness never depends on the in-flight depth
+
+The loop is parameterized by `in_flight_depth`: 0 (classic synchronous
+commit), 1 (the overlap schedule), or `pp_size` (the prefill chunk pipeline).
+Dispatched forwards await commit in the `in_flight` queue; the tail commits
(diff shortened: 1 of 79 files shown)


FC-Li:
Superseded by #1224.

The event-loop cohesion / prefill chunk pipeline half of this branch landed on `main` as #1214 (squash-merged), so this PR's diff now double-counts it. The four follow-up commits have been rebased onto current `main` in #1224, which is the reviewable one.

PR #1231, opened 2026-08-24
DeepSeek V4-Flash on AMD gfx950: portable kernels, decode optimizations, and an EP=1 routing fix

indianspeedster:
## Summary

Runs DeepSeek V4-Flash on AMD gfx950 (MI355X) without FlashMLA or DeepGEMM, then
closes part of the decode gap against the B200 reference build. The branch is the
portable-kernel port, four decode-kernel optimizations, one routing correctness
fix, and a fused indexer-state compressor.

## Correctness fix worth reading first

At EP=1 the MoE planner selected `gluon_mxfp4_dynamic_moe_apply`, which advertises
`kernel_routing`. V4 scores experts with `sqrt(softplus(logits))`, which no fused
backend routing reproduces, so the model selects its own top-k -- and the layer was
then re-encoding that already-chosen top-k as log-probs over 256 experts so the
backend's own `TopK` could recover it. The backend applied `correction_bias` and
`routed_scaling_factor` a second time, on top of the values routing had already
folded in.

The model stayed fluent throughout, so this read as a slow path rather than a wrong
one. It is not: **GSM8K at TP=1 was 3.56%** while the model answered "17+25=42"
correctly. Constraining the layer to `precomputed_topk` picks
`gluon_mxfp4_precomputed_moe_apply` -- same Gluon MXFP4 family -- and the scatter,
second top-k and second renormalize disappear with it.

EP=8 already resolved to a precomputed kernel, so the fix is a no-op there and that
configuration was never affected. The bug keys off **EP degree, not TP degree**:
any non-EP configuration, including TP=8 without expert parallelism, took the
broken path.

## Measurements

Single MI355X, TP=1, 1k/1k, `--moe-backend gluon`, `--kv-cache-dtype fp8_e4m3`.
Throughput is the mean of four consecutive runs on one server process, with the
observed range given, because cross-process spread on this benchmark is real.

| | baseline | this branch | |
|---|---|---|---|
| Throughput, c=1 | 20.00 tok/s | **27.45** (27.01-28.03) | +37% |
| Throughput, c=32 | 409.80 tok/s | **546.8** (503.5-577.2) | +33% |
| TPOT, c=1 | 48.65 ms | **35.38 ms** | -27% |
| GSM8K, 5-shot | 3.56% | **92.5-94.8%** across four runs | correctness fix |

Two caveats stated plainly. The c=32 figure spreads 13.5% across four back-to-back
runs, so treat it as approximately 545 +/- 40 rather than a precise value; c=1 is
tight at 3.7%. The baselines were captured on a different server process, and
cross-process variance is the same effect that produces the GSM8K spread above, so
the percentage gains carry that uncertainty -- most of it in the c=32 row.

TP=8 with expert parallelism has not been re-measured against this branch tip.

## Kernels added

- **Fused paged sparse decode attention** (`deepseek_v4.py`). Reads the decode
  caches inside the kernel instead of materializing them. `num_stages=1`, a scale
  broadcast and bf16 tiles took it 89.9 -> 32.8 us/layer (2.74x) by moving it off
  one wave per SIMD; 284 VGPRs still means one wave, so the remaining ~1.5x to
  FlashMLA's 21.4 us is an occupancy problem, not an algorithmic one.
- **Indexer select-all fast path.** Skips selection entirely when every candidate
  survives.
- **mHC mix partials reduced as tiles** rather than scalar loads.
- **Fused `sqrt(softplus)` top-k routing** (`deepseek_v4_softplus_sqrt_topk.py`).
  One bitonic pass over a packed ordered key; softplus uses a compensated `log1p`
  that holds where fp32 `log(1+exp(x))` underflows below x = -16.6. 36.5 -> 10.0 us.
  Verified to select the identical expert set as the tensor-op reference on 24,576
  tokens across five logit distributions, max weight error 6e-8.
- **One-launch indexer-state compressor**
  (`deepseek_v4_compress_windows.py`). Replaces roughly 25 tensor ops per call,
  twice per layer, by streaming the window in tiles under a running max and sum so
  the softmax never materializes and the RMS norm folds into the same pass. 12.1x
  at ratio 4, 5.9x at ratio 128. The tensor-op version is retained as the reference
  the kernel is tested against.

## Validation

`pre-commit run --all-files` clean. The pre-existing V4 runtime suite passes at
152, against 6 `kv_pool` failures that predate this branch and are unrelated to it.

This branch ships no new tests by request, so the numbers below come from
throwaway harnesses rather than anything committed here.

- The compressor kernel was checked bit-exact against the tensor-op reference it
  replaces, on the live cache geometry, across fp32/bf16/fp16 caches,
  prefill-scale sequential positions, non-unit feature strides, saturating score
  magnitudes and all-masked windows.
- The fused router top-k selected the identical expert set as the tensor-op
  reference on 24,576 tokens across five logit distributions, max weight error
  6e-8.
- End-to-end GSM8K and throughput are in the Measurements section above.

Two defensive paths in the compressor -- clamping the base-page index the way the
reference does, and reading the feature dimension through its stride -- are not
reachable in the current runtime: nothing populates the base-offsets map, and the
live cache is unit-stride in that dimension. They are one instruction each. With
the tests removed, nothing pins them if either assumption changes.


indianspeedster:
Closing: #1201 (`feat(deepseek-v4): add AMD MI350 support`) already covers this.

That PR landed while this branch was in flight and rebuilt the same subsystem, so
this is not a rebase away -- the architecture this branch is built on no longer
exists on `main`:

PR #1234, opened 2026-08-24
perf(comm): dispatch MNNVL all-reduce on one-shot traffic

nperrin-fr:
Our vendored MNNVL kernel picked one-shot vs two-shot from a bare token count, `token_num > kMnnvlOneShotMaxToken` with the cap at 128. FlashInfer, which this was adapted from, picks from the one-shot traffic instead:

    num_tokens * hidden_dim * tp_size * elem_size <= 1 MiB  ->  one-shot

One-shot has every rank publish its whole payload, so its cost scales with the rank count; dropping the tp_size and hidden_dim terms lost exactly the part that matters. For Kimi-K3 (hidden 7168, bf16) the upstream rule caps one-shot at 9 tokens on TP8 and 4 on TP16, so 128 was wrong at both world sizes, and every decode all-reduce at bs=64 took the slower path.

Measured on 4x GB200 (bench_mnnvl_vs_nccl, 200 iters), which also confirms the rule's shape -- one-shot degrades with world size, two-shot does not:

    64 tokens   one-shot  w8 19.0us -> w16 32.7us  (1.72x)
                two-shot  w8 14.0us -> w16 15.0us  (1.07x)

Crossovers land near 4.4 MiB (w16) and 5.3 MiB (w8) of one-shot traffic -- close enough to call the criterion constant-traffic. Upstream's 1 MiB is conservative for this shape, but both paths sit on the ~13us latency floor below the crossover, so the conservatism is free and the constant stays matched to FlashInfer rather than tuned to one model.

Kimi-K3 TP16 decode, 4096-in/1024-out at bs=64, same nodes back to back: mean TPOT 31.99 ms -> 30.16 ms (-5.7%), arms non-overlapping.

`use_oneshot` is a real parameter again. The device used to recompute the choice and ignore the caller, which let host and device disagree -- and the two paths need different workspace layouts, so a disagreement read and wrote the wrong region instead of failing. The launcher now honours the flag, one host helper resolves the rule for all three dispatch sites, and the one-shot lane is sized from that same rule. The derived cap is resolved once at workspace creation and stored on the workspace, so changing the override later cannot outrun the allocation: a request that would exceed the sized lane falls back to two-shot. Sizing is unchanged in practice at 56 MiB per buffer for both TP8 and TP16, since the two-shot term dominates.

Tests cover what the old split hid: both strategies forced at the same token count must agree, which fails if the device recomputes; and an override raised after the workspace exists must still fall back rather than overrun.


nperrin-fr:
> @nperrin-fr is this flaky https://github.com/lightseekorg/tokenspeed/actions/runs/32791145150?pr=1234

This is threshold tuning so feel like it shouldn't change the quality. But we can re-run once and see how it goes.

PR #1270, opened 2026-08-27
fix(kimi3): warm the MoE auxiliary stream before graph capture

nirvedhmeshram:
## Problem

`KimiLinearMoE` gates its stream fork on `get_is_capture_mode()`:

```python
with self.stream_fork.scope(enable=get_is_capture_mode()) as fork:
```

`_capture_one` runs several warmup forwards on the capture stream before recording, but those run with capture mode **off**. So `_active` stays `False`, `branch()` is a passthrough, and the auxiliary stream is never touched. Capture then activates that stream for the first time, and the first hipBLASLt call on it performs its lazy handle/workspace setup *inside the capturing stream*:

```
Hip error: 'operation not permitted when stream is capturing'(900)
  at .../hipblaslt/library/src/amd_detail/hipblaslt.cpp:147
```

One error per rank, then startup deadlocks with every GPU at 0% utilization. The server never reports healthy.

Reproduced on 8x gfx950 (MI355X, ROCm 7.2.2, torch 2.13.0+rocm7.2) serving Kimi-K3 MXFP4 at **TP8/EP1**, where the tensor-parallel MoE places a projection on the forked branch. **TP8/EP8 captures cleanly** both before and after this change, which is why it has gone unnoticed.

## Fix

Gate `enable` on the graph phase and `overlap` on capture mode:

```python
with self.stream_fork.scope(
    enable=get_is_cuda_graph_phase(),
    overlap=get_is_capture_mode(),
) as fork:
```

Warmup now runs the same branches on the auxiliary stream serially, so every backend is initialized before capture begins. This matches the contract `_capture_one` already documents —

> Capture-only auxiliary branches use this graph phase to warm their own streams serially.

— and the pattern the other fork-using models follow (`inkling`, `qwen3_5_moe`). `KimiLinearMoE` was the only caller passing the bare `enable=get_is_capture_mode()`.

Behaviour outside the graph phase is unchanged: eager serving still takes no fork.

## Validation

8x gfx950, Kimi-K3 MXFP4, `--tp 8`, MLA, fp8 KV cache, 4k input / 1k output.

| config | capture | probe accuracy | tok/s @ concurrency 1 / 8 / 16 |
|---|---|---|---|
| EP1 before | deadlock | n/a | n/a |
| **EP1 after** | clean | 8/8 | 53.98 / 259.08 / 448.57 |
| EP8 before | clean | 8/8 | 73.84 / 246.55 / 350.70 |
| **EP8 after** | clean | 8/8 | 73.83 / 247.51 / 351.51 |

EP8 is unchanged to three significant figures, so the change is inert where capture already worked. Accuracy is a fixed greedy prompt set (arithmetic, recall, short reasoning) diffed against an EP8 reference.

## Test

`test/runtime/test_kimi_k3_moe_fork_warmup.py` drives the real `KimiLinearMoE.forward` with a spy fork and asserts the enable/overlap decision for all three states: warmup, capture, and eager. CPU-only — no real streams or capture.

Verified the test fails on the previous call site and passes on this one, so it pins the regression rather than just the current behaviour.

`pre-commit run --all-files` passes.

### DIFF
--- a/python/tokenspeed/runtime/models/kimi_k3.py
+++ b/python/tokenspeed/runtime/models/kimi_k3.py
@@ -102,6 +102,7 @@ from tokenspeed.runtime.distributed.mapping import Mapping
 from tokenspeed.runtime.distributed.pp_stage import PPStageState, pp_layer_window
 from tokenspeed.runtime.execution.cuda_graph_wrapper import (
     get_is_capture_mode,
+    get_is_cuda_graph_phase,
 )
 from tokenspeed.runtime.layers.activation import SituAndMul
 from tokenspeed.runtime.layers.layernorm import (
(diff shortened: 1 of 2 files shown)


raikonenfnu (commented):
Looks good Nirvedh! Can we drop the moe_fork_warmup in favor of using TP8/EP1 in the test/eval K3 CI?

PR #1281, opened 2026-08-28
feat(npu): support qwen3-0.6B on ascend

chenhongluo:
• ## Summary

  - Add Ascend NPU support for Qwen3-0.6B inference.
  - Add the standalone `tokenspeed-kernel-ascend` package with Ascend implementations for:
    - RMSNorm and fused add RMSNorm
    - Rotary embedding
    - MHA prefill, extend, and decode using `npu_fused_infer_attention_score`
  - Add HCCL communication support and device-neutral runtime handling.
  - Support ACL Graph decode with paged KV cache using a 128-token page size.
  - Add installation scripts, documentation, and Ascend-specific unit tests.
  - Keep NVIDIA and AMD behavior unchanged; platform-specific logic remains behind the kernel boundary.

  ## Test Plan

  - Runtime tests: `53 passed, 5 skipped`
  - Ascend kernel tests: `24 passed`
  - Cross-platform LayerNorm dispatch tests: `10 passed`
  - Validated Qwen3-0.6B BF16 inference on Ascend NPU with ACL Graph decode for batch sizes 1, 2, and 4.
  - Verified an end-to-end chat request produced the expected response:
    - Prompt: `你好，请用一句话介绍你自己。`
    - Output: `我是AI助手，致力于帮助您解决问题和获取信息。`
  - Verified four concurrent requests with prompt lengths of 99, 179, 339, and 659 tokens all completed successfully with deterministic output.


borontion (commented):
overall looks good. if urgent we can merge this first, but better to have a npu ci to guard things so we can continue improving the code without breaking things. if not urgent for now, need to address a few things before merging:

- for torch npu can we also add torch shim like https://github.com/lightseekorg/tokenspeed/blob/main/tokenspeed-kernel/python/tokenspeed_kernel/_triton.py so runtime can just import torch from there instead of importing both torch or torch_npu and adding lots of branch.
- similarly for triton-ascend, if we can hide it behind the triton shim that would be great. ideally runtime should just use bare torch / triton and not care about the underlying device.
- for tokenspeed-kernel, for now we organize kernel registry as ops/<family>/<impl> where <impl>=triton,gluon,cutedsl. if npu is using triton, then we should register it under ops/<family>/triton instead of ops/<family>/ascend. this is just convention for now, maybe not ideally, but let's try to follow it.

PR #1301, opened 2026-08-29
fix(prefill graph): capture for MLA targets bound to the Cache contract

nperrin-fr:
## Why

`_dummy_group_tables` gated the capture-time cache metadata on `uses_cache_groups`, reading it as "this backend needs paged cache metadata". For MLA that reading is inverted. `flashmla.py` sets it to `bool(config.is_draft)` and says why: the draft consumes the wrapper's per-group `block_tables`, while the *target* reads the richer `cache_metadata` and must stay off that path. `trtllm_mla` does the same; `tokenspeed_mla` and `mla` never set it and inherit False.

So when the top-level backend was a plain MLA backend the registry bound to the contract, the gate returned `{}`, `make_dummy_batch` passed no metadata, and `init_forward_metadata` raised "bound to the Cache contract but received no paged cache metadata". `capture()` caught it, logged a warning, and the model family ran eager prefill for the whole process.

Visible on main in nightly run 33132689047, job
`eval-kimi-k2.5-nvfp4-eagle3-aime25`: four ranks warn, the bucket sweep stops at 0/40, and the job still passes. In the same run `eval-deepseek-v4-flash-gsm8k` and `eval-minimax-m3-nvfp4-aime25` capture 81 buckets each.

## What

`AttentionBackend.consumes_cache_metadata` asks the question the gate means: does a forward have to carry paged cache metadata for this backend? True when it advertises `uses_cache_groups` or when the registry has bound it to the Cache contract. `uses_cache_groups` keeps its value and its other meaning -- `cuda_graph_wrapper` still uses it for draft per-group table distribution.

Letting MLA reach this code at all exposed three more things:

* **Capture tables never use the reserved page.** They used to be filled with page 0 unless a backend asked otherwise, on the reasoning that a dummy batch reads nothing. It writes: attention is an eager break, so the latent KV store really runs, and adversarial review measured a K3 hybrid capture putting 2048 of 4096 write locations into page 0 -- the page that padding and table holes read expecting zeros. Every group now gets page 1, which removes the branch rather than adding one.
* **`cache_active_pages_must_be_real` named the wrong thing** and decided two unrelated questions at once. What actually differs per backend is the unit its capture table is addressed in, so it is now `capture_table_in_block_granularity`, and the fill is unconditional.
* **Delegating wrappers answer for their child.** `DSABackend` forwards `mark_cache_contract` to `_dense_backend` but kept its own flags, and `InklingAttnBackend` resolved one of the pair only through a `__getattr__` fallthrough.

Capture failure now stops the boot instead of degrading to eager, which is what let this bug live: the warning was indistinguishable from the families that are deliberately eager. A family that cannot capture belongs in `ModelExecutor`'s `disable_prefill_graph` condition, declared up front. This removes code rather than adding it -- the broad handler and the world-agreement all-reduce existed only to agree on degrading.

`--disable-prefill-graph` no longer changes what the dummy batch carries. `_autotune` shares `make_dummy_batch`, and the tables now follow from the backend's own contract rather than from a capture-only flag, so a contract-bound MLA backend is profiled on the path it will serve on. The graph-off arm below boots, tunes and serves normally. Backends that already advertised cache groups are untouched in both states.

Also corrects the bucket-ladder docstring, which claimed padded tails stay within ~12.5% anywhere. They do not below ~128 tokens, where the 16-token floor dominates: 17 tokens pad to 32.

## Testing

- `test/runtime/test_prefill_graph.py`: the failure policy, wrapper delegation for both wrappers, the reserved-page contract, and the disabled-graph path. Mutation testing during review found production rules whose sabotage killed no test; each now has one.
- The adjacent attention, cache and execution suites.
- `pre-commit run --all-files` clean.

End to end on 4xGB200, frozen tree, control arm on `origin/main`:

| arm | verdict | refusing | buckets |
| --- | --- | --- | --- |
| clean main | eager fallback | 8 | stuck 0/40 |
| this change | ready | 0 | 40/40 |

Also ready: Kimi-K3 (the hybrid path, whose capture geometry this moves), and Kimi-K2.5 under `TOKENSPEED_CACHE_DEBUG=1`, which turns on the null-page validators that reject the old tables.

## Performance and quality

Benchmarked on 4xGB200, Kimi-K2.5-NVFP4 with the CI eval server config. Two arms on one node, same tree, same weights, same client and seed, differing only by `--disable-prefill-graph`; the graph arm reports 40/40 buckets captured and the control arm reports none.

TTFT medians over 14 prompt lengths from 16 to 2000 words:

| | graph | eager | speedup |
| --- | --- | --- | --- |
| median | 86.9 ms | 181.0 ms | 2.13x |
| slowest case | 90.9 ms | 182.6 ms | 2.01x |
| fastest case | 58.2 ms | 153.8 ms | 2.64x |

The absolute gain is near-constant at 92-122 ms across the whole range, which is per-layer launch overhead leaving the critical path rather than anything that scales with the prefill. The ladder's padding does not overtake it: the 17-token case, which pads to 32, is the largest win.

Quality, same two arms:

* GPQA-diamond via the shipped CI config, all 198 questions: 0.8586 with the graph against 0.8232 without. One draw per arm cannot separate that gap from run-to-run variance, and no speedup in accuracy is claimed -- the hypothesis under test was a regression, and it is not there.
* Greedy completions are byte-identical between the two arms at all nine probed lengths, including 1, 17 and 2000. Replay equivalence had been argued from the code until now, not measured.

Stress, each arm, zero failures and zero server-side errors on both: 3000 requests at concurrency 64 over lengths 1-3000 words, which crosses the top bucket and so alternates captured and eager prefills throughout; 600 prefix-cache-hit requests whose extend tails are a few tokens; and 200 connections dropped mid-forward followed by a recovery batch.

## Summary

The prefill graph's capture gate asked `uses_cache_groups`, which MLA answers
inverted, so a contract-bound MLA target was handed a dummy batch with no cache
metadata, refused it, and the whole model family fell back to eager prefill
behind a warning indistinguishable from the families that are deliberately
eager. The gate now asks the question it means, capture failure stops the boot
instead of degrading silently, and capture tables no longer write into the
reserved page.

## Test Plan

Focused and adjacent suites (`test/runtime/test_prefill_graph.py` plus the
attention, cache-group and execution suites), `pre-commit run --all-files`, and
mutation testing over the production rules this change adds -- every sabotage
that survived review now has a test that kills it.

On hardware, 4xGB200, two arms on one node differing only by
`--disable-prefill-graph`, same tree, same weights, same client and seed:
capture coverage (40/40 vs stuck at 0/40 on a clean `origin/main` control),
TTFT across 14 prompt lengths, GPQA-diamond via the shipped CI config, greedy
output equivalence between the two arms, and a stress pass that alternates
captured and eager prefills across the top bucket. Numbers above.

### Known gaps

* Multi-node families are not covered by the hardware runs here. Kimi-K3 has no
  single-node config, so its capture path is exercised only by CI.
* DP/EP is untested: padding inflates collective payloads there, and neither
  arm above ran with data or expert parallelism.

### Follow-up, not in this PR

A separate change removes the per-backend capture-width knobs entirely in
favour of one rule, deleting `capture_table_in_block_granularity` and the
base-class `max_num_pages` that this PR introduces. It is held back until its
own validation is complete: sizing sliding-window groups by their window --
which looks correct, and which several reviewers independently recommended --
makes Inkling capture die with "extend write locations out of table bounds",
because capture fabricates one extend across the whole extent and derives a
write column for every position in it. That is a capture-side question, not the
decode-side one the existing width helper answers.


nperrin-fr:
@lightseek-bot Please hold, there is a failing case for some reason.

PR #1326, opened 2026-08-31
test: update k3 agentic bench

syuoni:


test/agentic_benchmark/kimi_k3/tokenspeed/configs/attn_tp8_moe_ep8.sh
```diff
@@ -19,7 +20,8 @@ exec ts serve \
     --speculative-num-draft-tokens 8 \
     --drafter-attention-backend tokenspeed_mla \
     --mm-encoder-tp-mode data \
+    --disable-kvstore \
```
syuoni: L2 cache seems broken. It crashes without `--disable-kvstore`.

@lightseek-bot  We need to fix it and enable it on CI. Currently the K3 CI cmds are with `--disable-kvstore`.

PR #1327, opened 2026-08-31
ci: make the Kimi-K3 TP8/EP1 jobs dispatchable

nirvedhmeshram:
Follow-up to #1283.

The two TP8/EP1 jobs added there are `manual`-trigger only, but `k8s-dispatch.yml` exposes its target as a hardcoded `type: choice` list and neither job was added to it. That combination leaves them **unreachable** — no per-commit trigger, and not selectable in the only workflow that can start them. They have never run.

Found while trying to fire the perf job by hand to confirm it works, which it turns out was not possible.

Registers both next to their TP8/EP8 counterparts:

```
  - test/ci/eval/kimi-k3-mxfp4-tp8ep8-evalscope-aime26-amd.yaml
+ - test/ci/eval/kimi-k3-mxfp4-tp8ep1-evalscope-aime26-amd.yaml
  ...
  - test/ci/perf/kimi-k3-mxfp4-tp8ep8-evalscope-random-4k-1k-mi35x.yaml
+ - test/ci/perf/kimi-k3-mxfp4-tp8ep1-evalscope-random-4k-1k-mi35x.yaml
```

The `amd` `runner_pool` option these jobs need already exists, so no other plumbing changes are required.

Note the thresholds in both EP1 jobs are still marked provisional in-file and have never been exercised on a runner. The first dispatches should be treated as calibration runs — particularly the perf reference, which is deliberately seeded loose because EP1 is slower than EP8 at the concurrency-1 point that job measures.

### DIFF
--- a/.github/workflows/k8s-dispatch.yml
+++ b/.github/workflows/k8s-dispatch.yml
@@ -31,6 +31,7 @@ on:
           - test/ci/eval/kimi-k2.5-nvfp4-eagle3-evalscope-ocr-bench.yaml
           - test/ci/eval/kimi-k2.5-mxfp4-eagle3-evalscope-aime25-amd.yaml
           - test/ci/eval/kimi-k3-mxfp4-tp8ep8-evalscope-aime26-amd.yaml
+          - test/ci/eval/kimi-k3-mxfp4-tp8ep1-evalscope-aime26-amd.yaml
           - test/ci/eval/minimax-m3-nvfp4-dspark-evalscope-aime26.yaml
           - test/ci/eval/minimax-m3-nvfp4-evalscope-aime25.yaml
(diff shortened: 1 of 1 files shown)


nirvedhmeshram:
Confirmed against the live API rather than by reading the workflow — dispatching the EP1 perf job against `main` today is rejected by GitHub itself:

```
$ gh workflow run k8s-dispatch.yml --ref main \
    -f yaml=test/ci/perf/kimi-k3-mxfp4-tp8ep1-evalscope-random-4k-1k-mi35x.yaml \
    -f runner_pool=amd

HTTP 422: Provided value 'test/ci/perf/kimi-k3-mxfp4-tp8ep1-evalscope-random-4k-1k-mi35x.yaml'
for input 'yaml' not in the list of allowed values
```

`workflow_dispatch` validates `type: choice` inputs against the workflow definition on the target ref, so the entry has to exist on `main` before either EP1 job can be started. Until this merges they are unreachable by any path.

PR #1332, opened 2026-08-31
perf(amd): widen gfx1250 MHA prefill M tile to 256 rows with 8 warps

nithinsubbiah:
## Summary

Raises `BLOCK_M` from 128 to 256 for the gfx1250 Gluon MHA prefill kernel, together with `num_warps` from 4 to 8. Up to 1.40x on long-sequence prefill, guarded so short and low-parallelism shapes keep today's tile.

The two knobs only work as a pair. Warps tile M only, so doubling them halves the registers each lane holds for the fp32 accumulator, which is what keeps the wider tile off the spill cliff. Separately, neither helps:

| | vs current default |
|---|---|
| 256 rows, 4 warps | 1.08x |
| 128 rows, 8 warps | 0.96x (slower) |
| 256 rows, 8 warps | 1.38x |

The wide tile halves the launch grid, and it doubles the causally-masked half of the diagonal `BLOCK_M x BLOCK_M` block, a waste proportional to `BLOCK_M / seqlen`. So `_select_m_tile()` widens only when `max_seqlen >= 1024` and the grid still covers the 256 CUs. Below either bound the narrow tile is up to 1.2x faster and is kept, following the shape-driven selection already used by gfx1250 decode and MoE.

`num_warps` and `num_buffers` now flow through `LaunchConfig` into the kernel, since `warp_bases` and the store layout assumed 4 warps. `num_buffers` stays at 2: the prologue/steady/epilogue pipeline needs double buffering, and a single buffer fails correctness.

## Test Plan

gfx1250 at a pinned 2400 MHz, bf16, GQA 32/8, head_dim 128. Every config validated against the per-sequence causal reference from `test_mha_prefill_lse` at its `8e-2` tolerance before being timed.

| shape | before | after | |
|---|---|---|---|
| 1x8192 | 1181 | 1651 | 1.40x |
| 4x4096 | 1136 | 1568 | 1.38x |
| 4x2048 | 983 | 1311 | 1.33x |
| 1x2048 | 734 | 872 | 1.19x |
| ragged 851/914/1053 | 510 | 575 | 1.13x |
| 1x1024 | 447 | 449 | guard holds |
| 6x512 | 480 | 482 | guard holds |
| 1x4096 GQA 8/1 | 629 | 625 | guard holds |
| 1x2048 GQA 8/1 | 289 | 288 | guard holds |

22 shapes in total, across bf16 and fp16, head_dim 64 and 128, GQA ratios 32/8, 32/32 and 8/1, sliding window, and ragged batching. No regressions.

### DIFF
--- a/tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx1250/attention/mha/prefill.py
+++ b/tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx1250/attention/mha/prefill.py
@@ -43,6 +43,8 @@ from tokenspeed_kernel_amd.ops.gfx1250.attention._common import (
 
 gfx1250 = gl.amd.gfx1250
 
+_GFX1250_NUM_CUS = 256
+
 
 @gluon.aggregate
 class AttentionConfig:
@@ -53,6 +55,7 @@ class AttentionConfig:
     BLOCK_M: gl.constexpr
     BLOCK_N: gl.constexpr
     NUM_BUFFERS: gl.constexpr
+    NUM_WARPS: gl.constexpr
     IS_FP8: gl.constexpr
     HAS_SINK: gl.constexpr
     HAS_LSE: gl.constexpr
@@ -80,6 +83,7 @@ class AttentionConfig:
         BLOCK_M,
         BLOCK_N,
         NUM_BUFFERS,
+        NUM_WARPS,
         IS_FP8,
         HAS_SINK,
         HAS_LSE,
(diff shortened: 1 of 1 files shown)


borontion:
btw @nithinsubbiah how much vgpr we are using for M 256 and 8 warps?

nithinsubbiah:
488 VGPRs per lane, no spills. Measured from `.vgpr_count` in the generated AMDGCN, head_dim 128, bf16:

| BLOCK_M | BLOCK_N | warps | VGPR/lane | spills |
|---|---|---|---|---|
| 128 | 64 | 4 | 616 | 0 |
| 128 | 64 | 8 | 302 | 0 |
| 256 | 64 | 4 | 906 | 0 |
| **256** | **64** | **8** | **488** | **0** |
| 256 | 128 | 8 | 512 | 474 |

488 is the number that makes the config work, because it sits just under the 512 budget for two waves per SIMD. Warps tile M only, so per-lane pressure for the fp32 accumulator scales as 1/num_warps: the same 256-row tile is 906 VGPRs at 4 warps, which fits without spilling but only ever gets one wave resident, and 488 at 8 warps, which gets two. That occupancy doubling is where the 1.38x comes from, and it is why neither knob helps alone (256/4w measures 1.08x, 128/8w measures 0.96x).

It is a narrow landing. 256x128 at 8 warps hits the 512 ceiling and then spills 474 registers, which costs about 4x rather than saving anything, so the accumulator width is the binding constraint on going wider.

Register allocation is identical whether the CU count is hardcoded or read from the device, so #1337 does not change any of the above.

PR #1337, opened 2026-09-01
test(amd): cover both gfx1250 prefill tile shapes

nithinsubbiah:
## Summary

Adds CI coverage for the 256-row / 8-warp prefill tile introduced in #1332. Nothing currently compiles it: the shapes in `test_mha_prefill_lse` select the narrow tile, and that test is deselected on the gfx1250 simulator runner in any case, so a regression in the wide tile's distinct WMMA and store layouts or its launch arguments would leave CI green.

Triggering the wide tile by shape needs a grid large enough to fill the device, which is more work than the simulator can run. `test_gluon_mha_prefill_gfx1250.py` therefore forces each tile shape on a small shape, and asserts the override took effect � without that check a missed override would silently exercise the other tile and report coverage it does not have. Both head sizes, full and sliding window.

A second test pins the two gates on the wide tile. Both are load-bearing: taking the 256-row tile when the sequence is too short, or when the grid does not fill the device, measured up to 1.2x slower.

Per review feedback the occupancy cutoff stays a hardcoded `_GFX1250_NUM_CUS = 256` rather than being read from the device, since CU count is fixed for this architecture. `prefill.py` is therefore unchanged from `main` and this PR is test-only.

## Test Plan

Runs in 8s on gfx1250 hardware, within the simulator budget:

| case | tile | result |
|---|---|---|
| narrow, d64 / d128, full + sliding | 128 / 4 warps | pass |
| wide, d64 / d128, full + sliding | 256 / 8 warps | pass |
| selector gates | � | pass |

Correctness is checked against the same per-sequence causal reference used by `test_mha_prefill_lse`, at its `8e-2` tolerance. `pre-commit run --all-files` is clean.

### DIFF
--- a/tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx1250/attention/mha/prefill.py
+++ b/tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx1250/attention/mha/prefill.py
@@ -27,6 +27,7 @@ optional sliding window, optional sinks, and optional natural-log LSE output.
 
 from __future__ import annotations
 
+import functools
 import math
 from typing import NamedTuple
 
@@ -42,8 +43,6 @@ from tokenspeed_kernel_amd.ops.gfx1250.attention._common import (
 )
 
 gfx1250 = gl.amd.gfx1250
-
-_GFX1250_NUM_CUS = 256
 
 
 @gluon.aggregate
@@ -704,8 +703,15 @@ class LaunchConfig(NamedTuple):
     grid: tuple[int, ...]
 
 
+@functools.lru_cache(maxsize=None)
+def _num_cus(device_index: int) -> int:
+    """Cached: querying it per launch costs ~2us, which is over 10% of a short
+    prefill, and a device's CU count cannot change within a process."""
+    return torch.cuda.get_device_properties(device_index).multi_processor_count
+
+
 def _select_m_tile(
-    *, batch_size: int, n_heads: int, max_seqlen: int
+    *, batch_size: int, n_heads: int, max_seqlen: int, num_cus: int
 ) -> tuple[int, int]:
     """Pick (BLOCK_M, num_warps) for prefill, measured on gfx1250.
 
@@ -722,13 +728,17 @@ def _select_m_tile(
         sequences pay a larger fraction of it.
 
     Below either bound the narrow tile wins by up to 1.2x, so keep it there.
+
+    num_cus comes from the running device rather than a constant, so a compute
+    partition exposing a fraction of the CUs compares its grid against the CUs
+    it actually has. The measured thresholds below come from a full 256-CU part.
     """
     wide_m, wide_warps = 256, 8
     narrow_m, narrow_warps = 128, 4
     if max_seqlen < 1024:
         return narrow_m, narrow_warps
     workgroups = batch_size * n_heads * triton_cdiv(max_seqlen, wide_m)
-    if workgroups < _GFX1250_NUM_CUS:
+    if workgroups < num_cus:
         return narrow_m, narrow_warps
     return wide_m, wide_warps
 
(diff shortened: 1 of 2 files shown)


tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx1250/attention/mha/prefill.py
```diff
@@ -704,8 +703,15 @@ class LaunchConfig(NamedTuple):
     grid: tuple[int, ...]
 
 
+@functools.lru_cache(maxsize=None)
+def _num_cus(device_index: int) -> int:
```
borontion, line 707: Why not hard code this to 256? This is scoped for gfx1250 anyway. In what environment we can have less than 256 WGP?

tokenspeed-kernel/test/ops/attention/test_gluon_mha_prefill_gfx1250.py
borontion, line 1: Maybe move this test under tokenspeed-kernel-amd

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx1250/attention/mha/prefill.py
nithinsubbiah: Compute partitioning. The device exposes `current_compute_partition` (SPX here) and has 7 `amdgpu_xcp_*` partition nodes live in sysfs, so a gfx1250 can present as a fraction of the physical part. Under CPX-style partitioning `multi_processor_count` reports the partition's CUs, not 256, and the grid comparison should be against the CUs the workload can actually reach.

Two supporting points. `gfx1250/attention/mla/decode.py:782` already reads `multi_processor_count` rather than hardcoding, so this matches the existing convention for the same architecture. And cached, the lookup costs 44ns against a 19us short prefill, so it is not buying safety with latency.

Being straight about the limits of that argument: `available_compute_partitions` is not readable for me on this box, so I cannot enumerate which modes are actually deployable, and I have not measured a sub-256-CU partition. If your deployments are always SPX then hardcoding is genuinely simpler and I am happy to switch it to a constant. The thresholds themselves were measured on a full 256-CU part either way, which is worth noting whichever we pick.

tokenspeed-kernel/test/ops/attention/test_gluon_mha_prefill_gfx1250.py
nithinsubbiah: Happy to move it, but flagging a conflict first: `AGENTS.md` says AMD Gluon kernel tests should live in `tokenspeed-kernel/test/` specifically to reuse the shared platform utilities and reference computations.

> AMD Gluon Kernel tests should live in `tokenspeed-kernel/test/` to reuse
> common platform utilities and reference computations.

Consistent with that, all six existing AMD Gluon tests are in `tokenspeed-kernel/test/ops/attention/` (`test_gluon_kda_prefill_gfx950.py`, `test_gluon_mla_decode_gfx950.py`, `test_gluon_dsa_amd.py`, and others), and `tokenspeed-kernel-amd/` currently has no test directory at all. This test also uses `is_cdna5()` from the shared `utils`, which is one of the platform utilities that guidance points at.

So moving it would create a new test tree, diverge from the six siblings, and need a different arch guard.

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx1250/attention/mha/prefill.py
nithinsubbiah: Correcting my earlier answer here: you were right, and I was not. CU count is fixed at 256 for gfx1250, so the partition case I argued for cannot arise and the device query buys nothing.

#1337 merged a few minutes before I could apply the change, so I have opened #1357 to hardcode it. That drops the per-launch query, the `lru_cache` and the `functools` import, and states the constant the way `kda/decode.py` does. Behaviour is identical since the query returned 256 anyway.

What I had was sysfs showing `current_compute_partition` and 7 `amdgpu_xcp_*` nodes, which told me the partition machinery exists but not that it can ever yield a gfx1250 with fewer than 256 CUs. I should have treated that as inconclusive rather than as support.

PR #1342, opened 2026-09-01
docs: use --prefix-granularity instead of deprecated --block-size

fuergaosi233:
## Summary

Docs still presented `--block-size` as the current `tokenspeed serve` flag. Argparse exposes `--prefix-granularity` as the real option and keeps `--block-size` only as a deprecated alias. Update the server, compatible-parameters, and model-recipe docs to the current flag. Alias is mentioned where this repo already documents deprecations that way (parameter tables). Launch recipes use `--prefix-granularity` only. Checkpoint `block_size` (DFLASH/DSPARK verify width) is left unchanged.

`docs/design/cache-concepts.md` already names `--prefix-granularity` and the alias; skipped.

## Reproduce

CLI definition (`python/tokenspeed/runtime/utils/server_args.py`):

```python
parser.add_argument(
    "--prefix-granularity",
    "--block-size",  # deprecated alias
    dest="prefix_granularity",
    metavar="PREFIX_GRANULARITY",
    type=int,
    default=ServerArgs.prefix_granularity,
    help="Scheduler prefix granularity in tokens: the identity "
    "boundary of cache reuse. (--block-size is a deprecated alias.)",
)
```

Docs on `main` still treated the alias as current:

- `docs/configuration/server.md`: `| `--block-size` | KV cache block size. |`
- `docs/configuration/compatible-parameters.md`: `| `--block-size` | KV cache block size. |`
- `docs/recipes/models.md`: launch lines `--block-size 128` and DeepSeek V4 `auto-sets `block_size=256` (pass `--block-size N` ...)`

One recipe already used `--prefix-granularity 128` (Qwen path).

## Test Plan

- [x] Quoted argparse dest/flags vs the three docs files on `ba1b1c71`
- [x] Replaced CLI `--block-size` with `--prefix-granularity` in recipes and flag tables
- [x] Left checkpoint `block_size` / event payload `block_size` alone
- [x] Alias noted only in the parameter tables, matching cache-concepts.md / CLI help

### DIFF
--- a/docs/configuration/compatible-parameters.md
+++ b/docs/configuration/compatible-parameters.md
@@ -31,7 +31,7 @@ TokenSpeed-specific behavior explicitly.
 | `--chat-template` | Chat template name or path. |
 | `--gpu-memory-utilization` | GPU memory fraction used for weights and KV cache. |
 | `--max-num-seqs` | Maximum concurrent sequences. |
-| `--block-size` | KV cache block size. |
+| `--prefix-granularity` | Scheduler prefix granularity in tokens. `--block-size` is a deprecated alias. |
 | `--enable-prefix-caching` | Enable prefix cache reuse. |
 | `--no-enable-prefix-caching` | Disable prefix cache reuse. |
 | `--enforce-eager` | Disable device-graph execution (CUDA Graph on CUDA, ACL Graph on NPU). |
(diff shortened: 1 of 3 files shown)


fuergaosi233:
Closing: submitted from the wrong GitHub account. Will reopen from imitater-dou.

PR #1353, opened 2026-09-01
perf(kimi-k3): reduce the MoE tail's joined partials in symmetric memory

nirvedhmeshram:
## Problem

The K3 MoE tail's join tier concatenates the routed and shared partials into one
operand and hands that to `all_reduce`. The backend can only select its
symmetric, in-place reduction on a *tuple* of tensors it allocated —
`AutoBackend.all_reduce` consults `can_reduce_outputs` on its tuple branch only —
so a single packed operand always took the staged path, copying both partials
into the collective's buffer before reducing them.

This is invisible from the call site: the staged path is correct, just slower,
and no amount of allocating the packed operand from symmetric memory changes the
selection. The operand *shape* is what routes it.

## Change

Acquire the `(routed, shared)` pair from the collective, point the producers at
it, and reduce the pair. This is the shape the EP-only layout already uses via
`latent_moe_expert_shared_all_reduce`; the TP layout now reaches the same kernel.

The packed lane remains the fallback for backends offering no producer-direct
memory. The tail verifies both partials actually landed in the acquired buffers
before reducing there — a producer that cannot write in place returns its own
tensor, and reducing a pair where only one side was produced in place would
silently drop the other's contribution.

Acquisition is gated on a new `can_acquire_all_reduce_outputs` probe, because
`acquire_all_reduce_outputs` always succeeds — it falls back to ordinary
allocations the collective must still stage — so taking those unconditionally
would cost a buffer without buying the kernel. The probe must answer identically
on every rank: a group split over which collective the tail runs would hang
rather than fail, so that requirement is documented on the base method.

## Measurements

8×MI355X, Kimi-K3 tp8/ep1, against the same commit without this change:

| | before | after |
|---|---|---|
| conc 1 | 60.41 tok/s | **64.17** |
| conc 8 | 271.74 tok/s | **291.97** |
| conc 16 | 452.96 tok/s | **489.75** |
| median TPOT @ conc 1 | 16.15 ms | **15.18** |

Decode trace, same shape both sides (63 steps, ~113.7k kernel events):

| | before | after |
|---|---|---|
| MoE-tail collective | `iris_stage_one_shot_allreduce` ×6426 | `iris_reduce_symmetric` ×5796 + 630 staged |
| mean / median | 67.91 / 45.04 µs | **24.07 / 13.16 µs** |
| total kernel dispatch | 1432.89 ms | **1146.00 ms** |

The 5796 calls that moved are exactly 92 MoE layers × 63 steps — every MoE layer
takes the symmetric path, with no silent fallbacks. The 630 remaining staged
calls are a different site, present identically in both traces.

## Tests

`test_fused_lane_ar_symmetric_matches_reference` drives the new regime against
the same reference the other tiers use. It is the tier variant most able to
disagree while everything else passes, since it swaps both the operand shape and
the summing kernel; it also pins the order the tail consumes the pair, which
transposition would otherwise leave plausibly shaped.

Run with `torchrun --standalone --nproc-per-node=8 -m pytest -q
test/runtime/test_k3_moe_tail_equivalence.py`.

### DIFF
--- a/python/tokenspeed/runtime/distributed/comm_backend/auto.py
+++ b/python/tokenspeed/runtime/distributed/comm_backend/auto.py
@@ -161,6 +161,31 @@ class AutoBackend(CommBackend):
     def prepare_all_reduce_lane(self, group: Group, hidden_dim: int) -> bool:
         return self._trtllm_ar.ensure_group_lane(group, hidden_dim)
 
+    def can_acquire_all_reduce_outputs(
+        self,
+        shapes: tuple[tuple[int, ...], ...],
+        like: torch.Tensor,
+        group: Group,
+        op=None,
+    ) -> bool:
+        """Whether ``acquire_all_reduce_outputs`` returns producer-direct memory.
+
+        Mirrors the routing in ``acquire_all_reduce_outputs`` below: the cases
+        that fall through to ``super()`` there get ordinary allocations, so they
+        answer False here.
+        """
+        if (
+            self._force_deterministic_rsag()
+            or self._group_spans_nodes(group)
+            or self._trtllm_ar.has_trtllm_ar(group)
+        ):
+            return False
+        if current_platform().is_amd:
+            return self._triton_ar.can_acquire_outputs(shapes, like, group, op=op)
+        return self._triton_ar.can_acquire_all_reduce_outputs(
+            shapes, like, group, op=op
+        )
+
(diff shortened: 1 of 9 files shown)


nirvedhmeshram:
closing this as superseded by https://github.com/lightseekorg/tokenspeed/pull/1358

PR #1360, opened 2026-09-02
Add DeepSeek V4 Gluon sparse attention

aditithanekar:
# Summary
 Adds a gfx950 Gluon sparse-attention path for DeepSeek-V4 prefill selected attention.

 This replaces the previous external benchmarking adapter with an in-repo implementation under:

  - `tokenspeed.runtime.layers.attention.deepseek_v4.gluon_sparse_attn`

  The default path remains unchanged. The new kernel is selected only when explicitly enabled and shape guards pass; otherwise the backend falls back to `dsv4_prefill`.

  ## Implementation

  - A production-focused Gluon sparse-attn kernel under:
    - `tokenspeed.runtime.layers.attention.deepseek_v4.gluon_sparse_attn`
  - Adds a dispatch adapter from TokenSpeed's selected-attention ABI:
    - `q: [tokens, heads, 512]`
    - `kv: [1, rows, 512]`
    - `indices: [tokens, topk]`
    - `lens: [tokens]`
  - Forwards real `topk_lens` to the Gluon wrapper.
  - Keeps fallback to `dsv4_prefill`.
  - Uses the `BLOCK_K=32` specialization for model-scale DeepSeek-V4 prefill shapes.
  - Removed experimental kernels from this PR.

## Benchmarking
  Setup: DeepSeek-V4-Flash, TP8, streaming EvalScope random workload, `ignore_eos=true`.

- 8K/1K: `prompt=8192`, `output=1024`, `number=concurrency`, `warmup-num=0.2`.
- 50K/500: `prompt=50000`, `output=500`, `number=concurrency`, `warmup-num=0`.
- 50K c1/c2/c4/c8 used `--max-total-tokens 524288 --max-num-seqs 8`.
- 50K c16 used `--max-total-tokens 1048576 --max-num-seqs 16`.

Compared variants:

- **TokenSpeed default**: built-in `dsv4_prefill` path.
- **AITER-main Triton**: `aiter/ops/triton/_triton_kernels/attention/sparse_attention_dsv4.py`, wrapped to TokenSpeed dense topk/lens ABI.
- **This PR**: native TokenSpeed Gluon sparse-attn path.

### 50K/500 long-context result

| Concurrency | Latency vs TokenSpeed default | Output tok/s gain vs TokenSpeed default | Latency  vs AITER-main Triton | Output tok/s gain vs AITER-main Triton |
| ---: | ---: | ---: | ---: | ---: |
| 1 | -3.99% | +4.16% | -8.49% | +9.28% |
| 2 | -6.61% | +7.07% | -13.04% | +14.98% |
| 4 | -5.21% | +5.48% | -11.80% | +13.37% |
| 8 | -10.77% | +11.99% | -17.52% | +21.16% |
| 16 | -11.43% | +12.85% | -21.98% | +28.06% |

Raw 50K/500 output throughput:

| Concurrency | TokenSpeed default | AITER-main Triton | This PR |
| ---: | ---: | ---: | ---: |
| 1 | 18.17 tok/s | 17.32 tok/s | 18.93 tok/s |
| 2 | 27.42 tok/s | 25.53 tok/s | 29.36 tok/s |
| 4 | 39.26 tok/s | 36.53 tok/s | 41.41 tok/s |
| 8 | 47.55 tok/s | 43.95 tok/s | 53.25 tok/s |
| 16 | 54.86 tok/s | 48.34 tok/s | 61.91 tok/s |

### 8K/1K non-regression check

8K/1K is neutral to slightly positive through concurrency 8 and roughly tied at concurrency 16.

Raw 8K/1K output throughput:

| Concurrency | TokenSpeed default | AITER-main Triton | This PR |
| ---: | ---: | ---: | ---: |
| 1 | 27.51 tok/s | 27.74 tok/s | 27.98 tok/s |
| 2 | 46.49 tok/s | 46.62 tok/s | 46.93 tok/s |
| 4 | 87.28 tok/s | 88.14 tok/s | 88.95 tok/s |
| 8 | 95.59 tok/s | 95.10 tok/s | 96.86 tok/s |
| 16 | 104.13 tok/s | 103.77 tok/s | 103.85 tok/s |

8K/1K latency note: this PR is slightly lower latency than TokenSpeed default through c8 in this run. At c16, latency is roughly tied/slightly higher (+0.91% latency), while TPOT is slightly better.

This run uses each row is one measured wave (`number == concurrency`), so this is a comparative smoke/perf sweep.

------------------------------------------------------
  ## Config

  Enable with:

  ```TOKENSPEED_DSV4_SELECTED_ATTN_IMPL=gluon_sparse```

  Optional threshold:

  ```TOKENSPEED_DSV4_GLUON_SPARSE_ATTN_MIN_TOKENS=8192```

  ## Validation

  Unit tests:

 ``` PYTHONPATH=python pytest -q test/runtime/test_deepseek_v4_gluon_sparse_attn.py```


python/tokenspeed/runtime/layers/attention/backends/deepseek_v4.py
```diff
@@ -40,6 +40,9 @@
 from tokenspeed.runtime.layers.attention.deepseek_v4.metadata import (
     DeepseekV4ForwardMetadata,
 )
+from tokenspeed.runtime.layers.attention.deepseek_v4.gluon_sparse_attn.dispatch import (
```
borontion: you need to register the gluon kernel under the shared attention api (`dsv4...`) and then use it. do not expose the gluon - which is an implementation detail - to the runtime.

python/tokenspeed/runtime/layers/attention/deepseek_v4/gluon_sparse_attn/dispatch.py
borontion, line 1: move this to tokenspeed-kernel-amd

python/tokenspeed/runtime/layers/attention/deepseek_v4/gluon_sparse_attn/dispatch.py
```diff
@@ -0,0 +1,111 @@
...
+
+def deepseek_v4_selected_attn_impl() -> str:
+    """Return the configured DeepSeek-V4 prefill selected-attention impl.
+
+    ``default`` preserves TokenSpeed's built-in kernel path. ``auto`` and
+    ``gluon_sparse`` enable the native gfx950 Gluon sparse-attention path when
+    shape guards pass.
+    """
+
+    impl = os.environ.get("TOKENSPEED_DSV4_SELECTED_ATTN_IMPL", "").strip().lower()
```
borontion: these backend routing should be handled by tokenspeed-kernel selection logic instead of env variables.

python/tokenspeed/runtime/layers/attention/deepseek_v4/gluon_sparse_attn/sparse_attn.py
```diff
@@ -0,0 +1,417 @@
...
+    valid0 = (k_pos < topk_len) & (k_pos < active_topk_len)
+    kv_off0 = (
+        dim_kv[:, None] * stride_kvd
+        + gl.where(valid0, index0, 0)[None, :] * stride_kvn)
+    cdna4_async.buffer_load_to_shared(kv_smem.index(0), kv, kv_off0)
+    cdna4_async.commit_group()
+
+    # Stage indices two tiles ahead and KV one tile ahead.
+    for i in tl.range(0, num_iters - 2):
+        if BLOCK_K == 64:
```
borontion: it would be better to split the whole kernel body (prologue+hot loop+epilogue) into 2 version one for `BLOCK_K = 64` and another one for other cases because these 2 pipelines look quite different.

python/tokenspeed/runtime/layers/attention/deepseek_v4/gluon_sparse_attn/sparse_attn.py
borontion, line 1: do note mask and where operations sometime can generate lots of redundant instructions. so please double check to make sure all mask/where here is necessary for the target shape

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/attention/dsv4/prefill.py
```diff
@@ -26,10 +26,17 @@
...
 import torch
 from tokenspeed_kernel_amd._triton import gl, gluon, tl, triton
+from tokenspeed_kernel_amd.ops.gfx950.attention.dsv4.sparse_prefill import (
+    gluon_dsv4_sparse_prefill_gfx950,
+)
 
 __all__ = ["gluon_dsv4_prefill_gfx950"]
 
 
+def _use_sparse_prefill(q: torch.Tensor, indices: torch.Tensor) -> bool:
```
borontion, line 36: I am wondering if this check can cover all test case? if so we should just default to use it instead of keeping the current gluon implementation.

can you check at least our current pro and flash test this is will always be true?

https://github.com/lightseekorg/tokenspeed/blob/main/test/ci/eval/deepseek-v4-flash-mtp-evalscope-gsm8k-amd.yaml
https://github.com/lightseekorg/tokenspeed/blob/main/test/ci/eval/deepseek-v4-pro-evalscope-gsm8k-amd.yaml

aditithanekar: I checked both current AMD eval configurations. The condition is not always true:

- DeepSeek-V4-Flash uses 64 global heads with TP2, resulting in H32 on each rank.
- DeepSeek-V4-Pro uses 128 global heads with attention TP8, resulting in H16 on each rank.

GFX950 preserves both H16 and H32 rather than padding them to H64, so neither job selects the new sparse implementation. Both satisfy the selected-width requirement, but fail the q.shape[1] in (64, 128) condition.

For now, we still need the existing Gluon implementation as the fallback. The new sparse implementation remains the default only for its supported H64/ H128 shapes. The Flash and Pro EvalScope jobs passed at 94.8% and 93.2%, respectively, but those runs validate the fallback paths rather than the new sparse path.

borontion (dismissed):
please fix the conflict. thanks.
also need to make sure nvidia kernel ut can pass: https://github.com/lightseekorg/tokenspeed/actions/runs/34532472037/job/103162334047?pr=1360

borontion (approved):

PR #1376, opened 2026-09-02
[DO NOT MERGE] ci(kimi-k3): GSM8K pair to compare the two MoE placements

nirvedhmeshram:
> **DO NOT MERGE.** Opened so the two new task files can be dispatched — a YAML
> that exists only in a PR is discoverable via `yaml=all` plus `match`. Whether
> this pair is worth keeping permanently depends on what it finds.

## Question

Does A8W4 (fp8 activations, tp8ep1) cost accuracy against A16W4 (bf16
activations, tp8ep8)?

## Why AIME26 cannot answer it

It runs **30 problems at temperature 1.0**. Sampling noise alone is ±1.6
problems, and separating a 3-point gap at 80% power needs roughly **1200
problems per arm** — about 40 repeats of the set.

Six runs across both placements so far:

| arm | scores |
|---|---|
| tp8ep8 DSpark | 0.900, 0.933, 0.933 |
| tp8ep1 | 0.867, 0.900, 0.900 |

The arms overlap and the difference is 0.78 sigma — equally consistent with a
real gap and with none. Two consecutive runs of the *same* tp8ep1 job scored
0.867 then 0.900, which is the noise floor making the point directly.

## Why GSM8K can

| benchmark | n | SE(diff) | 3.3pp gap | sampling |
|---|---|---|---|---|
| aime26 | 30 | 7.75pp | 0.43σ | temperature 1.0 |
| gpqa_diamond | 198 | 2.56pp | 1.29σ | temperature 1.0 |
| **gsm8k** | **1319** | **0.85pp** | **3.89σ** | **greedy** |
| mmlu | 14042 | 0.43pp | 7.74σ | greedy |

GSM8K fixes both defects at once. The sample is 44× larger, and it runs greedy
(`do_sample: false, temperature: 0`) so each arm is a deterministic function of
its weights — repeated runs of one arm agree exactly, and every difference
between arms is attributable rather than sampled. MMLU is stronger still but
costs far more runtime for a question GSM8K already settles.

## Design

The two files differ **only in `--ep-size`**. Everything else — server flags,
generation config, batch size, dataset args — is byte-identical, so the
comparison isolates the placement.

Neither carries a `score_threshold`. The quantity of interest is the *difference
between the arms*; a per-arm bar reports on one at a time and says nothing about
it. Manual trigger, since this answers a specific question rather than guarding
every commit.

## Also

Registers two DSpark tp8ep8 entries missing from the K8s dispatch choice list.
`test_k8s_dispatch_lists_every_supported_ci_yaml` requires every AMD-backed CI
YAML to appear there, and it already fails on `main` without them. Same two
entries are also added in #1374; whichever lands first, the other rebases.

## Result: GSM8K

| cap | EP8 (A16W4) | EP1 (A8W4) | gap | |
|---|---|---|---|---|
| 256 (truncated) | 73.6% | 70.7% | 2.9 pp | 1.66σ |
| **4096 (clean)** | **98.0%** | **97.5%** | **0.5 pp** | **0.87σ** |

**No meaningful accuracy difference.** Seven problems out of 1319, at 0.87σ,
against a detection floor of 1.14 pp — the effect is bounded, not merely
unfound.

The first attempt used a 256-token cap inherited from the non-reasoning GSM8K
jobs. Mean output was 206 tokens against that cap, so responses were being cut
off before reaching an answer. Raising the cap moved both arms up ~25 points and
collapsed the gap from 2.9 pp to 0.5.

The mechanism is worth recording, because it is the trap this pair was built to
avoid. tp8ep1 emits about **4% more output** (260 vs 250 tokens). Under a
binding cap that verbosity pushes proportionally more of its responses past the
ceiling, manufacturing a 38-problem "accuracy gap" out of a length difference.
Above the cap it costs nothing.

The verbosity itself looks real and is unexplained. One plausible mechanism:
perturbation is symmetric in logit space but the outcome space is not — there
are many ways to continue reasoning and one way to stop — so noise at a near-tie
biases toward continuing. Testing that needs per-sample capture, which the
artifacts do not retain. Practically it only matters under a tight `max_tokens`,
where tp8ep1 wants slightly more headroom than tp8ep8 for the same task.

## GPQA-Diamond

GSM8K puts both arms near ceiling, so it bounds the difference on *easy*
material. The GPQA pair extends the comparison to graduate-level reasoning at
lower power (198 problems, ~5 pp detection floor against GSM8K's 1.1). Running;
results to follow.

## Caveat on interpretation

### DIFF
--- a/.github/workflows/k8s-dispatch.yml
+++ b/.github/workflows/k8s-dispatch.yml
@@ -34,6 +34,9 @@ on:
           - test/ci/eval/kimi-k2.5-mxfp4-eagle3-evalscope-aime25-amd.yaml
           - test/ci/eval/kimi-k3-mxfp4-tp8ep8-evalscope-aime26-amd.yaml
           - test/ci/eval/kimi-k3-mxfp4-tp8ep1-evalscope-aime26-amd.yaml
+          - test/ci/eval/kimi-k3-dspark-mxfp4-tp8ep8-evalscope-aime26-amd.yaml
+          - test/ci/eval/kimi-k3-mxfp4-tp8ep1-evalscope-gsm8k-amd.yaml
+          - test/ci/eval/kimi-k3-mxfp4-tp8ep8-evalscope-gsm8k-amd.yaml
           - test/ci/eval/minimax-m3-nvfp4-dspark-evalscope-aime26.yaml
           - test/ci/eval/minimax-m3-nvfp4-evalscope-aime25.yaml
           - test/ci/eval/qwen3.5-122b-a10b-nvfp4-epd-1e1p2d-evalscope-ocr-bench.yaml
@@ -48,6 +51,7 @@ on:
           - test/ci/perf/kimi-k2.5-nvfp4-evalscope-agentic.yaml
           - test/ci/perf/kimi-k3-mxfp4-tp8ep8-evalscope-random-4k-1k-mi35x.yaml
           - test/ci/perf/kimi-k3-mxfp4-tp8ep1-evalscope-random-4k-1k-mi35x.yaml
+          - test/ci/perf/kimi-k3-dspark-mxfp4-tp8ep8-evalscope-random-4k-1k-mi35x.yaml
           - test/ci/perf/qwen3.5-397b-a17b-nvfp4-evalscope-agentic-b200-8gpu.yaml
           - test/ci/perf/qwen3.5-397b-a17b-nvfp4-evalscope-agentic.yaml
           - test/ci/perf/qwen3.5-397b-a17b-nvfp4-evalscope-longctx.yaml
--- a/.github/workflows/k8s-dispatch.yml
+++ b/.github/workflows/k8s-dispatch.yml
@@ -36,6 +36,8 @@ on:
           - test/ci/eval/kimi-k3-mxfp4-tp8ep1-evalscope-aime26-amd.yaml
           - test/ci/eval/kimi-k3-dspark-mxfp4-tp8ep8-evalscope-aime26-amd.yaml
           - test/ci/eval/kimi-k3-mxfp4-tp8ep1-evalscope-gsm8k-amd.yaml
+          - test/ci/eval/kimi-k3-mxfp4-tp8ep1-evalscope-gpqa-diamond-amd.yaml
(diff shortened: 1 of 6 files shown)


nirvedhmeshram:
Closing — these jobs did their work and the conclusion is folded into #1374.

**What they established.** The question was whether A8W4 (fp8 activations, the
tensor-parallel placement) costs accuracy against A16W4 (bf16, expert-parallel).
AIME26 could not answer it: 30 problems at temperature 1.0 gives a ±1.6-problem
noise floor, and six runs across both placements produced 0.867–0.933 with the
arms overlapping.

| test | n | tp8ep8 | tp8ep1 | gap | sigma |
|---|---|---|---|---|---|
| AIME26 | 30 | 0.922 | 0.883 | 3.9 pp | 0.78 |
| GSM8K | 1319 | 0.980 | 0.975 | 0.5 pp | 0.87 |
| GPQA-Diamond | 198 | 0.778 | 0.748 | 3.0 pp | 0.70 |

Every comparison leans the same way, so a small real deficit is more likely than
not (Stouffer p = 0.087) — but it is **bounded under 1.07 pp** by GSM8K, and it
cannot be measured more precisely at any affordable cost: resolving 0.5 pp needs
~12,300 samples per arm at GSM8K's accuracy, or ~80,000 at MMLU's. MMLU in full
would move the bound to 0.84 pp for hours of 8-GPU time per arm and change no
decision.

**Along the way, two things worth keeping.** A 256-token cap inherited from the
non-reasoning GSM8K jobs truncated a reasoning model — mean output 206 against
the cap — and manufactured a 2.9-point "accuracy gap" out of a 4% verbosity
difference between the arms. Raising the cap moved both arms up ~25 points and
collapsed the gap to 0.5. Any future eval pair comparing two configurations
needs headroom above the reasoning budget, or it measures truncation.

And the obvious remedy for fp8 error does not apply: per-token scaling gains
**1.00x** on e4m3 against int8's 3.64x, because e4m3's exponent bits carry the
dynamic range. The 2.65% error is intrinsic to three mantissa bits.

**Why not keep the jobs.** They answered a specific question rather than
guarding a contract, the answer will not change without a kernel change, and
each costs an 8-GPU runner. Easy to resurrect from this branch if the question
returns.

PR #1380, opened 2026-09-03
[WIP] feat(memory): reserve the CUDA-graph pools in the KV cache budget

nperrin-fr:
profile_available_cache_memory_bytes sizes the KV cache from driver free memory at call time, and both CUDA-graph pools are allocated after that out of the same remainder. Nothing told the budget they exist, so raising the prefill-graph ceiling OOMed during capture instead of shrinking the cache, and even when it fit the graphs ate the headroom the operator asked for: on Inkling-NVFP4 TP4 at utilization 0.95, Device free landed at 4.53 GB against the 9.22 GB that setting promises.

A throwaway probe now runs before the budget, captures the largest few entries of each graph family against a minimal arena and a private pool, measures driver-level deltas, tears everything down, and subtracts a projection:

    estimate = first + sum(marginals) + max(median(marginals), 0) * remaining

summed across families, because the prefill and decode pools are disjoint at runtime; only pools sharing one runtime allocation may be combined with a max.

The median is what keeps a one-off allocator growth step out of the rate. A step of 0.650 GB against a 0.046 GB mean appears on one 48-entry ladder; extrapolating the max over-projects it by 608% and the mean of two samples by 301%. Observations are charged at their net value, including negative ones -- a decode family sampled (-48, +26, +30, +26) MiB when a segment release landed on the pool-creating capture -- because the budget is derived from driver free memory. Only the extrapolated rate and the final estimate are floored at zero.

`--disable-cudagraph-memory-reserve` turns the probe off. The gate sits at the call site, so the disabled path builds no probe, arena, executor or captures and subtracts a literal zero.

Accuracy, cross-checked by differencing two boots rather than by instrumenting one: the pools cost 3.96 GB at the default Inkling ceiling and this projects 4.16 GB, +5%. At ceiling 8192, where main dies at bucket 21 of 56, the branch captures all 56 and serves. Device free returns to 9.53 GB on Inkling and 18.78 GB on Qwen3-8B, against promises of 9.22 and 18.43. With the flag off, KV bytes, Device free and max_total_num_tokens match main exactly on both.

Known limits: a growth step landing outside the sampled window is not bounded by the estimator; the probe leaves ~1 GB of lazy module loads, library workspaces and fragmentation that it cannot return, charged to the budget once; and the 3-parent probe arena is unexercised on paged-state families.

### DIFF
--- a/docs/design/cache-concepts.md
+++ b/docs/design/cache-concepts.md
@@ -174,6 +174,13 @@ to the kernel page tables that attention kernels consume (the
 happen at **one designated point**; beyond that point, kernels see physical
 page tables and nothing upstream sees them at all.
 
+`create_attn_components` must stay callable more than once for the same
+startup arguments: the CUDA-graph memory probe builds a throwaway arena before
+the real one. The backend overrides it derives from the model architecture are
+idempotent, and the one input they destroy -- the operator's own
+`attention_backend` -- is preserved in `original_attention_backend` on first
+resolution, so a second call resolves the same sub-backends as the first.
+
 Outside the mapping point, Python code should perceive `prefix_granularity`
 and `page_size` as little as possible. If a Python component needs either
 value, that is a design smell to justify, not a default to reach for.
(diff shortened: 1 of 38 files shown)


nperrin-fr:
reworking to cope with the recent refactoring

nperrin-fr:
do in #1509

PR #1399, opened 2026-09-04
perf(dflash2): Optimize dflash2 implementation

yweng0828:
### 1. Speculative semantics alignment
- For all dspark algorithms, the `block_size / dspark_block_size` is equal to `speculative-num-step`. 
- For dflash / dflash2, the `block_size` is equal to `speculative-num-step + 1`.

#### Kernel optimization:
- **Add sliding window `tokenspeed_mla` decode**. Both the `fp16` and `fp8` kernels take `window_left`: the KV walk starts at the window rather than at key 0, and the split-KV schedule and reduction follow, so cost tracks the window and not the cache. `window_left = -1 `compiles the kernel as it always did.
- **Route the draft's own projections through the measured decode-GEMV table.** The conv `kernel_projection` and fc move to the repo's `ReplicatedLinear`, which is what consults `routed_gemv`; the table gains DFlash2's TP=8 shapes (M = batch × block width, up to 64).
- Shard-local top-k for DFlash2 candidate selection. Instead of all-gathering whole 163840-wide logits rows, each rank takes its own `top-16` over its 20480-column shard and gathers only those. And packs the top-k values and ids into one all-gather.
- Single-pass radix kernel for the selector's shard top-k. Measured 3.1–4.2× faster.

#### Kernel fusion:
- Grouped conv fused into one Triton kernel. DFlash2's four per-layer grouped dynamic depthwise convolutions were a storm of small torch elementwise ops; one kernel now derives the tap mask from `program_id` and materializes no positions, coefficients, or padded shift.
- Fused MLA draft context-KV write, overlapped and pipelined. Every layer's latent down-projection stacks into one GEMM, followed by one `mla_latent_norm_rope_scatter` Triton launch (RMSNorm + RoPE + scatter into each layer's cache plane). The write runs on an aux stream overlapping the draft forward.
- Move the layer's reductions onto its norms, past the grouped conv. The conv is linear in its input, and its coefficients come from an already-reduced hidden, so the reduction commutes with it and lands where the next norm can absorb it.
- Greedy path walk fused into one Triton kernel. Seven torch gather/argmax/gather become one.

#### Misc:
- Drop the dummy-query rope, the per-step residual, and the per-request syncs. 
- Build the query-axis fold's per-request rows once per forward. 

#### agentic benchmark
- before optimization

| config                          | Conc. | Latency (tps/user) | Throughput (tps/gpu) | Approx Cache Hit | Decoded Tok/Iter |
|---------------------------------|-------|--------------------|----------------------|------------------|------------------|
| attn_tp8_moe_tp8_dflash2_main   | 1     | 130.38             | 1614.43              | 90.86            | 3.5171           |
| attn_tp8_moe_tp8_dflash2_main   | 2     | 91.58              | 2273.96              | 91.01            | 3.0462           |
| attn_tp8_moe_tp8_dflash2_main   | 4     | 64.31              | 2898.61              | 90.17            | 3.3228           |
| attn_tp8_moe_tp8_dflash2_main   | 8     | 40.82              | 3850.09              | 90.03            | 3.7336           |
| attn_tp8_moe_tp8_dflash2_main   | 16    | 22.87              | 4476.65              | 91.23            | 3.5553           |

- after optimization

| config | Conc. | Latency (tps/user) | Throughput (tps/gpu) | Approx Cache Hit | Decoded Tok/Iter |
|---|---|---|---|---|---|
| attn_tp8_moe_tp8_dflash2 | 1  | 225.73 | 2540.00  | 90.86% | 3.6930 |
| attn_tp8_moe_tp8_dflash2 | 2  | 146.84 | 3411.45  | 91.01% | 3.2879 |
| attn_tp8_moe_tp8_dflash2 | 4  | 95.79  | 4352.11  | 90.17% | 3.1266 |
| attn_tp8_moe_tp8_dflash2 | 8  | 63.98  | 5513.09  | 90.03% | 3.3803 |
| attn_tp8_moe_tp8_dflash2 | 16 | 39.14  | 7312.59  | 91.22% | 3.4736 |

Note: Why the Why is the performance of main so poor?
Flatten main‑path blocks into 8 rows → Triton kernel without split‑KV → 64 CTAs each serially traverse their scan range
→ Per‑layer cost = scanned tokens × 0.115 us/1k
→ Layer 5 uses full attention; its scan covers the full context and grows linearly with context length
→ Drafter overhead: 1.78 ms (9.2%) at ctx 900 → 4.96 ms (23.5%) at ctx 11.7k → ~11 ms (~41%) at ctx 57k
→ Per‑iteration multiplier: 1.065× → 1.23× → 1.65×
After optimization, we folded onto the query axis → CuteDSL splits KV into 32 chunks → Per‑layer latency stays constant at 9.1‑9.6 us, insensitive to context length

#### timeline
- before optimization: ~380 kernels with ~1700us
image

- after optimization: ~150 kernels with ~725us
image

#### Time breakdown in DFlash2
| category                          | main k | main us | main % | opt k | opt us | opt % | delta us |
|-----------------------------------|--------|---------|--------|-------|--------|-------|----------|
| draft MLA decode                  | 5.0    | 541.8   | 31.3%  | 10.0  | 57.8   | 7.5%  | -484.0   |
| unfused torch ops inside the layers | 183.7  | 403.3   | 23.3%  | 3.7   | 8.4    | 1.1%  | -394.9   |
| GEMM projections                  | 46.0   | 292.5   | 16.9%  | 46.0  | 304.4  | 39.7% | +11.8    |
| TP all‑reduce                     | 11.0   | 112.6   | 6.5%   | 11.0  | 87.9   | 11.5% | -24.7    |
| RMSNorm                           | 16.0   | 75.0    | 4.3%   | 16.0  | 75.4   | 9.8%  | +0.4     |
| GEMM split‑K reduction            | 15.5   | 48.8    | 2.8%   | 1.0   | 4.5    | 0.6%  | -44.3    |
| draft LM head                     | 1.0    | 48.3    | 2.8%   | 1.0   | 48.6   | 6.3%  | +0.2     |
| candidate top‑k / sort            | 13.0   | 45.8    | 2.6%   | 3.0   | 21.0   | 2.7%  | -24.8    |
| greedy path walk                  | 21.0   | 43.5    | 2.5%   | 2.0   | 6.6    | 0.9%  | -36.9    |
| torch fragments in the selector tail | 15.7   | 33.8    | 2.0%   | 10.7  | 23.8   | 3.1%  | -10.0    |
| candidate all‑gather              | 1.0    | 27.1    | 1.6%   | 1.0   | 17.0   | 2.2%  | -10.1    |
| rope + latent KV write            | 5.0    | 13.9    | 0.8%   | 5.0   | 13.8   | 1.8%  | -0.2     |
| prepare + block bookkeeping       | 8.0    | 13.4    | 0.8%   | 9.0   | 13.1   | 1.7%  | -0.2     |
| SiLU                              | 5.0    | 12.3    | 0.7%   | 5.0   | 17.2   | 2.2%  | +4.9     |
| codebook gather / cat             | 4.0    | 9.2     | 0.5%   | 4.0   | 10.2   | 1.3%  | +1.0     |
| embedding lookup                  | 1.0    | 6.2     | 0.4%   | 1.0   | 6.2    | 0.8%  | -0.0     |
| selector einsum                   | 1.0    | 2.5     | 0.1%   | 1.0   | 2.5    | 0.3%  | -0.0     |
| grouped conv (fused Triton)       | 0.0    | 0.0     | 0.0%   | 20.0  | 48.5   | 6.3%  | +48.5    |
| **total**                         | **352.9** | **1730.1** | 100% | **150.5** | **766.8** | 100% | **‑963.3** |

### Accuracy
AIME26
```
┌─────────┬───────────┬────────────┬──────────┬───────┬─────────┐
│ Model   │ Dataset   │ Metric     │ Subset   │   Num │ Score   │
├─────────┼───────────┼────────────┼──────────┼───────┼─────────┤
│ kimi-k3 │ AIME-2026 │ Accuracy   │ default  │    30 │ 96.7%   │
└─────────┴───────────┴────────────┴──────────┴───────┴─────────┘ 
```

## Next TODO
**The optimization of DFlash2 is not yet complete. We will conduct more detailed optimizations in the following PRs.**


yweng0828:
cc @Dogacel for comment and review.

python/tokenspeed/runtime/execution/drafter/dflash2.py
```diff
@@ -86,12 +104,28 @@ def wire_target(self, target_model) -> None:
...
+            dp_sampling_enabled=processor.dp_sampling_enabled,
+        )
 
     def _compute_candidates(
         self, hidden_states: torch.Tensor
     ) -> tuple[torch.Tensor, torch.Tensor]:
         if self.candidate_logits_processor is None:
             raise RuntimeError("DFlash2 must be wired to the target before drafting.")
+        if self.candidate_topk.enabled:
+            return self.candidate_topk(hidden_states)
```
Dogacel, line 128: Torch's `topk` returns `(unary_logits, candidate_ids)` but the `candidate_topk` returns it in the reverse order. Maybe we should update `VocabParallelTopK` to return the same output, so they are interchangable?

tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/routed_gemv.py
```diff
@@ -626,6 +675,38 @@ def _mark_warmed(backend: str, dev: int, m: int, n: int, k: int) -> None:
...
+# already reach by 4%. ``test/gemm_tuning/tune_splitk_tactic.py`` reproduces it.
+SPLITK_TACTIC_ROUTE: MappingProxyType[
+    tuple[int, int, int], tuple[int, int, int, int]
+] = MappingProxyType(
+    {
+        # DFlash2 draft at TP8, M = batch * 8. Left out on purpose:
+        # 2112x7168 is also the K3 target's replicated latent projection and
+        # already runs a dedicated fused kernel, and 20480x7168 is the head,
+        # whose call site is a bare matmul the router never sees.
+        (8, 1536, 1536): (64, 8, 4, 12),  # 1.98 vs ll_bf16 2.54 us (1.28x)
```
Dogacel: Those absolute numbers / speedups only relate to GB300s right?

test/ci/eval/minimax-m3-nvfp4-dspark-evalscope-aime26.yaml
```diff
@@ -25,9 +25,9 @@ server:
     --moe-backend flashinfer_trtllm
     --speculative-algorithm DSPARK
     --speculative-draft-model-path nvidia/MiniMax-M3-DSpark
-    --speculative-num-steps 7
+    --speculative-num-steps 8
     --speculative-eagle-topk 1
-    --speculative-num-draft-tokens 8
+    --speculative-num-draft-tokens 9
```
Dogacel, line 30: Interesting, I think M3 is trained with block size 8 instead of 9.

tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/routed_gemv.py
```diff
@@ -564,6 +565,54 @@
         # Unstable exact keys stay on the incumbent: (2,512,2560),
         # (4,640,2560), (17/21/23,2560,320), (1,6656,2560), and
         # (3/4,12800,2560). For 12800x2560, M=5..32 also stay on cuBLAS.
+        # TP8 Kimi-K3 DFlash2 block drafter; M = batch * block width, so the
+        # served M runs well past the M<=8 the other drafters stop at.
+        # (2112x7168) stays out: it is the target's replicated latent
+        # projection too, and already runs a dedicated fused kernel.
+        # 20480x7168 (the head) and 7168x35840 (the context fc) measured
+        # non-monotonic across M, so they keep the incumbent.
+        (5, 1792, 7168): "ll_bf16",  # 5.63 vs 7.75 us (1.38x)
```
Dogacel: Question, did you auto-generate those?

Dogacel (commented):
I couldn't review the actual mla / splitk kernels, but I think it looks good directionally. I've seen TP replication help throughput at high TP, especially for models with ultra-high interactivity (i.e. ~30B param models).

Also the block size / sliding window fix for DFlash family looks good. Setting SWA support per kernel backend makes sense rather than hard-coding per speculative decoding algorithm.

tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/routed_gemv.py
yweng0828: Yes, thanks for the reminder. I may consider removing these comments to avoid any misunderstanding.

tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/routed_gemv.py
yweng0828: Sorry, the agent did that for me.

test/ci/eval/minimax-m3-nvfp4-dspark-evalscope-aime26.yaml
yweng0828: Yeah, you could see the `1. Speculative semantics alignment.` and [here](https://github.com/vllm-project/vllm/blob/main/vllm/v1/worker/gpu/spec_decode/dspark/speculator.py#L5-L20).

dishengbin:
Changes on tokenspeed-mla LGTM, thanks.

PR #1400, opened 2026-09-04
fix(scheduler): bank a growth block per snapshot-state group at admission (local and remote)

dongjiyingdjy:
## Summary

A D-role remote admission materialized exactly one block per KDA snapshot-state group (the peer's endpoint snapshot) with `reserve_tokens = 0`: the shared `setSnapshotStatePrefillReserve(..., split_tail_tokens)` call ran for both prefill sources, and `split_tail_tokens` is always 0 on `Role::kD` because `shouldSplitFinalStateCheckpoint` is false there. At its first 128-token boundary every landed request therefore needed a fresh **empty** parent per state group. Once the pool held fewer empty parents than state groups no resident could progress, none finished, and none was retractable because their MLA headroom already covered their remaining generation (`ReserveCoversGeneration`). Result: a silent deadlock (`#running-req` stuck, no `Finish!`, no retraction, P-side bootstrap timeouts). Observed on Kimi-K3-NVFP4 1P1D (GB300, 2 P + 2 D nodes) at 48 and 64 concurrent 50-69K-token agentic prompts; reproduced three times.

Fix (`schedulePrefillFirstChunk` / `schedulePrefill`, unified per review): every snapshot-state group reserves `max(block_granularity, split_tail_tokens, decode_input_tokens)` on the admission that finishes shaping it -- the chunk that completes the prompt, the body that banks a split tail, or a remote landing (`reserveSnapshotStateGrowth`, replacing `setSnapshotStatePrefillReserve`; no role-specific branch). Every prompt length, every role and both prefill sources end with the endpoint block plus the block the first boundary crossing writes into, and the first decode step is consumable in place. Intermediate chunks and the round that spends a banked tail reserve nothing.

The same rule closes the one-block end state on the local path that the split-tail mechanism never covered: a prompt whose length is a multiple of the state-group granularity has no tail to split and used to materialize a single state block with zero spare (Fused, P bootstrap decode, D-role local recovery). The split-body path is unchanged (`max(g, r, d) == r` in block count and spare for `r >= 1`, `d <= g`). Capacity planning counts the growth block on every role: unchanged for prefix-cached local roles; one more state block per group for chunked D-role recovery and `disable_prefix_cache` roles.

Cost: one parent per state group per admitted request (3 parents, ~90 MB, on K3; ~6 % of a 57K-token request); on the local split-tail path the block count is unchanged. A full pool now back-pressures at admission (`#queue-req > 0`) instead of wedging.

Also: `docs/design/scheduler.md` §1.2 documents the D-role growth reserve.

## Test Plan

- New `tokenspeed-scheduler/python/tests/test_local_state_growth_reserve.py` (Fused role): six aligned one-block prompts fill the 32-page K3-shaped pool; unfixed, all six admit and wedge at their first decode step; fixed, admission back-pressures at four and everything completes. A second test asserts the completing admission owns two blocks per state group and the first decode reuses them.
- New `tokenspeed-scheduler/python/tests/test_d_role_state_reserve.py` (K3-shaped 4-group config, D role, P = 2 tokens/block, 32 pages): five one-block prompts admitted before landing fill the pool to 2 < 3 empty parents. Unpatched scheduler never produces another forward op (all 3 tests fail); patched scheduler queues and completes. Two more tests assert one extra block per state group on remote admission and that a landed request decodes its first step from its own reserve.
- gtests: `test_outside_event_handler.cpp` (D-role two-block shape, `PdLocalRecoveryCapacityTestSuite` 24 -> 20), `test_cache_scenarios.cpp` (`MambaSparsePrefillSuite` / `MambaOverlapRollingStateSuite` two-block end state, `MambaFusedRetractionDrainSuite` pool re-sized). Full suites: 63 python tests, 454 gtests pass.
- Live (GB300, Kimi-K3-NVFP4 1P1D, attn tp8 / moe ep8, 32K prefill chunks, D util 0.92 = 1989 parents, `max_tokens` 500, 64 sessions): conc 48 and conc 64 each complete 743/743 with 0 retractions and 0 P timeouts on the exact config that deadlocked unpatched (pool pinned at 1989/1989 becomes `#queue-req` up to 17). With the L2 host cache enabled on P and D: conc 48 = 107.8k tok/s, conc 64 = 94.0k tok/s. Unpatched + L2 still deadlocks at conc 64 within ~2 minutes.
- The same 4-line change backported to 4cb77148 runs conc 64 cleanly (101.1k tok/s, D util 0.98) where the unpatched old commit deadlocked twice.
- pre-commit hooks checked on the changed files: clang-format 18.1.8, black, isort, trailing-whitespace, end-of-file.

### DIFF
--- a/docs/design/scheduler.md
+++ b/docs/design/scheduler.md
@@ -95,6 +95,15 @@ Consequences worth knowing:
   an open optimization (it would lower `MaxSingleRequestTokens`, so it needs
   its own capacity baseline).
 
+- **Remote (D-role) admissions bank one growth block per state group.** A
+  D-role admission has no local tail to split; the peer lands only the endpoint
+  snapshot, so `schedulePrefillFirstChunk` (remote branch) reserves
+  `block_granularity` extra tokens in every `LatestSnapshot` group, i.e. one
+  more block for every prompt length. Without it each landed request needs a
+  fresh **empty** parent per state group at its first boundary, and a full pool
+  deadlocks silently because residents are retraction-exempt. Capacity planning
+  counts 2 state blocks per admitted request, as for a local prefill.
+
 ### 1.3 What bounds a single request
 
 `MaxSingleRequestTokens` is a **startup** bound computed by binary search over
--- a/tokenspeed-scheduler/csrc/scheduler/operations/forward.cpp
+++ b/tokenspeed-scheduler/csrc/scheduler/operations/forward.cpp
@@ -348,6 +348,7 @@ std::optional<fsm::SchedulePrefillFirstChunkEvent> Scheduler::schedulePrefillFir
         makeGroupDemands(tables, GroupDemand{.num_tokens = tokens_this_round, .reserve_tokens = admission_reserve});
     if (source == fsm::PrefillSource::kLocal) {
         makeSnapshotStatePrefillSparse(demands, config_.cache_groups, coordinator_, hit_tokens + tokens_this_round);
+        setSnapshotStatePrefillReserve(demands, config_.cache_groups, split_tail_tokens);
     }
 
     if (source == fsm::PrefillSource::kRemote) {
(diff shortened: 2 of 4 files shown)


dongjiyingdjy:
Addressed both review items in 0c7c5bb5 (amended):

- `test_d_role_state_reserve.py` docstring: dropped the incident reference and the path to a non-public report; the root-cause description is now self-contained.
- `_d_role_config(decode_input_tokens)`: removed the default; all three call sites pass `decode_input_tokens=1` explicitly.

`pre-commit run --all-files` is clean and the three regression tests pass.

The `perf-qwen3.5-397b-a17b-nvfp4-agentic-b200-8gpu` gate on the previous push failed on the conc=1 per-user decode speed only (520.83 tok/s vs a 522 floor, 89.8 % of reference; throughput 134.8 % of reference). That job runs the aggregate role (no disaggregation, `kvstore` off), which this change does not touch: the local-prefill branch keeps the same `setSnapshotStatePrefillReserve` call, only the D-role remote branch gains the extra reserve. Same gate passed on `daedc96b`.

dongjiyingdjy:
(Edited.) The follow-up commit a727b469 mentioned here was withdrawn from this PR before merge; the role-neutral version of the reserve, with the reviewer feedback folded in, is now #1401.

PR #1421, opened 2026-09-07
perf(autotune): tune exact decode shapes and refresh GEMM routes

chenht2022:
## Summary

Tune the actual decode path at configured batch sizes before CUDA graph capture, and replace packaged MoE tactic tables with a persistent FlashInfer startup cache.

Prefill-only warmup does not exercise all decode variants. Under EP, enumerating candidates with the global token count and local expert count can also exclude useful smaller tiles.

- **Exact decode tuning:** run each configured decode batch through the existing forward path, including speculative and attention-DP token multiples. Prefill retains FlashInfer's native tuning buckets.
- **EP candidate expansion:** combine candidates enumerated at `M` and `ceil(M * local_experts / total_experts)`. Profiling still uses the original full tensors and FlashInfer's timing and selection logic.
- **Persistent tuning cache:** synchronize cache contents across ranks, tune missing entries, and save the resulting tactics. Remove the packaged MoE tables and their offline sweeper. `--disable-autotune` still loads a matching cache while leaving uncovered shapes to the library's heuristic.
- **Refresh K3 GEMM routes:** update target/DSpark TP8/TP16 shapes, remove obsolete shapes, and extend the GB300 sweep to exact `M=1..64`. Autotune TGV and FI LL-BF16 before comparing backends with cold-L2 graph timing. The table stores backend choices; each deployment selects its own tactics. TGV tuning is restricted to 1-CTA variants.

## Performance

Full-model Kimi-K3 against public main (`3b6e097f`), with main's MoE tables regenerated for the installed FlashInfer version. Both arms enable FlashInfer autotuning and prefill/decode CUDA graphs, and receive the same requests.

Each configuration uses eight GPUs with attention TP8 and either MoE EP8 or MoE TP8. GB300 results average matched latencies across three setups for EP and two for TP before calculating speedups.

The table shows geometric-mean speed changes across the tested BS/ISL values: `100 * (geomean(old / new) - 1)`. Positive means faster. No-spec decode measures server step latency; spec measures verify-cycle latency; prefill measures server prefill latency.

| GPU | MoE layout | No-spec decode | Spec cycle | Prefill |
|---|---|---:|---:|---:|
| GB300 | EP8 | +0.47% | +1.13% | +4.01% |
| GB300 | TP8 | +0.60% | +1.00% | -0.38% |
| B300 | EP8 | +0.26% | +1.55% | +8.15% |
| B300 | TP8 | +0.61% | +1.03% | +0.68% |

The separate GEMM-route A/B showed a 0.68% geometric-mean speedup across 28 comparisons.

Spec results report verify-cycle latency only and should not be interpreted as end-to-end generation speedups. Decode measurements use fixed OSL4096 with `ignore_eos`.

Cold startup is more expensive: GB300 EP/no-spec averages about 119 minutes versus 37 minutes for the baseline, excluding its offline sweep. The resulting tactics are cached for reuse on later starts.

<details open>
<summary>Detailed performance: all BS and ISL values</summary>

All latencies are in milliseconds. Each entry shows **old → new**; speed change is `100 * (old / new - 1)`, so positive means faster. GB300 EP values are arithmetic means over three matched setups, and TP values over two. Old and new are averaged separately before calculating speed changes. B300 values come from one eight-GPU setup.

No-spec decode reports server step latency. Spec reports **verify-cycle latency only**, not per-output-token latency. Prefill reports server prefill latency.

### GB300 no-spec decode

| BS | MoE EP8 old → new (ms) | Speed change | MoE TP8 old → new (ms) | Speed change |
|---:|---:|---:|---:|---:|
| 1 | 10.533 → 10.593 | -0.56% | 10.115 → 10.116 | -0.01% |
| 2 | 11.464 → 11.451 | +0.11% | 10.980 → 10.958 | +0.20% |
| 4 | 12.763 → 12.750 | +0.10% | 12.104 → 12.056 | +0.41% |
| 8 | 15.486 → 15.489 | -0.02% | 14.743 → 14.735 | +0.06% |
| 16 | 19.298 → 19.020 | +1.46% | 18.574 → 18.288 | +1.56% |
| 32 | 25.386 → 24.996 | +1.56% | 24.940 → 24.559 | +1.55% |
| 64 | 34.934 → 34.710 | +0.65% | 35.324 → 35.160 | +0.47% |

### GB300 spec decode — cycle latency

| BS | MoE EP8 old → new (ms) | Speed change | MoE TP8 old → new (ms) | Speed change |
|---:|---:|---:|---:|---:|
| 1 | 15.635 → 15.604 | +0.20% | 15.164 → 15.032 | +0.88% |
| 2 | 18.608 → 18.328 | +1.53% | 17.930 → 17.689 | +1.37% |
| 4 | 23.281 → 22.851 | +1.88% | 22.843 → 22.510 | +1.48% |
| 8 | 31.398 → 31.251 | +0.47% | 31.621 → 31.438 | +0.58% |
| 16 | 38.397 → 38.490 | -0.24% | 39.326 → 39.415 | -0.23% |
| 32 | 50.708 → 50.403 | +0.61% | 53.447 → 51.949 | +2.89% |
| 64 | 71.593 → 69.159 | +3.52% | 71.913 → 71.867 | +0.06% |

### B300 no-spec decode

| BS | MoE EP8 old → new (ms) | Speed change | MoE TP8 old → new (ms) | Speed change |
|---:|---:|---:|---:|---:|
| 1 | 11.045 → 11.156 | -0.99% | 10.638 → 10.654 | -0.15% |
| 2 | 11.990 → 12.006 | -0.13% | 11.533 → 11.550 | -0.15% |
| 4 | 13.392 → 13.313 | +0.60% | 12.825 → 12.708 | +0.92% |
| 8 | 16.063 → 16.145 | -0.51% | 15.505 → 15.479 | +0.17% |
| 16 | 19.966 → 19.752 | +1.09% | 19.469 → 19.294 | +0.91% |
| 32 | 26.026 → 25.667 | +1.40% | 25.911 → 25.545 | +1.43% |
| 64 | 35.556 → 35.409 | +0.42% | 36.667 → 36.235 | +1.19% |

### B300 spec decode — cycle latency

| BS | MoE EP8 old → new (ms) | Speed change | MoE TP8 old → new (ms) | Speed change |
|---:|---:|---:|---:|---:|
| 1 | 16.426 → 16.459 | -0.20% | 15.905 → 15.984 | -0.49% |
| 2 | 19.528 → 19.148 | +1.99% | 19.017 → 18.578 | +2.37% |
| 4 | 24.179 → 23.783 | +1.67% | 23.926 → 23.633 | +1.24% |
| 8 | 32.350 → 32.517 | -0.52% | 33.175 → 32.892 | +0.86% |
| 16 | 39.484 → 39.615 | -0.33% | 41.071 → 40.914 | +0.38% |
| 32 | 52.325 → 51.915 | +0.79% | 55.923 → 54.245 | +3.09% |
| 64 | 76.763 → 71.266 | +7.71% | 74.627 → 74.766 | -0.19% |

### GB300 prefill

| ISL | MoE EP8 old → new (ms) | Speed change | MoE TP8 old → new (ms) | Speed change |
|---:|---:|---:|---:|---:|
| 1024 | 109.395 → 109.375 | +0.02% | 107.878 → 109.460 | -1.45% |
| 2048 | 142.160 → 128.881 | +10.30% | 133.476 → 133.570 | -0.07% |
| 4096 | 223.508 → 210.722 | +6.07% | 218.390 → 218.500 | -0.05% |
| 8192 | 371.279 → 371.271 | +0.00% | 383.479 → 383.296 | +0.05% |

### B300 prefill

| ISL | MoE EP8 old → new (ms) | Speed change | MoE TP8 old → new (ms) | Speed change |
|---:|---:|---:|---:|---:|
| 1024 | 109.284 → 95.644 | +14.26% | 97.273 → 96.591 | +0.71% |
| 2048 | 147.752 → 132.296 | +11.68% | 137.257 → 136.328 | +0.68% |
| 4096 | 234.492 → 219.292 | +6.93% | 228.618 → 226.521 | +0.93% |
| 8192 | 389.397 → 388.434 | +0.25% | 401.250 → 399.695 | +0.39% |

</details>

## Test Plan

- Decode: BS `1/2/4/8/16/32/64`, ISL512/OSL4096, no-spec and DSpark8.
- Prefill: single-request ISL `1024/2048/4096/8192`, one output token, prefix cache flushed per request.
- Mixed coding, AIME, MMLU, ShareGPT, and wiki inputs. Decode retains five rounds at BS1/2/4 and three at larger batches; prefill retains three rounds. Warmup rounds are discarded.
- Verified identical request sequences across old/new, unchanged tactic-cache files during measurement, and no new GPU Xid errors.
- Performance measurements preceded the final cleanup and rebase onto main.
- After cleanup and rebase: 14 focused tests passed; `pre-commit run --all-files` and `git diff --check` passed.


python/tokenspeed/runtime/execution/model_executor.py
```diff
@@ -140,6 +142,30 @@ def _cache_arena_attr(pool, name: str, default):
...
+        "moe_backend": str(server_args.moe_backend),
+        "attention_backend": str(server_args.attention_backend),
+        "attention": (mapping.attn.tp_size, mapping.attn.cp_size, mapping.attn.dp_size),
+        "dense": (mapping.dense.tp_size, mapping.dense.dp_size),
+        "moe": (mapping.moe.tp_size, mapping.moe.ep_size, mapping.moe.dp_size),
+        "linear_attention_tp": mapping.linear_attn.tp_size,
+        "pipeline_parallel": mapping.pp_size,
+        "speculative_algorithm": server_args.speculative_algorithm,
+        "speculative_num_draft_tokens": server_args.speculative_num_draft_tokens,
+    }
```
syuoni, line 166: Should we add flashinfer version and SM version to the cache key?

python/tokenspeed/runtime/execution/model_executor.py
```diff
@@ -484,7 +512,15 @@ def __init__(
...
-            self.forward_step.capture()
+            # Capture must use the same bucket mapper as decode tuning.
+            with autotune(
+                tune_mode=False,
+                tuning_buckets=self._decode_autotune_buckets(
+                    tuple(self.forward_step.capture_bs)
+                ),
+                round_up=False,
+            ):
+                self.forward_step.capture()
```
syuoni, line 523: This is more like a WAR. For token numbers outside the CUDA graph range, but still has a specialized tunable dispatch, it's still left untuned.

I think a better approach is let the dispatch place aware of the autotuner. Namely, if it's a autotuner pass, then run all dispatch branches, so all branches are tuned.

python/tokenspeed/runtime/execution/model_executor.py
```diff
@@ -545,55 +591,79 @@ def _autotune(self) -> None:
...
+        if self.config.disable_autotune:
+            logger.info("Kernel tuning disabled (--disable-autotune)")
+            return
+
+        decode_cases = tuple(sorted(self.forward_step.capture_bs))
+        logger.info(
+            "FlashInfer startup tuning: prefill=%d tokens, decode cases=%s",
+            num_tokens,
+            decode_cases if not self.forward_step.disable else (),
+        )
```
syuoni, line 626: Please use f-string.

python/tokenspeed/runtime/execution/model_executor.py
```diff
@@ -545,55 +591,79 @@ def _autotune(self) -> None:
...
-                else ib.positions_buf[:num_tokens]
-            )
-            with active_forward(ctx):
-                self.model_runner.forward(
-                    ctx=ctx,
-                    input_ids=ib.input_ids_buf[:num_tokens],
-                    positions=positions,
-                )
-        set_autotune_process_group(None)
+        try:
```
syuoni, line 631: If autotuner fails, I think we should let it fail.

python/tokenspeed/runtime/execution/model_executor.py
chenht2022: Already included via _collect_metadata(): FI version and GPU model.

python/tokenspeed/runtime/execution/model_executor.py
chenht2022: Updated.

python/tokenspeed/runtime/execution/model_executor.py
chenht2022: The finally only clears the process group; errors still propagate.

python/tokenspeed/runtime/execution/model_executor.py
syuoni: I see.

python/tokenspeed/runtime/execution/model_executor.py
syuoni: If the engine crashes, why we need to restore the process group for autotuner?

syuoni (commented):
Thanks @chenht2022 . I agree we should remove the offline tactic sweep. cc @dongjiyingdjy 

One major concern is the decode autotuner warmup, which looks more like a WAR to me.

chenht2022:
Hi, @syuoni 

Thanks for the suggestion. I replaced the per-decode-batch autotune warmup and graph-derived bucket overrides with **dispatch-owned tuning**: the startup traversal discovers operators, their dispatch sites expose the tunable branches, and FlashInfer enumerates its native buckets. Tuning coverage no longer depends on the CUDA graph capture ladder.

I also removed the ordinary GEMV backend table and its offline sweep. With tactics properly tuned, **TGV and FI cute-dsl—the previous `ll_bf16` path—already account for almost all small-M winners**. In our representative comparison around `flashinfer-ai/flashinfer#4908`, they covered **215/220 eligible points before the change and 220/220 afterward**.

Eligible **BF16-output GEMMs with M <= 32** now use one joint FI runner/tactic search, with selection and persistence handled by FI. **This also extends optimization to shapes missing from the old table**, including eligible projections in other models such as DeepSeek-V4.1 and shapes introduced by different tensor-parallel layouts. No model-specific table entries or additional offline sweeps are needed.

The comparison below uses **FI 0.6.18, without #4908**, against main `1e0be659b`. A is the previous PR implementation with the route table and exact-M tuning; B is the refactor with joint selection and native buckets.

| M <= 32 workload | A: latency reduction vs. main | B: latency reduction vs. main |
| --- | ---: | ---: |
| Original 36 N/K shapes, 1,152 points | 9.22% | **10.47%** |
| K3 target/draft subset: 22 shapes, 704 points | 13.25% | **13.36%** |

These are geometric-mean latency reductions from **B300 cold-weight CUDA-graph operator benchmarks**, not E2E speedups. The refactor preserves the aggregate small-M benefit while removing the offline table and extending coverage beyond its listed shapes; it does not claim that every individual point improves.

borontion (dismissed):

tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/flashinfer.py
```diff
@@ -701,29 +706,96 @@ def has_flashinfer_cute_dsl_bf16() -> bool:
...
+    if (
+        not flashinfer_joint_bf16_supported(x, weight, out)
+        or x.shape[0] > BF16_GEMM_MAX_M
+    ):
+        raise ValueError("Unsupported input to joint FlashInfer BF16 GEMM")
+    if out is None:
+        out = torch.empty((x.shape[0], weight.shape[0]), dtype=x.dtype, device=x.device)
+    workspace = _fi_gemm._get_cache_buf(
+        "mm_bf16_workspace", _fi_gemm.DEFAULT_WORKSPACE_SIZE, x.device
+    )
```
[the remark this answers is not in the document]
chenht2022: Checked against the pinned FlashInfer 0.6.18 implementation. This PR's `_bf16_gemm_runner_names()` selects only `tgv` and `cute-dsl`.

The selected runners unpack `a, b, bias, pdl, out, *_ = inputs`, discarding the sixth argument (`workspace_buffer`):

- [TGV runner](https://github.com/flashinfer-ai/flashinfer/blob/v0.6.18/flashinfer/gemm/gemm_base.py#L1480-L1501)
- [CuTeDSL split-K runner](https://github.com/flashinfer-ai/flashinfer/blob/v0.6.18/flashinfer/gemm/gemm_base.py#L1550-L1568)
- [CuTeDSL direct runner](https://github.com/flashinfer-ai/flashinfer/blob/v0.6.18/flashinfer/gemm/gemm_base.py#L1628-L1648)

Their kernel calls do not receive this buffer, so concurrent launches through this path do not read or overwrite the shared `mm_bf16_workspace`. The reported scratch-buffer race therefore does not apply to these selected runners. This conclusion is based on source inspection of the pinned version; it is not a general claim about other FlashInfer backends.

tokenspeed-kernel/python/tokenspeed_kernel/ops/tuning.py
```diff
@@ -126,185 +167,178 @@ def get_autotune_max_num_tokens() -> int:
...
-        # consequence at warning level so it is visible in serving logs.
-        logger.warning(
-            f"flashinfer tuning cache {path} rejected: environment metadata "
-            "(GPU model / flashinfer / CUDA versions) does not match this "
-            "host; the startup autotune window will tune instead. Re-run the "
-            "MoE tactic sweeper on this environment to regenerate it."
-        )
+        _install_autotune_cache_bytes(path, payload)
+    except OSError:
+        logger.warning("Could not mirror FlashInfer cache to %s", path, exc_info=True)
```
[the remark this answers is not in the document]
chenht2022: Fixed in 262eb4bce7e3e580f1ebcc60339be9d9fe7da7b1: converted all five parameterized cache log messages to f-strings. Validation: 10 cache tests passed, `pre-commit run --all-files` passed, and `git diff --check` passed.

PR #1423, opened 2026-09-07
fix: DFlash2 on Kimi-K3

Dogacel:
Fix DFlash2 failing to serve on Kimi-K3 in every configuration I tried: both MLA backends, both KV dtypes, several cudagraph capture sets, and `--enforce-eager`.

### What was wrong

- `DFlash2DecoderLayer.forward` took a required `out_cache_loc` that `DFlashDraftModel.forward` never passes, so the override could not be called at all. Both attention implementations already derive write locations from `ctx.attn_backend.write_locations()`, so the parameter is removed rather than defaulted.
- `DFlash2DraftModel` did not publish `attention_kind`, so DFLASH fell back to `"qwen_mha"` and set `num_extends = bs` instead of 0. Each MLA leaf then sliced away decode rows its own queries still needed: `tokenspeed_mla` went 32 to 28 rows and raised `Mismatched page_table.shape[0]`, `mla` went 32 to 0 rows and hit an illegal memory access.
- The absorbed MLA factors were computed at the end of `load_weights`, which `DummyModelLoader` never calls, so `--load-format dummy` left the decode kernel with null `w_kc`/`w_vc`. Moved to `post_load_weights()` and guarded for GQA checkpoints that have no `kv_b_proj`.

### Tested

- Kimi-K3 with `kimi-k3-dflash2` at 7 steps / 8 draft tokens on `--attention-backend mla`: reaches `engine ready` and serves, where it previously died in draft warmup.
- Same on `tokenspeed_mla`: also serves, and this patch touches no file that backend uses.
- Qwen3.8-27B with `Qwen3.8-27B-DFlash2` on `trtllm`: covers the GQA path through `post_load_weights`.
- Kimi-K3 with EAGLE-3.1 MLA at 3/4, real weights on TP16: accept length 3.032 / 2.490 against 2.972 / 2.490 before the patch, inside run to run noise.
- Reduced-layer Kimi-K3 under `--load-format dummy` on 4 GPUs: reproduces the original failure, serves after the fix.
- `test/runtime/test_dflash2.py` 17 passed, `pre-commit run --all-files` clean.

### Notes

`tokenspeed_mla` never reads `layer.sliding_window_size` while `mla`, `mha` and `trtllm` do, so `kimi-k3-dflash2` (4 sliding_attention layers at window 4096) will now run there with those layers unwindowed. This causes acceptance rate degredation, but will have a follow-up PR to fix it.

### DIFF
--- a/python/tokenspeed/runtime/models/dflash2.py
+++ b/python/tokenspeed/runtime/models/dflash2.py
@@ -286,6 +286,7 @@ class DFlash2DecoderLayer(DFlashDecoderLayer):
                 attention.group_id = FULL_ATTENTION
                 # Storage remains in Kimi-K3's full-attention group. This field
                 # is only the compute visibility contract for the MLA backend.
+                # TODO: mla honours it but tokenspeed_mla ignores it.
                 attention.sliding_window_size = sliding_window
             self.comm_manager = CommManager(
                 mapping=mapping,
@@ -309,13 +310,10 @@ class DFlash2DecoderLayer(DFlashDecoderLayer):
         positions: torch.Tensor,
         hidden_states: torch.Tensor,
         ctx: ForwardContext,
-        out_cache_loc: torch.Tensor,
         residual: torch.Tensor | None,
     ) -> tuple[torch.Tensor, torch.Tensor]:
         if ctx.forward_mode.is_idle():
-            return super().forward(
-                positions, hidden_states, ctx, out_cache_loc, residual
-            )
+            return super().forward(positions, hidden_states, ctx, residual)
 
         if residual is None:
             residual = hidden_states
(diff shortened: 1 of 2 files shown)


yubofredwang (commented):
Please add the follow up support of SWA for dflash2 in descriptions

yweng0828:
Hi @Dogacel , could you please take a look at this PR? I have fixed the functionality of K3 + DFlash2. https://github.com/lightseekorg/tokenspeed/pull/1399

Dogacel:
> Hi @Dogacel , could you please take a look at this PR? I have fixed the functionality of K3 + DFlash2. #1399

Sure, the PR looks pretty big, but I can close this PR in favor of yours.

yweng0828:
> > Hi @Dogacel , could you please take a look at this PR? I have fixed the functionality of K3 + DFlash2. #1399
> 
> Sure, the PR looks pretty big, but I can close this PR in favor of yours.

Thank you. I'd like to invite you to help me review #1399. That PR is indeed very big. I think half of the file changes are related to `tokenspeed_mla`. I'd like to invite the relevant maintainers to review it. : )

PR #1427, opened 2026-09-08
[DO NOT MERGE] fix(pd): stop failing prefill rooms that are only queued on the decode

slin1237:
## Summary

Under prefill/decode disaggregation, a burst of requests larger than the decode's running window makes the prefill fail requests that are merely queued, and every such failure then pins a decode slot for the full waiting timeout. Observed behind SMG with one prefill and one decode worker on Qwen3.5-9B at `--max-num-seqs 4` and 32 concurrent requests: 14 of 64 requests failed, the eval took 1392 s instead of ~35 s, and the failures arrived in waves of exactly four per 300 s.

Mechanism, on `main`:

- The prefill's bootstrap deadline starts when the request arrives (`MooncakeKVSender.__init__`, `init_time = time.time()`) and trips at `bootstrap_time_out` (120 s). The only thing that clears `Bootstrapping` is the decode's manifest, which the decode sends only once its scheduler has admitted the prompt (`buildDecodeWorkerPlan` admits at most one remote prompt per round, and a pre-allocated room holds one of the `max_num_seqs // dp_size` req-pool slots through the transfer and the whole generation). With a window of 4, only four of the 32 rooms are ever visible to the prefill while the other 28 clocks run. The deadline therefore measures the decode's queue, which the prefill cannot see, instead of the decode's liveness.
- On timeout, `poll()` called only `record_failure`, which writes a dict entry and never marks the room `Failed`. `_drop_request_state` then pops `request_status`, so both sticky-Failed guards in the bootstrap handler read `None` and accept the late pre-allocation: the room is stored and logged `-> Bootstrapped` with no sender behind it. Nothing transfers, and the decode holds its slot until the 300 s waiting timeout. `abort_room`'s docstring already names the gap: a room whose decode has not pre-allocated is only marked Failed locally.

The guards and `abort_room` already exist here; they were never wired to the timeout path.

The change, no wire-format change:

- `sender.py`: the bootstrap deadline is measured from the later of the request's arrival and the decode's last observed progress, so it detects a dead peer rather than a busy queue.
- `prefill.py`: `note_decode_progress()` is stamped on session registration and on every committed pre-allocation; a timed-out room goes through `abort_room` instead of a bare `record_failure`, so a decode that did pre-allocate is released immediately; timed-out rooms get a bounded-TTL tombstone (`failed_rooms`, TTL = the waiting timeout) that both bootstrap guards consult through `_is_room_failed`, so a late manifest is rejected instead of resurrecting the room.

Reviewer notes: `last_decode_progress` is manager-wide, so with several decode workers a live one masks a dead one and an abandoned room can outlive its client while traffic flows; the complete cure is a decode-to-prefill room-abort message. `abort_room` now runs on the event-loop thread. Adding pre-allocation headroom in the decode req pool, beyond the running window, is a worthwhile follow-up but is a scheduler change and is not part of this fix.

## Test Plan

- Five regression tests added to `test/runtime/distributed/test_cache_pd_executors.py`: a room stays `Bootstrapping` past the deadline while a sibling room's pre-allocation shows the decode is alive; a dead decode still trips the deadline; a timed-out room is aborted rather than only recorded; a late pre-allocation for a timed-out room does not enter `transfer_infos`, does not reach `Bootstrapped`, and pushes `Failed`; the tombstone expires after its TTL. All five fail on `main` and pass with this change (verified locally through a scratch import harness because the module needs Triton at import; they need the GPU CI to run them for real).
- `pre-commit run --all-files` clean.
- End-to-end: the SMG PD lane that produced the failure (run 34173426995) versus the same lane with a 32-sequence window (run 34180087896: zero bootstrap or transfer failures, ~35 s per eval). A rerun of that lane at `--max-num-seqs 4` with this fix is the confirmation that the cascade is gone.

### DIFF
--- a/python/tokenspeed/runtime/pd/mooncake/prefill.py
+++ b/python/tokenspeed/runtime/pd/mooncake/prefill.py
@@ -23,7 +23,7 @@ import os
 import socket
 import threading
 import time
-from collections import defaultdict
+from collections import OrderedDict, defaultdict
 from collections.abc import Iterator
 from dataclasses import replace
 from itertools import chain, islice
@@ -120,6 +120,16 @@ class MooncakeKVManagerPrefill(MooncakeKVManagerBase):
             )
         self.start_transfer_thread(transfer_thread_pool_size, transfer_queue_size)
         self.bootstrap_time_out = envs.TOKENSPEED_DISAGGREGATION_BOOTSTRAP_TIMEOUT.get()
+        # Last time any decode made bootstrap progress with this prefill; the
+        # bootstrap deadline is measured from this as well as from arrival.
+        self.last_decode_progress = time.time()
+        # Rooms this prefill gave up on (room -> monotonic stamp), kept so a
+        # late pre-allocation for one of them is rejected instead of accepted.
+        self.failed_rooms: OrderedDict[int, float] = OrderedDict()
+        self.failed_room_ttl = max(
+            envs.TOKENSPEED_DISAGGREGATION_WAITING_TIMEOUT.get(), 0
+        )
+        self.failed_room_lock = threading.Lock()
         # Publish this manager only after every field used by its bootstrap and
         # transfer threads has been initialized.
         self.start_prefill_thread()
@@ -178,6 +188,42 @@ class MooncakeKVManagerPrefill(MooncakeKVManagerBase):
         with self.bootstrap_token_cond:
             self.prefill_metadata.pop(room, None)
             self.bootstrap_token_cond.notify_all()
+
+    def note_decode_progress(self) -> None:
+        """Record that a decode just made bootstrap progress with this prefill."""
+        self.last_decode_progress = time.time()
+
+    def _reap_failed_rooms(self) -> None:
+        """Drop expired tombstones; callers hold ``failed_room_lock``."""
+        if self.failed_room_ttl <= 0:
+            self.failed_rooms.clear()
+            return
+        now = time.monotonic()
+        while self.failed_rooms:
+            room, failed_at = next(iter(self.failed_rooms.items()))
+            if now - failed_at < self.failed_room_ttl:
+                return
+            del self.failed_rooms[room]
+
+    def _mark_room_failed(self, room: int) -> None:
+        """Tombstone one aborted room so a late pre-allocation is rejected."""
+        if self.failed_room_ttl <= 0:
+            return
+        with self.failed_room_lock:
+            # Re-insert at the tail so insertion order stays expiry order.
+            self.failed_rooms.pop(room, None)
+            self.failed_rooms[room] = time.monotonic()
+            self._reap_failed_rooms()
+
+    def _is_room_failed(self, room: int) -> bool:
+        """True when the room was failed, whether still tracked or tombstoned."""
+        if self.request_status.get(room) == TransferPoll.Failed:
+            return True
+        if self.failed_room_ttl <= 0:
+            return False
+        with self.failed_room_lock:
+            self._reap_failed_rooms()
+            return room in self.failed_rooms
 
     def _wait_prefill_metadata(
         self,
@@ -742,10 +788,14 @@ class MooncakeKVManagerPrefill(MooncakeKVManagerBase):
         status to every decode endpoint that already pre-allocated for this room
         (mirrors the in-transfer failure path), so the decode raises a FailedEvent and
         the client gets an error instead of hanging. A room whose decode has not
-        pre-allocated yet is only marked Failed locally (no endpoint to notify).
+        pre-allocated yet is only marked Failed locally (no endpoint to notify)
+        -- but the verdict is tombstoned, so if that decode does pre-allocate
+        later, ``_handle_bootstrap_message`` rejects the manifest and pushes
+        Failed to it then, rather than accepting a room that has no sender.
         """
         self.record_failure(room, reason)
         self.update_status(room, TransferPoll.Failed)
+        self._mark_room_failed(room)
         for req in list(self.transfer_infos.get(room, {}).values()):
             if not req.is_dummy:
                 try:
@@ -934,6 +984,7 @@ class MooncakeKVManagerPrefill(MooncakeKVManagerBase):
             self.rejected_decode_sessions.pop(session_id, None)
             with self.session_lock:
                 self._clear_failed_session(session_id)
+            self.note_decode_progress()
             logger.info(
                 "[Prefill bootstrap_thread] registered kv_args from decode session=%s",
                 session_id,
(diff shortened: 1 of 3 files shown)


stmatengss:
**[P1] Keep bootstrap-timeout notifications off the synchronous polling path**

At [sender.py:143–147](https://github.com/lightseekorg/tokenspeed/blob/495dcd04d6d9285830c74091b93b982bc8dd6d24/python/tokenspeed/runtime/pd/mooncake/sender.py#L143-L147), the new `abort_room()` call makes `poll()` acquire the cached endpoint lock and call `send_multipart()` synchronously. A room can still be `Bootstrapping` after only part of its decode fanout has supplied manifests, so there can already be endpoints to notify when this timeout fires.

If a decode stops receiving and its PUSH queue reaches the high-water mark, a send can block indefinitely; the socket has no `SNDTIMEO` or nonblocking flag. Another sender can also hold the same endpoint lock while blocked in a send. This propagates into `generate_events() → poll_and_all_reduce() → sender.poll()`, stopping the prefill event loop and preventing it from reaching the following all-reduce. A failed peer can therefore stall unrelated requests as well. See the [ZMQ PUSH blocking semantics](https://libzmq.readthedocs.io/en/latest/zmq_socket.html#zmq-push).

**Reproduction**

1. Fetch the reviewed PR head: `git fetch origin refs/pull/1427/head`.
2. Run the script below from the repository root with Python 3.10+. It compares merge base `4a60ccf9` with reviewed head `495dcd04`, without changing the checkout.
3. Seed an expired bootstrapping room with one registered, non-dummy destination. Hold the cached endpoint lock, invoke the original `poll()` method in a thread, and check whether it returns before releasing the lock.

Observed output:

```text
4a60ccf9 blocked_on_notification_lock= False returned= [0]
495dcd04 blocked_on_notification_lock= True returned= []
```

Both return `TransferPoll.Failed` after the lock is released. This is a bounded, isolated reproduction of lock contention, **not a real network/HWM or GPU E2E test**: the script extracts the unchanged method bodies from Git and substitutes only the connection boundary with a fake socket and a real held lock. The new notification tests use immediately returning lambdas and do not exercise this case.

<details>
<summary>Runnable reproduction (standard library only)</summary>

```python
import ast
import subprocess
import threading
import time
from collections import OrderedDict
from types import SimpleNamespace

ROOT = "python/tokenspeed/runtime/pd/"

def source(ref, path):
    return subprocess.check_output(["git", "show", ref + ":" + ROOT + path], text=True)

def methods(ref, path, class_name, names, env):
    tree = ast.parse(source(ref, path))
    cls = next(n for n in tree.body if isinstance(n, ast.ClassDef) and n.name == class_name)
    selected = [n for n in cls.body if isinstance(n, ast.FunctionDef) and n.name in names]
    module = ast.Module(body=[
        ast.ImportFrom(module="__future__", names=[ast.alias(name="annotations")], level=0),
        *selected,
    ], type_ignores=[])
    exec(compile(ast.fix_missing_locations(module), path, "exec"), env)
    return {n.name: env[n.name] for n in selected}

for ref in ("4a60ccf9", "495dcd04"):
    env = dict(time=time, threading=threading, OrderedDict=OrderedDict,
               logger=SimpleNamespace(warning_once=lambda *a: None,
                                      exception=lambda *a: None))
    exec(source(ref, "base/status.py"), env)
    state = env["TransferPoll"]
    base = methods(ref, "base/manager.py", "DisaggManagerBase",
                   {"check_status", "update_status", "record_failure"}, env)
    prefill = methods(ref, "mooncake/prefill.py", "MooncakeKVManagerPrefill",
                      {"abort_room", "_mark_room_failed", "_reap_failed_rooms",
                       "sync_status_to_decode_endpoint"}, env)
    poll = methods(ref, "mooncake/sender.py", "MooncakeKVSender", {"poll"}, env)["poll"]
    manager = type("Manager", (), dict(base, **prefill))()
    manager.request_status = {9: state.Bootstrapping}
    manager.failure_records = {}
    manager.failure_lock = threading.Lock()
    manager.failed_rooms = OrderedDict()
    manager.failed_room_ttl = 300
    manager.failed_room_lock = threading.Lock()
    manager.bootstrap_time_out = 120
    manager.last_decode_progress = time.time() - 600
    manager.transfer_infos = {
        9: {"session": SimpleNamespace(is_dummy=False, room=9, mooncake_session_id="session")}
    }
    manager.decode_kv_args_table = {
        "session": SimpleNamespace(endpoint="127.0.0.1", dst_port=9000)
    }
    manager.rejected_decode_sessions = {}
    manager._status_prefill_rank = 0

    # Model another sender holding the cached endpoint lock.
    lock = threading.Lock()
    lock.acquire()
    manager._connect = lambda endpoint: (
        SimpleNamespace(send_multipart=lambda frames: None), lock
    )
    sender = SimpleNamespace(kv_mgr=manager, bootstrap_room=9,
                             conclude_state=None, init_time=time.time() - 600)
    results = []
    worker = threading.Thread(target=lambda: results.append(poll(sender)))
    worker.start()
    try:
        worker.join(0.2)
        blocked = worker.is_alive()
        print(ref, "blocked_on_notification_lock=", blocked, "returned=", results)
        assert blocked == (ref == "495dcd04")
    finally:
        lock.release()
        worker.join(1)
    assert not worker.is_alive() and results == [state.Failed]
```

</details>

Please keep the local failure/tombstone update synchronous, but enqueue remote notifications for background delivery with bounded send waits, so timeout polling never waits on an endpoint lock or network send.

FC-Li:
We need more discussions on this PR. Please don't merge before we figure out the trade off between different solutions.

FC-Li:
@kangguangli could you take a look at this PR?

kangguangli:
Thanks for investigating this issue and putting together this PR. I have a couple of concerns about the current approach.

First, I think it would be better not to extend the prefill-side timeout this way. The current behavior at least allows requests to terminate, whereas the behavior introduced by this PR could leave requests hanging indefinitely.

Second, abort_room() is not a perfect solution either. We recently encountered a similar issue, and our initial proposal was to have prefill proactively notify decode when a request fails. The problem is that if prefill times out before receiving decode’s bootstrap request, it does not know where to send the abort message.

We are therefore also considering whether SMG would be a better place to initiate cancellation: regardless of whether P or D times out, let SMG abort the request, while SMG is a necessary component of the PD architecture.

FC-Li:
> Thanks for investigating this issue and putting together this PR. I have a couple of concerns about the current approach.
> 
> First, I think it would be better not to extend the prefill-side timeout this way. The current behavior at least allows requests to terminate, whereas the behavior introduced by this PR could leave requests hanging indefinitely.
> 
> Second, abort_room() is not a perfect solution either. We recently encountered a similar issue, and our initial proposal was to have prefill proactively notify decode when a request fails. The problem is that if prefill times out before receiving decode’s bootstrap request, it does not know where to send the abort message.
> 
> We are therefore also considering whether SMG would be a better place to initiate cancellation: regardless of whether P or D times out, let SMG abort the request, while SMG is a necessary component of the PD architecture.

@zhyncs Let’s organize a meeting for deeper discussion?

PR #1428, opened 2026-09-08
test: add DeepSeek V4 Flash agentic benchmark

Xiangyi1996:
## Summary

- add a SWE-Smith multi-turn agentic benchmark for DeepSeek V4 Flash
- add one manually checked four-GPU deployment config using Attention TP4, Dense TP4, and routed MoE TP1 x EP4
- collect per-user decode rate, per-GPU throughput, cache hit rate, and decoded tokens per iteration using the existing GLM5.2 benchmark format
- adapt dataset generation to the checkpoint-provided DeepSeek V4 chat encoding

## Validation

- shell syntax checks for the benchmark runner and deployment config
- Python compile checks for the tokenizer adapter and output collector
- dataset-builder patch dry run against the pinned EvalScope revision
- manual p1, p2, p4, and p8 end-to-end sweep with all requests completing successfully

### DIFF
--- a/test/agentic_benchmark/deepseek_v4_flash/tokenspeed/agentic_bench.sh
+++ b/test/agentic_benchmark/deepseek_v4_flash/tokenspeed/agentic_bench.sh
@@ -116,7 +116,7 @@ for CONFIG in "${CONFIGS[@]}"; do
     fi
     echo "Warmup..."
     evalscope perf \
-        --model deepseek-v4-flash \
+        --model deepseek-ai/DeepSeek-V4-Flash-0731 \
         --url http://127.0.0.1:8000/v1/chat/completions \
         --api openai \
         --dataset swe_smith \
@@ -131,15 +131,15 @@ for CONFIG in "${CONFIGS[@]}"; do
 
     echo "Benchmark..."
     evalscope perf \
-        --model deepseek-v4-flash \
+        --model deepseek-ai/DeepSeek-V4-Flash-0731 \
         --url http://127.0.0.1:8000/v1/chat/completions \
         --api openai \
         --dataset swe_smith \
         --dataset-path agentic_dataset.json \
         --max-tokens 500 \
         --multi-turn \
-        --number 4 8 8 16 \
-        --parallel 1 2 4 8 \
+        --number 4 8 8 16 32 \
+        --parallel 1 2 4 8 16 \
         --extra-args '{"ignore_eos": true, "temperature": 0}' \
         --name $CONFIG \
         --outputs-dir $SWEEP_DIR \
--- a/test/agentic_benchmark/deepseek_v4_flash/tokenspeed/agentic_bench.sh
+++ b/test/agentic_benchmark/deepseek_v4_flash/tokenspeed/agentic_bench.sh
@@ -1,0 +1,148 @@
+#!/usr/bin/bash
+
+set -euo pipefail
+
+# Prepare dataset
+EVALSCOPE_COMMIT=acd09b44384d53174768bb1063f675420f76fae9
+pip install "evalscope[perf] @ git+https://github.com/modelscope/evalscope.git@${EVALSCOPE_COMMIT}"
+
+if [ ! -f agentic_dataset.json ]; then
+    (
+        BUILD_SCRIPT=$(mktemp)
+        trap 'rm -f "$BUILD_SCRIPT"' EXIT
+        wget \
+            https://raw.githubusercontent.com/modelscope/evalscope/${EVALSCOPE_COMMIT}/examples/perf/build_swe_smith_dataset.py \
+            -O "$BUILD_SCRIPT"
+        patch -s "$BUILD_SCRIPT" build_swe_smith_dataset.patch
+        PYTHONPATH=. python3 "$BUILD_SCRIPT" \
+            --model-path deepseek-ai/DeepSeek-V4-Flash-0731 \
+            --dataset-name SWE-bench/SWE-smith-trajectories \
+            --split tool \
+            --first-turn-length 50000 \
+            --subsequent-turn-length 800 \
+            --min-turns 10 \
+            --max-turns 15 \
+            --number 128 \
+            --output-path agentic_dataset.json \
+            --chars-per-token 3.0 \
+            --seed 42 \
+            --num-workers 1
+    )
+fi
+
+# Sweep configs
+CONFIGS=(
+    attn_tp4_moe_ep4
+)
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SERVER_PID=
+SERVER_LOG=
+
+launch_server() {
+    local config=$1
+    SERVER_LOG=/tmp/tokenspeed_server_${config}.log
+    setsid ${SCRIPT_DIR}/configs/${config}.sh > "$SERVER_LOG" 2>&1 &
+    SERVER_PID=$!
+}
+
+wait_for_ready() {
+    local TIMEOUT=3600
+    local START=$SECONDS
+    until curl -sf -o /dev/null http://127.0.0.1:8000/readiness; do
+        if ! kill -0 "$SERVER_PID" 2>/dev/null; then
+            echo "Server died early. Last log lines:" >&2
+            tail -100 "$SERVER_LOG" >&2
+            return 1
+        fi
+        if grep -qE "CUDA out of memory|OutOfMemory|RuntimeError|Killed" "$SERVER_LOG"; then
+            echo "Server hit a fatal error:" >&2
+            tail -100 "$SERVER_LOG" >&2
+            return 1
+        fi
+        if (( SECONDS - START > TIMEOUT )); then
+            echo "Timeout after ${TIMEOUT}s waiting for server" >&2
+            return 1
+        fi
+        sleep 5
+    done
+    echo "Server ready after $((SECONDS - START))s"
+}
+
+stop_server() {
+    if [[ -n "$SERVER_PID" ]] && kill -0 "$SERVER_PID" 2>/dev/null; then
+        echo "Stopping ts serve (pgid $SERVER_PID)..."
+        kill -TERM -"$SERVER_PID" 2>/dev/null || true
+        for _ in {1..20}; do
+            kill -0 "$SERVER_PID" 2>/dev/null || break
+            sleep 1
+        done
+        kill -KILL -"$SERVER_PID" 2>/dev/null || true
+    fi
+    SERVER_PID=
+}
+
+wait_for_port_free() {
+    local port=${1:-8000}
+    local timeout=${2:-90}
+    local start=$SECONDS
+    while ! python3 -c "import socket; s=socket.socket(); s.bind(('127.0.0.1', $port)); s.close()" 2>/dev/null; do
+        if (( SECONDS - start > timeout )); then
+            echo "Port ${port} still in use after ${timeout}s" >&2
+            return 1
+        fi
+        sleep 1
+    done
+}
+
+trap stop_server EXIT  # safety net for Ctrl-C / errors
+
+# Preflight: bail out if port 8000 is already in use
+wait_for_port_free 8000
+
+SWEEP_TS=$(date +%Y%m%d_%H%M%S)
+SWEEP_DIR="${SCRIPT_DIR}/outputs/${SWEEP_TS}"
+echo "Sweep outputs: ${SWEEP_DIR}"
+
+for CONFIG in "${CONFIGS[@]}"; do
+    echo "=== Running $CONFIG ==="
+    launch_server "$CONFIG"
+
+    if ! wait_for_ready; then
+        stop_server
+        exit 1
+    fi
+    echo "Warmup..."
+    evalscope perf \
+        --model deepseek-v4-flash \
+        --url http://127.0.0.1:8000/v1/chat/completions \
+        --api openai \
+        --dataset swe_smith \
+        --dataset-path agentic_dataset.json \
+        --max-tokens 500 \
+        --multi-turn \
+        --number 2 \
+        --parallel 2 \
+        --extra-args '{"ignore_eos": true, "temperature": 0}' \
+        --dataset-offset 68 \
+        --outputs-dir /tmp/outputs
+
+    echo "Benchmark..."
+    evalscope perf \
+        --model deepseek-v4-flash \
+        --url http://127.0.0.1:8000/v1/chat/completions \
+        --api openai \
+        --dataset swe_smith \
+        --dataset-path agentic_dataset.json \
+        --max-tokens 500 \
+        --multi-turn \
+        --number 4 8 8 16 \
+        --parallel 1 2 4 8 \
+        --extra-args '{"ignore_eos": true, "temperature": 0}' \
+        --name $CONFIG \
+        --outputs-dir $SWEEP_DIR \
+        --no-timestamp
+
+    stop_server
+    wait_for_port_free 8000
+done
--- a/test/agentic_benchmark/deepseek_v4_flash/tokenspeed/agentic_bench.sh
+++ b/test/agentic_benchmark/deepseek_v4_flash/tokenspeed/agentic_bench.sh
@@ -2,6 +2,9 @@
 
 set -euo pipefail
 
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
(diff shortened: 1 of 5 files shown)


test/agentic_benchmark/deepseek_v4_flash/tokenspeed/agentic_bench.sh
```diff
@@ -0,0 +1,148 @@
...
+
+if [ ! -f agentic_dataset.json ]; then
+    (
+        BUILD_SCRIPT=$(mktemp)
+        trap 'rm -f "$BUILD_SCRIPT"' EXIT
+        wget \
+            https://raw.githubusercontent.com/modelscope/evalscope/${EVALSCOPE_COMMIT}/examples/perf/build_swe_smith_dataset.py \
+            -O "$BUILD_SCRIPT"
+        patch -s "$BUILD_SCRIPT" build_swe_smith_dataset.patch
+        PYTHONPATH=. python3 "$BUILD_SCRIPT" \
```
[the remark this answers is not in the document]
Xiangyi1996: Fixed in 902b437f. The script now resolves SCRIPT_DIR and changes to it before dataset preparation, so the patch, tokenizer module, dataset, config, and output paths are independent of the caller working directory. I also verified the patch against the pinned EvalScope builder.

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/agentic_bench.sh
```diff
@@ -0,0 +1,148 @@
...
+            sleep 1
+        done
+        kill -KILL -"$SERVER_PID" 2>/dev/null || true
+    fi
+    SERVER_PID=
+}
+
+wait_for_port_free() {
+    local port=${1:-8000}
+    local timeout=${2:-90}
```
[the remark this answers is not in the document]
Xiangyi1996: Fixed in 902b437f. wait_for_port_free now requires both port and timeout, and both call sites pass 8000 and 90 explicitly. Shell syntax checks and pre-commit run --all-files pass.

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/bash
+
+set -euo pipefail
+
+exec ts serve \
+    --model deepseek-ai/DeepSeek-V4-Flash-0731 \
+    --revision 7872f01b1d1fe23eabc4c98b48bffcef5a386062 \
```
syuoni: why do we pin the revision?

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
```diff
@@ -0,0 +1,39 @@
...
+    --served-model-name deepseek-v4-flash \
```
syuoni: Any reasons not to use `deepseek-ai/DeepSeek-V4-Flash-0731`?

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
```diff
@@ -0,0 +1,39 @@
...
+    --trust-remote-code \
+    --world-size 4 \
+    --nprocs-per-node 4 \
+    --nnodes 1 \
+    --pipeline-parallel-size 1 \
```
syuoni:
```suggestion
```

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
```diff
@@ -0,0 +1,39 @@
...
+    --attn-tp-size 4 \
+    --data-parallel-size 1 \
```
syuoni:
```suggestion
```

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
```diff
@@ -0,0 +1,39 @@
...
+    --draft-moe-backend mega_moe \
+    --kv-cache-dtype fp8_e4m3 \
+    --attention-use-fp4-indexer-cache \
+    --max-model-len 96000 \
+    --max-num-seqs 8 \
+    --max-total-tokens 2560000 \
+    --max-prefill-tokens 8192 \
+    --chunked-prefill-size 8192 \
+    --gpu-memory-utilization 0.90 \
+    --disable-kvstore \
```
syuoni, line 24: why `--disable-kvstore`?

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
```diff
@@ -0,0 +1,39 @@
...
+    --nprocs-per-node 4 \
+    --nnodes 1 \
+    --pipeline-parallel-size 1 \
+    --attn-tp-size 4 \
+    --data-parallel-size 1 \
+    --dense-tp-size 4 \
+    --moe-tp-size 1 \
+    --ep-size 4 \
+    --moe-backend flashinfer_trtllm \
+    --draft-moe-backend mega_moe \
```
syuoni, line 15: why do we use different MoE backends for target and draft models?

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/agentic_bench.sh
```diff
@@ -0,0 +1,150 @@
...
+    evalscope perf \
+        --model deepseek-v4-flash \
+        --url http://127.0.0.1:8000/v1/chat/completions \
+        --api openai \
+        --dataset swe_smith \
+        --dataset-path agentic_dataset.json \
+        --max-tokens 500 \
+        --multi-turn \
+        --number 4 8 8 16 \
+        --parallel 1 2 4 8 \
```
syuoni: can we scale it to bs16, following other models' agentic bench?

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
Xiangyi1996: The revision is pinned so that the performance gate always uses the same checkpoint, model configuration, tokenizer, and V4 encoding files. Just for the reproduction, can delete now.

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
Xiangyi1996: Just keep the name short, can move to '-0731' now

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/agentic_bench.sh
Xiangyi1996: Yes. Previously validated V4 deployment range on v4 pro. We can extend the sweep to parallelism 16.

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
Xiangyi1996: Current dspark will report ValueError when using the kvstore. `DSPARK same-checkpoint decoding does not support KVStore`

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
Xiangyi1996: The two paths execute different graphs and tensor shapes: the target evaluates the full model and currently uses the validated flashinfer_trtllm MoE path, whereas the much smaller draft path uses the validated mega_moe integration for its speculative-decoding workload.
Using mega_moe for both is supported, but it changes the target execution path and has not yet been established as a better drop-in replacement under the same end-to-end configuration. We should evaluate that as a separate matched benchmark rather than change this baseline implicitly. I’ll add a comment to the config to make the distinction explicit.

syuoni: This is a very basic perf knob. We should make sure it's optimal.

Xiangyi1996: I ran a matched p1/p2/p4/p8/p16 sweep on the same 4x GB200 node, changing only the target MoE backend. Compared with using `mega_moe` for both models, the current `flashinfer_trtllm` target + `mega_moe` draft configuration improved total throughput by 6.18%, 3.79%, 6.08%, -2.96%, and 4.92%, respectively. Median TPOT was also 10.42%, 5.00%, and 8.04% lower at p1/p2/p4; p8 was effectively flat (-1.14%), while p16 was 2.20% higher. All requests completed successfully and DSpark acceptance remained about 80% in both runs.

The current mixed-backend configuration gives the better overall p1-p16 curve, so I retained it as the single perf-gate configuration rather than adding the generally slower dual-`mega_moe` variant.

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
syuoni: Any reasons to keep them?

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
syuoni: Any reasons to keep it?

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
syuoni: Not blocking this PR, but while-box analysis is needed here.

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/agentic_bench.sh
```diff
@@ -0,0 +1,150 @@
...
+pip install "evalscope[perf] @ git+https://github.com/modelscope/evalscope.git@${EVALSCOPE_COMMIT}"
+
+if [ ! -f agentic_dataset.json ]; then
+    (
+        BUILD_SCRIPT=$(mktemp)
+        trap 'rm -f "$BUILD_SCRIPT"' EXIT
+        wget \
+            https://raw.githubusercontent.com/modelscope/evalscope/${EVALSCOPE_COMMIT}/examples/perf/build_swe_smith_dataset.py \
+            -O "$BUILD_SCRIPT"
+        patch -s "$BUILD_SCRIPT" build_swe_smith_dataset.patch
```
syuoni, line 19: Would you please explain why the patch is needed?

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
syuoni: Also, another baseline would be `flashinfer_trtllm` for both target and drafter.

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
Xiangyi1996: has already been deleted

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/agentic_bench.sh
Xiangyi1996: The upstream EvalScope builder assumes a standard AutoTokenizer.apply_chat_template path. DeepSeek V4 Flash instead requires the checkpoint-provided encoding/encoding_dsv4.py; without it, the generated conversations and token-length filtering do not use the model’s actual chat encoding.
This patch makes the builder load that V4 encoding and use the same chat-template path for token counting. It also pins the SWE-Smith dataset revision for reproducibility and supports the single-worker invocation used here. The patch only affects dataset generation, not the serving or benchmark execution path.

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
Xiangyi1996: No reason to keep them. Removed `--nprocs-per-node 4` and `--nnodes 1`; for this single-node `--world-size 4` configuration, TokenSpeed derives them as 4 and 1.

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
Xiangyi1996: No reason to keep it. `--data-parallel-size 1` was removed; it is implied by this TP4 configuration.

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
Xiangyi1996: has been deleted.

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
Xiangyi1996: Thanks. I’ll add the missing flashinfer_trtllm target + drafter baseline and profile the target and draft MoE paths, and update the gate configuration if the additional baseline performs better.

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/agentic_bench.sh
syuoni: Thanks for explanation.

test/agentic_benchmark/deepseek_v4_flash/tokenspeed/configs/attn_tp4_moe_ep4.sh
Xiangyi1996: I completed the missing matched baseline on the same 4x GB200 node, using the same runtime, dataset, and TP4+EP4 settings, and changing only the draft MoE backend from `mega_moe` to `flashinfer_trtllm`.

| concurrency | total throughput vs mixed | median TPOT vs mixed | acceptance delta |
|---|---:|---:|---:|
| p1 | +1.86% | -3.66% | +0.37 pp |
| p2 | +1.47% | -1.29% | +0.24 pp |
| p4 | -4.74% | +5.00% | -1.27 pp |
| p8 | +4.74% | -7.04% | +0.82 pp |
| p16 | +2.58% | flat | +0.17 pp |

All requests completed successfully in both runs. The dual-`flashinfer_trtllm` baseline is not uniformly better: it improves four throughput points, but regresses at p4 together with lower acceptance and higher median TPOT. Given that mixed-backend performance also has some run-to-run movement, this single paired run is not strong enough to replace the current gate configuration. I will keep the current mixed configuration for this PR and follow up with the requested white-box profiling of the target and draft MoE paths separately.

Xiangyi1996: The requested white-box analysis has also been completed separately. It did not show a draft-MoE kernel regression at p4, so I’m keeping the mixed-backend configuration as the perf gate. Since this follow-up is non-blocking for the benchmark PR, I’ll track the profiling analysis separately.

syuoni (approved):

PR #1430, opened 2026-09-08
perf(amd): Optimize GLM-5.3-Flash recurrent verification

Max191:
## Summary

Optimize the recurrent-attention path used by GLM-5.3-Flash speculative-token
verification, building on the KPool branch below it.

- Use the measured two-warp, three-stage recurrent schedule for the exact TP4,
  batch-16, four-token verification shape while preserving the existing
  schedule for other shapes.
- Read initial recurrent state directly from committed cache pages and keep
  speculative checkpoint writes in scratch storage, removing the full state
  seed copy without changing accepted-state commits.
- Reuse graph-stable scratch base rows, removing 34 small integer launches from
  the 34-layer verification graph.
- Expose packed query, key, and value output as zero-copy views when safe;
  retain materialization when replay must preserve inputs for accepted-state
  reconstruction and on other model paths.
- Remove an unused padded-batch argument from prompt convolution so warmup and
  tail partitions reuse one compiled kernel.

The direct committed-state path applies to the portable recurrent verifier.
The related recurrent architecture retains full seeding, and NVIDIA's fused
verifier path is unchanged.

## ShareGPT performance

The five-commit recurrent-attention group was measured against PR 1 (https://github.com/lightseekorg/tokenspeed/pull/1429):

| Metric | PR 1 | This PR | Change |
|---|---:|---:|---:|
| Output throughput | 1,020.272 tok/s | 1,103.155 tok/s | +8.12% |
| Decode/user | 86.265 tok/s | 92.197 tok/s | +6.88% |
| Mean TPOT | 11.592 ms | 10.846 ms | 6.43% lower |
| Mean ITL | 36.622 ms | 35.500 ms | 3.06% lower |
| Mean TTFT | 501.905 ms | 501.476 ms | 0.09% lower |
| Mean end-to-end latency | 6,425.501 ms | 6,042.897 ms | 5.95% lower |

Supporting graph-replay measurements showed a 12.10% faster recurrent launch,
a 22.93% faster 34-layer verifier region from direct state reads, a 4.02%
improvement from cached scratch rows, and a 12.24% reduction for the verifier
producer-plus-recurrence graph from zero-copy packed views. These isolated
percentages overlap and should not be added together; the table is the
PR-level result.

The ShareGPT workload used four MI350X GPUs, attention TP4, MoE TP4/EP1, MTP3
with four-token verification, 16 concurrent requests, and 512 output tokens
per request. These measurements were collected on `upstream/main` `b174a318`.
The current branch carries the same executable feature changes on newer
`upstream/main` `5af23865`, but that resulting tree was not benchmarked.

### DIFF
--- a/python/tokenspeed/runtime/layers/attention/backends/state/kda.py
+++ b/python/tokenspeed/runtime/layers/attention/backends/state/kda.py
@@ -101,6 +101,7 @@ class KdaAttnBackend(MambaAttnBackend):
     """
 
     _verify_reads_committed_recurrent_state = True
+    _verify_packed_qkv_views = True
 
     def __init__(
         self,
--- a/python/tokenspeed/runtime/layers/attention/backends/state/kda.py
+++ b/python/tokenspeed/runtime/layers/attention/backends/state/kda.py
@@ -100,6 +100,8 @@ class KdaAttnBackend(MambaAttnBackend):
     the conv, the gate GEMV and the recurrence into a single launch.
     """
 
+    _verify_reads_committed_recurrent_state = True
+
     def __init__(
         self,
         config: AttnConfig,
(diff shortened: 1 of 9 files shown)


borontion (approved):

borontion:
@Max191 there are some conflicts after rebase

borontion (approved):

PR #1434, opened 2026-09-08
perf(amd): Optimize GLM-5.3-Flash MoE decode on MI350X

Max191:
## Summary

Optimize the GFX950 BF16 mixture-of-experts (MoE) path for GLM-5.3-Flash
decode and four-token verification shapes.

- Use a smaller final expert-reduction tile for graph shapes through 64 rows
  while retaining the existing schedule for larger workloads.
- Build expert block ranges directly for small routing workloads instead of
  scanning every expert for every output block.
- Keep the live row count dynamic so sigmoid-routing kernels are reused across
  request counts within the same padded graph shape.

The changes preserve route offsets, the existing unordered atomic-scatter
contract, packed route IDs and routing weights, the sequential FP32 reduction
order, and BF16 output arithmetic.

## ShareGPT performance

The three-commit MoE group was measured against the DSA boundary (measured against PR 4, https://github.com/lightseekorg/tokenspeed/pull/1433):

| Metric | PR 4 | This PR | Change |
|---|---:|---:|---:|
| Output throughput | 1,211.568 tok/s | 1,226.596 tok/s | +1.24% |
| Decode/user | 101.652 tok/s | 103.361 tok/s | +1.68% |
| Mean TPOT | 9.838 ms | 9.675 ms | 1.65% lower |
| Mean ITL | 31.867 ms | 31.340 ms | 1.65% lower |
| Mean TTFT | 480.003 ms | 479.703 ms | 0.06% lower |
| Mean end-to-end latency | 5,507.244 ms | 5,420.548 ms | 1.57% lower |

The smaller reduction tile improved its full stage-two chain by 2.65% at 16
rows and 4.63% at 64 rows. Direct expert-block construction reduced its
42-call graph region by 59.40%. Routing-kernel variants across rows 1-16 fell
from 11 to five. The table is the accepted PR-level performance result; the
individual measurements should not be added together.

The workload used four MI350X GPUs, attention TP4, MoE TP4/EP1, MTP3 with
four-token verification, 16 concurrent requests, and 512 output tokens per
request. The primary operation-boundary measurements were collected over
`upstream/main` `b174a318`. The isolated MoE results used earlier source and
checkpoint stacks; their relative deltas are valid within each experiment, but
their absolute values are not cross-comparable. The current branch applies the
same executable feature changes to newer `upstream/main` `5af23865`, but that
resulting tree was not benchmarked.

### DIFF
--- a/tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/moe/fp16/moe_align_device.py
+++ b/tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/moe/fp16/moe_align_device.py
@@ -188,6 +188,7 @@ def _prepare_small_kernel(
     BLOCK_G: gl.constexpr,
     BLOCK_E: gl.constexpr,
     BLOCK_NB: gl.constexpr,
+    MAX_BLOCKS_PER_EXPERT: gl.constexpr,
     EM_MAX: gl.constexpr,
     INIT_TILE: gl.constexpr,
     EXPERT_START: gl.constexpr,
@@ -197,7 +198,6 @@ def _prepare_small_kernel(
     LG: gl.constexpr = gl.BlockedLayout([1], [64], [NW], [0])
     LE: gl.constexpr = gl.BlockedLayout([1], [64], [NW], [0])
     LI: gl.constexpr = gl.BlockedLayout([1], [64], [NW], [0])
-    LT: gl.constexpr = gl.BlockedLayout([1, 1], [1, 64], [NW, 1], [1, 0])
 
     for r0 in gl.static_range(0, EM_MAX, INIT_TILE):
         r = r0 + gl.arange(0, INIT_TILE, layout=LI)
@@ -226,6 +226,7 @@ def _prepare_small_kernel(
     padded = blocks_pe * block_m
     row_off = gl.associative_scan(padded, 0, _add) - padded
     blocks_incl = gl.associative_scan(blocks_pe, 0, _add)
+    block_start = blocks_incl - blocks_pe
     num_valid = gl.sum(padded, 0)
     num_blocks = gl.sum(blocks_pe, 0)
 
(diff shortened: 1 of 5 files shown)


tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/moe/fp16/moe_align_device.py
```diff
@@ -362,6 +362,7 @@ def moe_align_block_size_device(
             BLOCK_G=triton.next_power_of_2(N),
             BLOCK_E=BLOCK_E,
             BLOCK_NB=64,
+            MAX_BLOCKS_PER_EXPERT=triton.cdiv(N, block_m),
```
borontion, line 368: this line reads weird - why using block_m to divide N?

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/moe/fp16/moe_align_device.py
```diff
@@ -239,24 +240,23 @@ def _prepare_small_kernel(
...
+        block_ids = b0 + gl.arange(0, BLOCK_NB, layout=LE)
+        gl.store(
+            sei_ptr + block_ids,
             gl.full([BLOCK_NB], -1, gl.int32, layout=LE),
+            mask=block_ids < nb_max,
+        )
+
+    # The sentinel fill is owned by one wave; synchronize before other waves
+    # overwrite live block ranges with expert IDs.
+    gl.barrier()
```
borontion, line 252: I don't think we typically need to use `gl.barrier()` - that should be automatically inserted?

Max191: According to the agent, compiling without the `gl.barrier()` doesn't automatically insert synchronization. I believe we need the barrier here. Is there something here that is expected to cause a barrier to be inserted automatically?

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/moe/fp16/moe_align_device.py
Max191: Added a comment here.

PR #1441, opened 2026-09-08
fix(runtime): preserve draft prefill KV write locations

minedec:
## Summary
 Fix Qwen3.5-397B-A17B-NVFP4 MTP acceptance-rate regression on B200 TP8 CI.

## Root Cause
Draft step 0 (which is draft prefill step) locally dispatches attention as DECODE while retaining the full EXTEND K/V rows. The decode request offset incorrectly removed the EXTEND write locations, causing K/V rows to be truncated and skipped.

The missing draft KV writes changed subsequent draft candidates, causing the target model to reject more speculative tokens.

The fix composes the EXTEND span with the DECODE window for draft step 0 while preserving the existing behavior for target decode and later draft steps.

## Example
For a draft step-0 prefill with one request and 22 new tokens:

```text
Q shape:                (1, 2048)
K/V rows:               22
EXTEND write locations: 22
DECODE write locations: 0
```

Before the fix, the local DECODE dispatch selected the empty DECODE window. The TRT-LLM backend therefore truncated K/V from 22 rows to 0 and wrote nothing to the draft KV cache.

After the fix, draft step 0 selects all 22 EXTEND locations. For a MIXED batch, it selects the EXTEND locations 
followed by the DECODE locations, matching the K/V row order.
 
## Test Plan
- E2E correctness test passed.
- TP8 CI performance test passed.
- Average decoded tokens per iteration recovered from 3.27 to approximately 3.6.

### DIFF
--- a/docs/design/unified_path.md
+++ b/docs/design/unified_path.md
@@ -453,7 +453,10 @@ buffer; `fill_input_buffers` takes no table.
   write. A model path that writes multiple mode windows in one shot (the
   MLA draft's step-0 whole-batch write) concatenates the EXTEND span and the
   DECODE window — eager-only, MIXED rounds never run under a captured
-  graph. V4 composes the shared token-shaped resolve
+  graph. The router performs the same composition when a draft step-0
+  forward locally dispatches as DECODE while retaining the round's full K/V
+  rows; target MIXED decode halves and later draft steps keep their ordinary
+  decode-only windows. V4 composes the shared token-shaped resolve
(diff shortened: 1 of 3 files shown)


tuanzhangCS (approved):

tuanzhangCS (approved):

minedec:
> > issue: https://github.com/lightseekorg/tokenspeed/actions/runs/34304628266/job/102370479875?pr=1441 timeout
> 
> @minedec @tuanzhangCS I felt this is a separate issue, so let’s open another PR for it.

ok

PR #1443, opened 2026-09-08
feat: report per-request speculative acceptance

Dogacel:
**Motivation:** Track spec decoding metrics for each request over API.

Previously, spec acceptance was only observable by parsing the per-request `Accept_num_tokens_avg` log line. This reports the counts on the request and fixes two defects in the existing Prometheus counters.

- Added to `meta_info` fields `spec_accepted_tokens` and `spec_draft_tokens`.
- Accept count is `completion_tokens - 1 - spec_verify_ct` (one token from prefill, one bonus per verify step).
- Drafted counts `num_draft_tokens - 1` per verify step, since slot 0 is already accepted.
- `spec_decode_num_draft_tokens` overcounted by one per step; it now uses the same width.

Engine half of smg-project/smg#2481. Refs smg-project/smg#1965.

### Testing

Qwen3.8-27B with Qwen3.8-27B-DFlash2 at 7 steps / 8 draft tokens behind smg-project/smg#2481, for non-streaming and streaming chat and across 8 aiperf requests.


yubofredwang (approved):
Nice feature!

PR #1446, opened 2026-09-09
 refactor(iris): size all-reduce buffers independently and centralize kernel parameters

efric:
## Summary

  Previously, IrisAllReduce collapsed the ordinary and producer-direct capacities into one value:

```
  P = max(
      ordinary_all_reduce_max_numel,
      producer_direct_max_bytes / dtype_size,
  )

```

That single capacity then determined the size of otherwise unrelated buffers. Increasing the producer-direct capacity—for example, to accommodate a large MoE output—also enlarged the ordinary all-reduce and fused AttnRes buffers.

Before the ordinary two-stage path was added, the symmetric payload buffers were effectively:

```
  producer-direct input             P
  producer-direct scratch           P
  ordinary one-shot input slots    2P
  AttnRes input slots              2P
                                  ───
                                   6P
```

Current main also contains an ordinary two-stage all-reduce, which requires a symmetric input buffer and a rank-local partition in symmetric scratch. Under the old shared-capacity model, those were also derived from P.

These buffers have different requirements:

- The producer-direct input must hold the complete producer output.
- Producer-direct two-stage scratch holds only the partition reduced by the current rank, approximately Pproducer / world_size. It is unnecessary when producer-direct uses only the one-stage kernel.
- Ordinary one-shot all-reduce uses two alternating input slots to prevent a new invocation from overwriting data that peers may still be reading.
- Ordinary two-stage all-reduce uses one copied input plus approximately Pstaged / world_size of scratch. Its exit barrier makes a second input slot unnecessary.
- Fused Kimi-K3 AttnRes uses two alternating input slots, but its capacity is limited to the token range supported by that fused kernel.
- The reduced output is ordinary device memory rather than symmetric-heap memory. One output buffer is shared and sized to the largest path that uses it.

This PR gives each path an independent capacity:

```
  producer-direct input             Pproducer
  producer-direct scratch           ceil(Pproducer / world_size), when needed

  ordinary one-shot input slots     2 × Pstaged
  ordinary two-stage input          Pstaged, when supported
  ordinary two-stage scratch        ceil(Pstaged / world_size), when supported

  Kimi-K3 AttnRes input slots       2 × Pattnres
```

The symmetric heap retains its 256 MiB minimum. Above that floor, its size is calculated from the buffers and synchronization state actually required, plus the existing 16 MiB headroom. Increasing one path’s capacity therefore no longer enlarges every unrelated buffer.

For a TP8 Kimi-K3 state sized for both an 8,192-token attention all-reduce and an 8,192-token producer-direct MoE all-reduce:

```
  Pproducer = 8192 × (7168 + 3584) × sizeof(BF16) = 168 MiB
  Pstaged   = 8192 × 7168 × sizeof(BF16)          = 112 MiB
  Pattnres  = 16 × 7168 × sizeof(BF16)            = 0.219 MiB
```

```
   Allocation                              Old shared-capacity layout    Independent capacities
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━
   Producer-direct input                                      168 MiB                   168 MiB
  ──────────────────────────────────────  ────────────────────────────  ────────────────────────
   Producer-direct scratch                                    168 MiB                    21 MiB
  ──────────────────────────────────────  ────────────────────────────  ────────────────────────
   Ordinary one-shot input, two slots                         336 MiB                   224 MiB
  ──────────────────────────────────────  ────────────────────────────  ────────────────────────
   AttnRes input, two slots                                   336 MiB                 0.438 MiB
  ──────────────────────────────────────  ────────────────────────────  ────────────────────────
   Ordinary two-stage input                                   168 MiB                   112 MiB
  ──────────────────────────────────────  ────────────────────────────  ────────────────────────
   Ordinary two-stage scratch                                  21 MiB                    14 MiB
  ──────────────────────────────────────  ────────────────────────────  ────────────────────────
   Synchronization flags                                   ~1.320 MiB                ~0.881 MiB
  ──────────────────────────────────────  ────────────────────────────  ────────────────────────
   Automatic symmetric-heap reservation                     1,360 MiB               556.319 MiB
  ──────────────────────────────────────  ────────────────────────────  ────────────────────────
   Shared local reduced-output buffer                         168 MiB                   168 MiB
  ──────────────────────────────────────  ────────────────────────────  ────────────────────────
   Total persistent state                                   1,528 MiB               724.319 MiB
```
 
The old heap calculation conservatively charged the ordinary two-stage scratch as another full P, even though its actual allocation was P / world_size.

For the same 8,192-token capacity:

- Symmetric-heap reservation falls from 1,360 MiB to 556.319 MiB, a 59.1% reduction.
- Total persistent all-reduce state falls from 1,528 MiB to 724.319 MiB, a 52.6% reduction.

This makes larger Iris payloads practical where the kernels already support them but the previous shared-capacity model made increasing max_bytes unnecessarily
  expensive.

The PR also puts each kernel’s settings in one place. Buffer allocation and kernel launch now read the same definitions for packed-word geometry, supported group sizes, workgroup shape, program limits, and producer-direct switching thresholds. A kernel-setting change therefore cannot silently diverge from the workspace sizing that supports it.

## Test Plan

The normal staged all-reduce algorithm and existing dispatch thresholds remain unchanged. Tests cover protocol-specific state construction, exact workspace sizing, producer-direct admission and thresholds, supported dtypes and group sizes, and multi-GPU correctness. No performance optimizations in this PR.


nirvedhmeshram (approved):
LGTM

efric:
superceded by #1465 (needed to change branch configuration to start a stack to help with dependent PR management)

PR #1452, opened 2026-09-09
refactor(l2): name every stream the executor orders against; refuse cache ops with no transfers

FC-Li:
## Summary

Two related tightenings of the L2 (Host cache) executor's contracts, both replacing implicit state with an explicit argument or an explicit failure.

### 1. Every stream the executor orders against is named by the caller

`submit_write_backs` takes `fence_stream` -- the stream a stream-ordered op's completion fences -- and `submit_load_backs` takes `producer_stream` -- the stream that zeroed its destinations -- mirroring how the write side already names the stream that wrote its sources (#1450).

Until now the fence (`device_module.current_stream().wait_event(fence)`) and the load start event (`start_event.record()`) landed on whatever stream was current on the forward thread. That was correct only by convention: `ForwardThread` never sets a stream, so "current" resolves to the device's default stream, which is also where `zero_cache_pages` runs. Nothing in the code expressed that dependency. The call site in `DeviceHandle.execute` now reads as the argument -- zeroing runs on `executor.default_stream`, the fence lands on `executor.default_stream`, the loads order after `executor.default_stream`.

`shutdown` synchronizes the device instead of the caller's current stream (and no longer bypasses `device_module` via `torch.cuda`). The tests that patched `current_stream` to inject a fake now pass the stream in; one asserts the executor never consults it.

### 2. Cache ops with no transfers are refused, not acknowledged

`WriteBackBatch` / `LoadBackBatch` deduplicated `(group, source, destination)` across the ops of one plan, so the runtime had to accept an op whose page lists came through empty and acknowledge it at the next poll without a copy (`_ready_write_op_ids` / `_ready_load_op_ids`). Neither path is reachable: `StartPendingStores` skips keys already in flight (`store_keys_`) and dedups keys within a call, and each load pair's destination is a freshly acquired page (`AppendHostExtension`), so a transfer cannot repeat within a plan and an op cannot arrive empty.

The batch constructors now `_assert` both invariants, the runtime raises on an op with no transfers, and the side channel in `poll_results` is gone. An op is acknowledged by its copy's completion event alone; an op that never copies would hold its tickets forever, so a violation fails loudly rather than being papered over. The invariant is recorded in `docs/design/cache-concepts.md`.

The two halves are independent: the runtime change is safe against the currently pinned scheduler (it never emits an empty op), and the scheduler change ships with its next release.

Also drops the `stream` local alias in `_start_writing` in favour of the explicit `self.write_stream`, matching `_start_loading`.

## Test plan

- [x] `tokenspeed_scheduler_tests`: 462 gtest cases pass, including the rewritten `CacheOperationTest` (flatten order, repeated transfer refused within and across ops, empty op refused). The new assertions run inside every scheduler-plan scenario test without firing.
- [x] `tokenspeed-scheduler/python/tests`: 65 binding tests pass against the rebuilt extension.
- [x] `test/runtime/test_host_executor.py` + `test/runtime/test_device_handle.py`: 47 cases pass; `test_l2_cache_hooks.py` passes.
- [x] `pre-commit run --all-files` clean.
- [ ] End-to-end run with the Host cache enabled (kvstore config) to confirm the assertions stay silent under real load.

### DIFF
--- a/python/tokenspeed/runtime/cache/l2/executor.py
+++ b/python/tokenspeed/runtime/cache/l2/executor.py
@@ -177,17 +177,17 @@ class L2CacheExecutor:
             tracker = LayerwiseLoadTracker(len(layout.consumers))
             pool.register_layerwise_load_tracker(tracker)
             self._load_trackers.append((tracker, len(layout.consumers)))
-        # Write-backs run on their own stream, ordered after the producer
-        # stream the caller names per submission -- the one the forwards wrote
-        # the pages on. What differs per op is who waits on the copy: a
-        # stream-ordered op (a retraction's snapshot, whose sources this very
-        # plan may re-grant) fences the fence stream the caller names on its
-        # completion, so the plan's zeroing, load-backs and forwards stay
-        # behind it; a pinned op (an ordinary publication, whose sources the
-        # scheduler holds until the ACK) fences nothing and never holds up the
-        # round. Loads keep their own stream, ordered after the producer stream
-        # that zeroed their destinations: their consumers are fenced per layer
-        # by the tracker events.
+        # Every copy runs on its own stream, ordered after the prerequisite
+        # stream the caller names per submission: for a write-back the one
+        # the forwards wrote the source pages on, for a load the one that
+        # zeroed the destination pages. What differs per write-back op is who
+        # waits on the copy: a stream-ordered op (a retraction's snapshot,
+        # whose sources this very plan may re-grant) fences the fence stream
+        # the caller names on its completion, so the plan's zeroing,
+        # load-backs and forwards stay behind it; a pinned op (an ordinary
+        # publication, whose sources the scheduler holds until the ACK) fences
+        # nothing and never holds up the round. A load's consumers are fenced
+        # per layer by the tracker events.
         self.write_stream = _new_cache_stream(None)
         self.load_stream = _new_cache_stream(_load_stream_priority())
         device = self.layout.buffers[0].device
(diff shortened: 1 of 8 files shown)


OftenDream (commented):
One non-blocking naming suggestion for the explicit stream contract.

python/tokenspeed/runtime/cache/l2/executor.py
```diff
@@ -322,16 +322,23 @@ def submit_write_backs(self, plan, *, producer_stream) -> None:
...
         self._start_writing(
             pinned_op_ids,
             pinned_transfers,
             lane=self._pinned_write_lane,
             producer_stream=producer_stream,
         )
 
-    def submit_load_backs(self, plan) -> None:
-        """Launch the plan's H2D loads; runs after the plan's page zeroing."""
+    def submit_load_backs(self, plan, *, producer_stream) -> None:
```
OftenDream: A small naming question: I found `producer_stream` a little harder to follow on the load side. For writeback it refers to the stream producing the source KV, while here it refers to the stream preparing the destination pages. Would it be worth making that distinction a bit clearer? The ordering itself makes sense to me;

OftenDream: Perhaps a shared name like prerequisite_stream？

FC-Li: Agreed, and taken as suggested in 138d483a: `producer_stream` is now `prerequisite_stream` on both sides. "Producer" described what the caller had done on the stream, which only fit the write side; the executor does the same thing with it either way -- orders the copy after that stream's completed work -- so the name now describes that role, paired with `fence_stream` (the stream the copy's completion fences). Each docstring says which stream it is in practice: the execution stream that wrote the source pages for write-backs, the default stream that zeroed the destination pages for load-backs.

PR #1454, opened 2026-09-09
feat(kimi-k3): support Hopper PD with DeepEP and DSpark

sjqgogogogo:
## Summary

Enable Kimi-K3 MXFP4 serving on Hopper (H100/H200) with DeepEP, pipeline-parallel prefill and DSpark speculation. The changes address MoE token ownership across TP/DP/EP, production of draft context across pipeline stages, PD progress on idle decode ranks, and installation with CUDA 12.9.

### 1. MoE token layout and K3 DeepEP support

K3 attention and shared experts operate within an attention-TP replica, while DeepEP routes tokens across the expert-parallel group. Supporting attention DP requires explicit token ownership through dispatch, expert computation and combine. The Hopper Marlin bridge also needs to handle DeepEP's normal and low-latency receive layouts correctly.

- Split each attention batch into disjoint TP token slices before routing and latent projection. After DeepEP combine, gather latent results within the source attention-TP group; keep shared-expert and output-projection reductions within that replica. This avoids replicating full hidden states and residuals across attention-DP batches.
- Consume the local expert IDs returned by normal dispatch directly. For low latency, use GPU receive counts to pack valid expert rows, construct an aligned Marlin schedule, apply activation to valid rows, and unpack results into the receive buffer. Intermediate storage is bounded by source routes instead of the full expert-by-receive-capacity extent; low-latency combine applies routing weights once.
- Derive persistent DeepEP send capacity from the configured per-DP verify workload after TP slicing, including recovery chunks when low latency is pinned. Prepare communication buffers before KV memory profiling.
- Keep empty token slices and idle DP ranks in the EP collectives. Mask CUDA-graph padding using live device cache-write locations, so replay can handle changing live counts within a capture bucket.

### 2. DSpark with pipeline-parallel prefill and early context projection

DSpark consumes hidden-state taps from multiple target layers. With PP, those layers and the weights needed to interpret their taps reside on different stages, so collecting and projecting all taps at one stage is not a suitable ownership model.

- Project each tap on its owning stage as soon as it becomes available. Carry one FP32 `[tokens, draft_hidden]` accumulator alongside the ordinary per-forward PP state. Apply per-tap normalization before projection and final context normalization after the complete sum.
- Evaluate boundary AttnRes taps on the stage owning the next layer's mixing weights. Each stage loads its own tap-projection parameters; the final stage additionally owns the DSpark proposal network, embedding shard and draft cache.
- Use explicit target/draft cache-layer ownership for physical allocation, producer readiness and PD transfer planning. Draft cache layers do not change the target's PP partition boundaries.
- On the final P stage, write prompt context KV and run the ordinary DSpark proposal path. Publish the final producer barrier after drafting, then transfer the sampled anchor and real candidates with the target/draft cache state. D installs this candidate window before its first ordinary verify round.
- Reserve the transient proposal-write workspace after every prefill chunk through the existing scheduler cache groups, and include it in the single-request capacity bound.

Pipeline speculation is enabled for K3 DSpark on prefill workers, with an unquantized draft checkpoint, attention CP=1 and matching dense/attention TP groups.

### 3. Keep PD bootstrap/preallocation progressing on idle decode ranks

When one attention-DP rank had no local model tokens while other ranks were active, the event loop marked it as an idle round and skipped PD transfer-event polling. A completed bootstrap could therefore remain unreported to the scheduler, preventing the next remote-prefill admission/preallocation step; completed incoming transfers could also wait without advancing to decode.

Separate an explicit pause from the need for an idle model forward. An idle forward substitutes only the model batch: the scheduler plan, cache operations, PD event polling and scheduler feedback continue through the same round. Add per-DP/PP lifecycle counts and bootstrap diagnostics to make stalled handshakes and transfer-pinned requests visible.

### 4. CUDA 12.9 installation support

Hopper deployments using CUDA 12.9 need a consistent set of Torch, native wheels and locally compiled kernels. Reusing editable-install artifacts after switching CUDA toolchains can otherwise leave incompatible binaries in the checkout.

- Add `CUDA_VARIANT=cu129 bash test/ci_system/install_deps.sh` for Linux/Python 3.11 environments. Validate the toolkit and active venv, select CUDA 12.9 package sources explicitly, and reject an existing CUDA 13 package mix.
- Use the CUDA 12 CuTe runtime, omit the CUDA-13-only KDA AOT package, build in-tree kernels for `sm90a`, and rebuild the scheduler from the checkout.
- Record successful kernel build identities and invalidate cached binaries when the toolkit, compiler or build flags change. Finish installation with `pip check`.

The [Hopper PD serving guide](https://github.com/lightseekorg/tokenspeed/blob/50001c59e51e14e86d123a00a0005fb1eac639a3/docs/guides/kimi-k3-hopper-pd.md) and [CUDA 12.9 installation guide](https://github.com/lightseekorg/tokenspeed/blob/50001c59e51e14e86d123a00a0005fb1eac639a3/docs/guides/hopper-cu129.md) document the configuration and validation steps.

## Test Plan

- Functional correctness validated before submission.
- `pre-commit run --all-files` passed for the current commit.
- The full runtime/GPU suite was not rerun during PR preparation. The serving guide lists focused checks for normal/low-latency DeepEP, changing live counts under graph replay, PP projection equivalence, candidate handoff and repeated cache-slot reuse.
- Known test mismatch: `test_tap_count_must_match_num_target_layers` still expects the previous validation error text.

### DIFF
--- a/python/tokenspeed/runtime/execution/drafter/kimi_k3_dspark.py
+++ b/python/tokenspeed/runtime/execution/drafter/kimi_k3_dspark.py
@@ -1,0 +1,77 @@
+# Copyright (c) 2026 LightSeek Foundation
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+"""K3 model wiring around the ordinary DSpark proposal path."""
+
+from __future__ import annotations
+
+import torch
+
+from tokenspeed.runtime.execution.context import ForwardContext
+from tokenspeed.runtime.execution.drafter.dspark import DSpark
+from tokenspeed.runtime.layers.logits_processor import LogitsProcessorOutput
+
+
+class K3DSpark(DSpark):
+    """K3 owns capture configuration; candidate generation stays on DSpark."""
+
+    @staticmethod
+    def configure_target(target_model, draft_model) -> None:
+        """Configure K3 taps once after loading, including non-drafting PP stages."""
+        capture_layer_ids = list(draft_model.target_capture_layer_ids)
+        stream = draft_model.config.aux_hidden_stream
+        if draft_model.mapping.has_pp:
+            target_model.set_prefill_context_capture(
+                capture_layer_ids, stream, draft_model.hidden_size
+            )
+        else:
+            target_model.set_dflash_layers_to_capture(capture_layer_ids)
+            target_model.set_dflash_aux_hidden_stream(stream)
+
+    def wire_target(self, target_model) -> None:
+        """Bind execution resources without replacing the configured K3 taps."""
+        language_model = getattr(target_model, "language_model", target_model)
+        self.target_model = target_model
+        self.target_language_model = language_model
+        self.embed_tokens = (
+            self.model.embed_tokens
+            if self.model.mapping.has_pp
+            else target_model.get_input_embeddings()
+        )
+        self.lm_head = target_model.lm_head
+        self.logits_processor = language_model.logits_processor
+
+    def _update_native_cache_from_target(
+        self,
+        base_ctx: ForwardContext,
+        logits_output: LogitsProcessorOutput,
+        accept_lengths: torch.Tensor,
+    ) -> None:
+        if not self.model.mapping.has_pp:
+            super()._update_native_cache_from_target(
+                base_ctx, logits_output, accept_lengths
+            )
+            return
+        # PP capture already wrote this chunk's context on our stream. Seed
+        # the ordinary draft block at the complete prefix, including cached
+        # and chunked history, without projecting those taps a second time.
+        self.draft_seq_lens_buf[: base_ctx.bs].copy_(
+            self.input_buffers.seq_lens_buf[: base_ctx.bs]
+        )
--- a/python/tokenspeed/runtime/models/kimi_k3.py
+++ b/python/tokenspeed/runtime/models/kimi_k3.py
@@ -56,7 +56,7 @@ Module hierarchy matches the checkpoint::
 from __future__ import annotations
 
 import logging
-from collections.abc import Iterable
+from collections.abc import Callable, Iterable
 from typing import TYPE_CHECKING
 
 import torch
@@ -126,7 +126,12 @@ from tokenspeed.runtime.layers.moe.latent import (
 from tokenspeed.runtime.layers.moe.loader import build_moe_checkpoint_loader
 from tokenspeed.runtime.layers.moe.schema import ExpertCheckpointSchema
 from tokenspeed.runtime.layers.moe.topk import TopK, TopKOutput, TopKOutputFormat
-from tokenspeed.runtime.layers.moe.utils import RoutingMethodType, get_moe_backend
+from tokenspeed.runtime.layers.moe.utils import (
+    RoutingMethodType,
+    get_all2all_backend,
+    get_moe_backend,
+    use_deepep_low_latency,
+)
 from tokenspeed.runtime.layers.quantization.base_config import QuantizationConfig
 from tokenspeed.runtime.layers.quantization.modelopt_mixed import (
     preprocess_fp8_pb_wo_weights,
@@ -231,7 +236,13 @@ class KimiLinearMLP(nn.Module):
     ) -> None:
         super().__init__()
         self.mapping = mapping
-        if is_shared_expert:
+        if is_shared_expert and get_all2all_backend().is_deepep():
+            # Each attention-DP batch has its own shared expert computation.
+            # DeepEP only owns the routed experts, never these TP partials.
+            tp_rank = mapping.attn.tp_rank
+            tp_size = mapping.attn.tp_size
+            tp_group = mapping.attn.tp_group
+        elif is_shared_expert:
             tp_rank = mapping.moe.tp_ep_rank
             tp_size = mapping.moe.tp_ep_size
             tp_group = mapping.moe.tp_ep_group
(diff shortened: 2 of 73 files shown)


python/tokenspeed/runtime/models/kimi_k3.py
```diff
@@ -1460,7 +1471,21 @@ def __init__(
         # Attention DP partitions the batch when its TP group is smaller than
         # the MoE TP×EP group. Gather those DP shards so every rank enters the
         # K3 MoE with the same complete token batch.
-        self._gather_dp_tokens_for_moe = mapping.attn.tp_size != mapping.moe.tp_ep_size
+        self._use_deepep = get_all2all_backend().is_deepep()
```
syuoni, line 1474: The repeated `self._use_deepep` checks spread DeepEP-specific token layouts, projection sharding, and communication logic throughout KimiLinearMoE.

For this PR, I suggest encapsulating that logic in a dedicated `KimiLinearMoEDeepEP` module, selected at construction time. This would make the new implementation easier to follow and review, while minimizing changes to the existing paths.

python/tokenspeed/runtime/execution/drafter/kimi_k3_dspark.py
```diff
@@ -0,0 +1,77 @@
...
+from __future__ import annotations
+
+import torch
+
+from tokenspeed.runtime.execution.context import ForwardContext
+from tokenspeed.runtime.execution.drafter.dspark import DSpark
+from tokenspeed.runtime.layers.logits_processor import LogitsProcessorOutput
+
+
+class K3DSpark(DSpark):
```
syuoni, line 32: This subclass mainly handles PP-prefill integration rather than K3-specific drafting behavior. 

Could we handle the PP-specific embedding selection and cache preparation in the existing DSpark/setup code, while keeping K3’s capture and projection details in the model implementation?

python/tokenspeed/runtime/models/kimi_k3.py
```diff
@@ -231,7 +236,13 @@ def __init__(
...
         super().__init__()
         self.mapping = mapping
-        if is_shared_expert:
+        if is_shared_expert and get_all2all_backend().is_deepep():
+            # Each attention-DP batch has its own shared expert computation.
+            # DeepEP only owns the routed experts, never these TP partials.
+            tp_rank = mapping.attn.tp_rank
+            tp_size = mapping.attn.tp_size
+            tp_group = mapping.attn.tp_group
+        elif is_shared_expert:
```
syuoni, line 245: The DeepEP path changes token ownership, shared-expert and latent-projection sharding, and communication scheduling. These decisions are spread across `KimiLinearMoE`, `KimiLinearMLP`, and the communication code. In particular, `KimiLinearMLP` now determines its weight sharding from the global all-to-all backend. 

Could we encapsulate the DeepEP-specific composition in a dedicated KimiLinearMoEDeepEP, selected at construction, and pass the intended parallel group explicitly to its shared MLP? This would make the weight layout and corresponding communication behavior understandable together.

python/tokenspeed/runtime/models/kimi_k3.py
```diff
@@ -2785,25 +2896,65 @@ def forward(
...
+                        context_producer.add_capture(
+                            projected_context, capture_idx, captured
+                        )
+                    else:
+                        if ctx.target_capture_sink is not None:
+                            ctx.target_capture_sink.on_target_capture(
+                                capture_idx, captured
+                            )
+                        assert aux_hidden_states is not None
+                        aux_hidden_states.append(captured)
```
syuoni, line 2957: Could we make capture and per-tap projection a uniform operation for PP and non-PP, owned by the layer that produces the captured stream?

For an AttnRes tap labelled L, we could consistently capture/project at L+1’s entry, with the final tap handled by the output mixer, while preserving checkpoint tap indices. PP would pass the projected accumulator between stages; non-PP would keep it local. 

This would remove the separate PP-boundary deferral/recovery logic and let capture execution and projection-weight placement follow the same ownership rule.

syuoni (commented):
Thanks @sjqgogogogo for your contribution!

Left some comments from the modeling perspective.

python/tokenspeed/runtime/models/kimi_k3.py
```diff
@@ -1640,7 +1628,11 @@ def __init__(
         self.shared_experts = KimiLinearMLP(
             hidden_size=config.hidden_size,
             intermediate_size=config.moe_intermediate_size * config.num_shared_experts,
-            mapping=mapping,
+            tp_rank=0 if mapping.attn.dp_size > 1 else mapping.moe.tp_ep_rank,
+            tp_size=1 if mapping.attn.dp_size > 1 else mapping.moe.tp_ep_size,
+            tp_group=(
+                (mapping.rank,) if mapping.attn.dp_size > 1 else mapping.moe.tp_ep_group
+            ),
```
syuoni: NIT: `tp_group=None if mapping.attn.dp_size > 1 else mapping.moe.tp_ep_group`

syuoni (dismissed):
LGTM, thanks @sjqgogogogo !

FC-Li:
@sjqgogogogo I went through the CI failures and the branch. Your commits plus 21 follow-up commits (one topic each) are now up as #1634 against `main`, which supersedes this PR and runs the full CI directly; the same branch is also offered against `feature/k3_on_hopper` as sjqgogogogo/tokenspeed#1 if you prefer to keep this PR. The #1634 description has the full list; the short version:

**Why CI was red**
- `ut-runtime-1gpu`: `test_dspark_wire_target_uses_draft_head` still expected `wire_target` to install capture layers, which this branch moved into `configure_target`. Test aligned.
- `ut-runtime-2gpu` hanging 1800 s on every run: a real bug, in the kernel package rather than in your changes. `ops/{residual,gemm}/gluon.py` and `ops/sampling/gluon/__init__.py` keep the caught `ImportError` for the optional AMD package in a module global. The exception's traceback references every frame that was importing at the time, so on NVIDIA the first caller of `import tokenspeed_kernel` has its frame pinned for the process lifetime. Making `tokenspeed.runtime.distributed` lightweight moved that first import from the test worker's setup into the test body, whose local `CUDAGraph` (with a captured NCCL all-gather) then never died, and `ncclCommFinalize` waited for it forever. Fixed by keeping only `str(exc)`, with a regression test.
- The merge conflict with `main` is #1618 moving the capacity formula into `CapacityModel`; merged.
- Two more failures fail-fast had hidden: `test_kimi_k3_moe_attn_dp` (still expected `tp_group == (1,)` after the `None` change) and a `test_cache_pd_executors` fixture missing `cache_fields_by_stage`.

**`prefill_workspace_tokens`** — we traced what the P-side proposal actually writes: the block drafter's `W-1` rows land in the `decode_input_tokens` slots behind the prompt and are read back by paged attention in the same forward. The fused role already reserves exactly that window on the completing chunk; P was the only role scheduled with `decode_reserve=0`, and the mechanism is the same on PP=1 and for the other block drafters. So the PR gives P the fused rule (`buildPrefillWorkerPlan` reserves `config_.decode_input_tokens` on the completing chunk, `CapacityModel` role-neutral for the decode width) and removes the workspace field, its validation, binding and runtime resolver. Scheduler bumped to 0.1.18 for the lifecycle counters you added.

**Behavior worth your attention**
- `DSparkContextProducer` was enabled for every unquantized K3 draft, PP or not; on a fused PP=1 engine that switched off the drafter's fused-KV overlap and made prefill-graph capture (FULL) disagree with the contexts (NULL). It is now selected from `pp_size > 1` and the draft must implement `DSparkContextModel` (runtime-checkable) — no `getattr` probe.
- Three prefill-role facts were gated on `pp_size > 1` (RSAG prewarm, KDA verify staging, linear-backend verify width); they now follow `disaggregation_mode == "prefill"`.
- `pp_layer_partition` had no reader left after the placement change; removed from registration, bootstrap and topology.

**Scope** — the CUDA 12.9 install recipe is carried over unmodified to #1633 (draft) as its own change; `distributed/partition.py`, the dead `KimiLinearMoE` DeepEP branch, `test_draft_model_setup.py`, the 120-line bootstrap diagnostics and the per-PP-stage batch logging are trimmed per the commit messages.

**Two things only you can answer**
1. Low-latency combine dropping `moe_origin_input` vs. `deep_ep.py`'s note that the pinned DeepEP asserts `x_ori` for non-zero weights (Codex's P1): which API does the pinned wheel expose?
2. `TargetCaptureConfigurator` covers DFLASH/DSPARK while Eagle3 keeps `set_eagle3_layers_to_capture` — intentional?

The PP serving validation from your guide should be re-run on the merged state (P reserves the decode slot on the completing chunk only; the producer is off outside PP). Local results are in the PR description: scheduler C++/Python, kernel and two-GPU tests pass, and the runtime-1gpu file set passes except for files that fail identically on `main` in my environment.

PR #1455, opened 2026-09-09
fix(amd): distribute gfx1250 MXFP4 gather/scatter indices across warps

jerryyin:
The gfx1250 MXFP4 index layout gave a single warp ownership of the entire gather/scatter index list: `BlockedLayout([NUM_INDICES, 1], [1, 32], [1, NUM_WARPS], [1, 0])` sliced on dimension 1 leaves every index in one thread of one warp, replicated to all warps. The TDM lowering predicates off warps whose warp-id bits are free variables of the index tensor, because warps holding identical indices would issue identical redundant transfers, so exactly one warp ended up fetching X for the whole workgroup while the rest waited.

Return the row-distributed layout instead and slice dimension 0 in all three consumers, so warp `w` owns only its own slice of the index rows and every warp issues its share of the transfer. Guard the helper with the divisibility and positivity assumptions the partition relies on, and record the mechanism in a comment beside the layout it constrains.

Add `test_gfx1250_mxfp4_index_contract.py` as the off-device home for gfx1250 MXFP4 index invariants. It covers warp ownership here; further index invariants belong in the same file rather than in new ones.

### DIFF
--- a/tokenspeed-kernel/test/ops/moe/test_gfx1250_mxfp4_index_contract.py
+++ b/tokenspeed-kernel/test/ops/moe/test_gfx1250_mxfp4_index_contract.py
@@ -1,0 +1,127 @@
+# Copyright (c) 2026 LightSeek Foundation
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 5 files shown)


tokenspeed-kernel/test/ops/moe/test_gfx1250_mxfp4_index_contract.py
```diff
@@ -0,0 +1,127 @@
...
+"""Off-device contract for gfx1250 MXFP4 index ownership, width and narrowing."""
+
+from __future__ import annotations
+
+import ast
+import sys
+from pathlib import Path
+
+import pytest
+from triton._C.libtriton import gluon_ir, ir
```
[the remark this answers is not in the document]
jerryyin: Addressed in a9e162f

tokenspeed-kernel/test/ops/moe/test_gfx1250_mxfp4_index_contract.py
borontion, line 1: hi @jerryyin. this seems adding checks at gluon ir level? can we just do check on raw ttgir string from compiled kernel? 

I mainly don't want to directly use things from `tokenspeed_triton._C.libtriton`

jerryyin: That's a fair comment. This is the first of its kind test that import gluon IR and reason on that level. 

Regarding raw TTGIR string, it isn't the right candidate for this particular layout check: Either the buggy and fixed layout carry the same layout string (same warpsPerCTA). What matters here is order, sliced dimension and warp ownership. They are derived from the layout manipulation, not a field in the text directly. String matching on that level won't be able to catch the problem exposed here.

Given this, I think the compromise I can take is to check whether rows are truly distributed among warps, and only warps. This can correctly differentiate whether workload is duplicated among all threads or evenly distributed against warps.

Addressed in 46c2b4b

PR #1457, opened 2026-09-09
feat(k3): reserve Iris buffers before KV cache sizing

efric:
## Summary

Previously, Iris allocated its persistent all-reduce buffers lazily on the first reduction. This happened after KV-cache sizing, so the cache planner could claim memory that Iris would later need. A sufficiently large communication allocation could then reduce the intended memory headroom or fail at runtime.

This PR adds an explicit communication preparation phase after model weights are loaded and before KV-cache capacity is calculated:

```
  load model weights
      -> prepare persistent communication buffers
          -> measure available memory
              -> allocate KV cache
```
 
For Kimi K3 on AMD CDNA4, the model calculates the required capacities for:

- Ordinary staged all-reduce
- Producer-direct all-reduce
- Fused attention/AttnRes

The capacities cover the configured maximum forward size, capped at 8192 tokens for the current Iris path. The resulting Iris state and symmetric heap are allocated up front and reused by later reductions.

The preparation API is exposed through the communication backend, while unsupported models, platforms, group layouts, and backends remain no-ops. The NVIDIA path is unchanged.

This ensures that KV-cache sizing accounts for persistent communication memory and that the largest required Iris symmetric heap is created before any smaller lazy state can initialize the shared Iris context.

## Test Plan

Coverage included:

  - Iris buffer preparation and state reuse
  - Rejection of undersized existing states
  - Preservation of the existing 512 KiB ordinary all-reduce dispatch limit
  - Ensuring larger preallocated capacity does not silently expand dispatch eligibility
  - AMD versus non-AMD backend routing
  - Communication allocation occurring before KV-cache planning
  - Kimi K3 capacity calculations for its attention and MoE shapes


borontion:
@efric do you want to make this a stacked PR: https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests

nirvedhmeshram (dismissed):
LGTM

efric:
@nirvedhmeshram sorry it seems that resolving the conflict automatically dismissed the review; however it's NFC

nirvedhmeshram (approved):

nirvedhmeshram:
> @nirvedhmeshram sorry it seems that resolving the conflict automatically dismissed the review; however it's NFC

yes thats a know quirk of the repo :)

PR #1473, opened 2026-09-10
feat(models): support DeepSeek V4 Flash Vision

FlamingoPg:
## Summary

Add DeepSeek V4 Flash Vision support using upstream-preprocessed patches and expanded image blocks. Every image-block position uses the in-vocabulary `<｜deepseek_image｜>` token; per-item `types` metadata identifies the learned sentinels and image features. The reusable vision encoder uses the shared `VisionAttention` backends, and the multimodal wrapper reuses the V4 text model.

Keep image spans intact across chunked prefill and prefix-cache matches, rejecting blocks larger than the prefill budget. Preserve bidirectional image visibility and visual MoE routing through mixed batches and distributed routing. Encoded-image deduplication includes output length so differently aligned copies of an image cannot share incompatible blocks. The configuration enum is removed, and generic draft initialization remains unchanged.

Companion preprocessing PR: https://github.com/smg-project/smg/pull/2499

## Validation

- 51 targeted multimodal tests passed; selected router/metadata regressions passed (7 tests, 1 skipped).
- 24 real-PNG SMG-to-runtime cases and 5 repeated-image alignment/cache cases passed. Tests cover prompt expansion, token substitution, chunked and mixed prefills, image embedding and cached reuse.
- 128 exact old/new routing comparisons and two-rank NCCL routing checks passed, including non-hash visual gates with unequal rank token counts.
- Broader multimodal/DSpark regressions: 114 passed; one DSpark proposal bitwise floating-point equality test failed identically on the base and updated revisions.
- Previously verified scheduler regressions: 470 C++ and 72 Python tests passed. Vision FP16/BF16 comparisons and checkpoint weight-shard checks passed.

Full checkpoint serving/evaluation was not rerun for the placeholder representation change.


tokenspeed-scheduler/csrc/scheduler/operations/forward.cpp
```diff
@@ -370,14 +381,23 @@ std::optional<fsm::SchedulePrefillFirstChunkEvent> Scheduler::schedulePrefillFir
         if (shouldSplitFinalStateCheckpoint(config_, coordinator_)) {
             final_tail_tokens = FinalAlignedTailTokens(hit_tokens, unscheduled, remaining,
                                                        coordinator_.PrefixGranularity(), promotion_boundary_tokens);
+            if (final_tail_tokens) {
```
wangbo981016: Could we extract FinalAlignedTailTokens() and the subsequent unsplittable-span check into a small helper that returns the final std::optional<std::int32_t>? This logic is duplicated in the first-chunk and subsequent prefill paths, and sharing it would keep both paths consistent when the splitting policy changes.

tokenspeed-scheduler/csrc/scheduler/operations/forward.cpp
```diff
@@ -297,6 +297,17 @@ Scheduler::AdmissionMatch Scheduler::matchPrefixAtAdmission(Request* request) {
         return match;
     }
     match.probe = probe(probe_hashes);
+    while (true) {
```
wangbo981016: Maybe guard this adjustment with !request->UnsplittableSpans().empty() and use while (hit_tokens > 0)? This would make the intent clearer: only requests with unsplittable spans need boundary adjustment, and the loop keeps shrinking an existing cache hit until its boundary is valid or no hit remains.

PR #1487, opened 2026-09-10
ci(amd): Add registration-level AMD kernel benchmark CI

Max191:
## Summary

This adds the first registration-level kernel benchmark path for TokenSpeed and
an AMD pull request job that compares a candidate revision with its merge base.

- Add an operation-neutral benchmark harness with exact registration selection,
  fresh-input correctness checks, graph-replay device timing, and structured
  results.
- Add the first operation generator and suite entry for a BF16 batched GEMM on
  `gfx950` using `gluon_bmm_a16w16_gfx950`.
- Add revision-isolated base/candidate execution, regression and noise policy,
  job summaries, retained result artifacts, and a trusted pull request comment
  publisher.

This PR does not add benchmark history storage or graphs.

## Benchmark contract

Each benchmark case declares:

- an operation family and mode;
- operation-owned input parameters;
- an optional solution or exact registration override;
- timing settings; and
- per-case regression and noise limits.

The operation generator creates the opaque callable, inputs, and local reference
call. The harness then validates fresh input sets before timing the exact selected
registration. Compilation, input creation, correctness checks, warmup, and graph
capture are excluded from the reported device time.

Correctness remains local to each revision and process. The initial `close`
validator compares the selected GEMM registration with the registered `torch_bmm`
reference over five fresh input sets. Tensor data never crosses revision process
boundaries; only bounded JSON results and logs do.

## Pull request comparison

The AMD job checks out the pull request merge revision as the coordinator, then
runs the merge base and exact candidate commit in separate worktrees, Python
environments, and compilation caches. Each revision reads its own suite and
requirements and invokes its own harness.

Only cases with the same stable ID, definition, timing context, registration,
and hardware are compared. Added and changed cases are reported without
comparing incompatible measurements. Missing baseline cases make the result
inconclusive instead of silently disappearing.

The initial suite records 30 graph-replay timing samples, with 100 kernel calls
per captured graph. A result blocks only when the median slowdown exceeds both
10% and 0.5 microseconds. A relative median absolute deviation above 5% is
reported as inconclusive rather than as a regression.

The process exit contract is:

| Exit | Meaning |
| ---: | --- |
| 0 | No blocking regression, including reported but non-blocking inconclusive cases |
| 1 | Performance regression |
| 2 | Invalid comparison, such as correctness, environment, or infrastructure failure |

## CI and security model

Untrusted pull request code runs only in the measurement workflow, which has
read-only repository permission. It uploads a bounded report artifact but cannot
write to the pull request.

A separate `workflow_run` job checks out the publisher from the default branch,
validates artifact identity and freshness, and creates or replaces one bot-owned
summary comment. This gives the publisher pull request write permission without
giving that permission to measured pull request code.

The benchmark runs only for non-draft, same-repository pull requests that touch
kernel or benchmark infrastructure. A new commit cancels an older benchmark run
for the same pull request.

## Validation

- The full repository pre-commit hook set passed.
- The new and modified workflows passed `actionlint`; the environment installer
  passed `ShellCheck`.
- 279 focused harness, comparison, publisher, registration, and selection tests
  passed after rebasing onto current `upstream/main`.
- 269 adjacent benchmark, numerical, and kernel-selection tests passed, with 54
  hardware or parameter-set skips.
- The exact first-PR bootstrap path succeeded on a `gfx950` AMD GPU: correctness
  passed, 30 samples were produced, and the new case was reported as `added`.
- A separate two-revision, two-environment rehearsal completed within budget at
  -0.12% (`3.469 us` base, `3.465 us` candidate).
- A broader local fault-injection campaign produced 20/20 stable same-code
  comparisons, rejected 3/3 forced slowdowns, and exercised correctness,
  environment, malformed-result, suite-evolution, and comment-publisher paths.

## Initial rollout

The merge base for this introduction PR has no benchmark runner or suite, so the
first workflow run is intentionally candidate-only. It verifies environment
setup, registration selection, correctness, timing, artifact generation, and
the exit contract, but cannot produce a performance comparison.

The pull request introducing the trusted comment workflow also cannot trigger
that publisher because a `workflow_run` receiver must already exist on the
default branch. After this lands, a manual `main` versus `main` run can verify
the runner and artifact path, and the next qualifying same-repository pull
request can exercise the complete paired comparison and comment lifecycle.

The check should initially remain optional. Its path filters mean unrelated pull
requests do not create the check, which is incompatible with an unconditional
required-check rule.


borontion:
can we make this one stage of pr-test-amd if plan to make it part of the ci?

borontion (approved):

tokenspeed-kernel/python/tokenspeed_kernel/benchmark/graph.py
```diff
@@ -0,0 +1,513 @@
...
+            backend.synchronize_stream(stream)
+
+            graph = backend.new_graph()
+            with backend.capture(
+                graph,
+                stream,
+                capture_error_mode="global",
+            ):
+                for _ in range(self.config.calls_per_graph):
+                    prepared.invoke()
```
antiagainst, line 368: This is effectively making all kerrnels reading hot cache, which is not gonna be representative for kernel performance in model runs? It overestimates kernel performance and may miss regressions for cold cache. I think we should rotate input buffers or clear cache between kernel invocations?

You can refer to `do_bench`, `do_bench_cudagraph`, and `do_bench_proton` in https://github.com/triton-lang/triton/blob/main/python/triton/testing.py to see how different flavors look like.

tokenspeed-kernel/README.md
```diff
@@ -123,6 +123,10 @@ iteration.
   kernel against the reference impl.
 - `python -m tokenspeed_kernel.benchmark` — unified timing, throughput
   (FLOPs / bytes) per op family, tabular reports, and Proton integration.
+- `KernelBenchmarkHarness` — registration-level device timing through warmed
+  graph replay, with raw samples, resolved registration metadata, and explicit
+  failure outcomes. The first operation-owned generator covers dense BF16
```
antiagainst, line 128: The last sentence is agents adding "current" state claims, which can be outdated quickly. For such top-level README.md they are more geared towards human reading so we should be concise and put stuff in high level and long lasting in general.

tokenspeed-kernel/benchmarks/amd/gfx950.json
```diff
@@ -0,0 +1,43 @@
...
+          "dtype": "bfloat16",
+          "validation": {
+            "runs": 5,
+            "atol": 0.015,
+            "rtol": 0.015
+          }
+        },
+        "registration": "gluon_bmm_a16w16_gfx950",
+        "seed": 42,
+        "definition_version": 1
```
antiagainst, line 34: Do you see a need for version? This is pretty much for internal use so no stability guarantee. Having this will likely "misguide" agents in the future that they need to handle backward compatiblity and throw in extra complexity that is not needed.

antiagainst (commented):
This is a lot of new code actually. Now we have parallel benchmark mechanisms--the old one more for scanning and comparing various solutions, while this one more for specific solution to track regression across time. We may need to think about whether/how we should consolidate them.

tokenspeed-kernel/benchmarks/amd/gfx950.json
Max191: The point of the version is so that if we make significant changes to the semantics of an operation then we can tell the CI that comparisons aren't meaningful. Otherwise, I think agents might be inclined to create entire new operations to avoid "regressions" when a semantic change of an existing operation would suffice.

I agree that we should be wary of the backwards compatibility agent issue, though. I'll try to figure out a better naming/documentation so that this is clear.

PR #1498, opened 2026-09-11
feat(dcp): Add DeepSeek V4 decode context parallelism

kangguangli:
### 1. Cache Pool
Add shard-aware block allocation and recycling with balanced placement across DCP ranks. Update cache arena sizing to account for sharded cache groups and expose the resulting virtual block capacity.

**Adding DCP support for other models requires careful handling of attention metadata: distinguish virtual logical block IDs from rank-local physical pages, and refresh step-dependent metadata correctly.** 

### 2. DSV4 DCP support
Add decode context parallelism (DCP) for DeepSeek V4 by sharding the compressed KV cache while keeping the SWA cache, compression states, and indexer cache replicated. Further cache optimizations are planned for follow-up PRs.

During prefill, gather the required compressed KV entries across DCP ranks and compute attention over each TP rank’s local query heads. During decode, all-gather query heads, compute attention partials over local cache shards, reweight them using all-gathered log-sum-exp (LSE) values, and reduce-scatter the results back to the corresponding TP ranks.

This initial implementation prioritizes simplicity; further communication optimizations are left to follow-up PRs.

### 0. Device

8xH20

### 1. Accuracy Evaluation
Completed 3 full evaluations on GSM8K and GPQA Diamond, comparing TP and DCP with and without DSpark MTP over three runs at concurrency 64. 

| Dataset        | Mode         | TP Accuracy | DCP Accuracy |
|----------------|--------------|--------------|---------------|
| GSM8K          | Without MTP  | 97.04%       | 96.94%        |
| GSM8K          | DSpark       | 97.14%       | 97.09%        |
| GPQA Diamond   | Without MTP  | 73.23%       | 74.07%        |
| GPQA Diamond   | DSpark       | 74.24%       | 73.40%        |

### 2. Logits Consistency under Fixed-Token Replay
We used 16 fixed prompts and generated one reference continuation for each, then replayed the same token sequences three times each on TP and DCP. Since repeated runs produced different logits even within the same configuration, we compared TP–DCP differences against this repeat-run variation using RMSE, computed as the square root of the mean squared logit difference across all compared positions and vocabulary entries. Within-configuration results include three run pairs; TP–DCP results include all nine cross-configuration pairs.

  Each run covered 1,855 prediction positions: 16 prefill and 1,839 decode positions, with 129,280 vocabulary logits per position. All runs used batch size 1, eager execution, overlap disabled, and prefix caching disabled.

| Phase    | Comparison   | Logits RMSE | Maximum Absolute Error | Argmax Mismatch Rate |
|-----------|-------------|--------------|-------------------------|----------------------|
| Prefill   | Within TP   | 0.594546     | 4.000000                | 0.0000%              |
| Prefill   | Within DCP  | 0.537703     | 4.250000                | 0.0000%              |
| Prefill   | TP vs. DCP  | 0.567106     | 5.000000                | 0.0000%              |
| Decode    | Within TP   | 0.365015     | 20.796875               | 1.3776%              |
| Decode    | Within DCP  | 0.356254     | 17.187500               | 1.4863%              |
| Decode    | TP vs. DCP  | 0.368153     | 24.750000               | 1.5105%              |

### 3. KV Cache Token Capacity

| Configuration      | Original TP Capacity | DCP Capacity | DCP / Original TP |
|--------------------|----------------------|----------------------|----------------------------|
| DCP2 without MTP   | 739,584              | 1,423,360            | 1.925×                     |
| DCP2 with DSpark   | 689,408              | 1,294,336            | 1.877×                     |
| DCP4 without MTP   | 2,514,432            | 8,323,072            | 3.310×                     |
| DCP4 with DSpark   | 2,464,256            | 8,128,512            | 3.299×                     |
| DCP8 without MTP   | 6,063,616            | 27,758,592           | 4.578×                     |
| DCP8 with DSpark   | 6,013,440            | 27,496,448           | 4.572×                     |

### 4. TTFT

| Input Tokens | TP Median TTFT (s) | DCP Median TTFT (s) | DCP / TP |
|---------------|--------------------|----------------------|-----------|
| 2,048         | 0.336369           | 0.337726             | 1.004035  |
| 16,384        | 6.069444           | 6.067989             | 0.999760  |
| 65,536        | 59.409787          | 59.313910            | 0.998386  |
| 131,072       | 211.197984         | 210.806947           | 0.998148  |

### 5. Decode Throughput

Measured on 8 GPUs without MTP, using a 64 GiB cache arena budget per GPU, CUDA graphs, and overlap enabled. Each measurement used 64 concurrent requests sharing the first L−256 input tokens, with a distinct 256-token suffix per request. Each request generated 256 tokens with EOS ignored. Throughput was measured over a shared steady-state decode window; values are medians across repeated measurements.

| Input Tokens | TP Throughput (tokens/s/GPU) | DCP Throughput (tokens/s/GPU) | DCP Relative Change | TP Throughput (tokens/s/user) | DCP Throughput (tokens/s/user) |
|---------------|------------------------------|--------------------------------|----------------------|--------------------------------|---------------------------------|
| 2,048         | 6.500900                     | 6.483922                       | -0.2612%             | 0.812613                       | 0.810490                        |
| 16,384        | 6.496748                     | 6.482765                       | -0.2152%             | 0.812094                       | 0.810346                        |
| 65,536        | 6.489622                     | 6.475325                       | -0.2203%             | 0.811203                       | 0.809416                        |
| 131,072       | 6.481469                     | 6.464987                       | -0.2543%             | 0.810184                       | 0.808123                        |


python/tokenspeed/runtime/distributed/mapping.py
```diff
@@ -139,11 +148,32 @@ def __init__(
         tp_size: int | None = None,
         cp_size: int | None = None,
         dp_size: int | None = None,
+        dcp_size: int = 1,
     ):
         super().__init__(rank, world_size)
         self.tp_size, self.cp_size, self.dp_size = _resolve_parallelism_sizes(
             self.world_size, tp_size, cp_size, dp_size
         )
+        self.dcp_size = _resolve_dcp_size(self.tp_size, self.cp_size, dcp_size)
```
syuoni, line 157: Wondering why we introduce a new `dcp_size` instead of reusing the existing `cp_size`.

I suggest defining the terms consistently:
* TP partitions the layer’s weights across `tp_size` ranks.
* CP partitions attention computation or KV cache across `cp_size` ranks.

A basic DCP partitions the KV context while keeping attention projections replicated. Each rank computes local attention, the results are merged using LSE, and the replicated output projection needs no further reduction.

In this PR, the Attention projections are also sharded in CP ways, so could be interpreted as TP -- but that would cause definition confusion.  I would think of the Attention projection sharding as extensions to the basic CP -- repurposing CP ranks as TP ranks provisionally. And we pay extra comm operations.

I suggest expressing context partitioning (CP) and projection weight partitioning (TP) separately in Mapping. Replicated Attention projections would be the basic CP configuration; sharded Attention projections would be an explicit extension.

python/tokenspeed/runtime/execution/distributed_initializer.py
```diff
@@ -166,6 +166,8 @@ def initialize(config: DistributedConfig) -> float:
         )
         pg_manager.init_process_group(config.mapping.world_group)
         pg_manager.init_process_group(config.mapping.attn.tp_group)
+        if config.mapping.attn.has_dcp:
+            pg_manager.init_process_group(config.mapping.attn.dcp_group)
```
syuoni: I think we can init process group for group of size 1. Namely, the condition can be removed.

python/tokenspeed/runtime/layers/attention/backends/specific/deepseek_v4.py
```diff
@@ -1422,6 +1522,156 @@ def forward_deepseek_v4_mixed(
...
+            seq_cpu,
+            query_cpu,
+            num_reqs=count,
+            window_size=window_size,
+            compress_ratio=compress_ratio,
+        )
+        row_counts = [s // compress_ratio for s in seq_cpu.tolist()]
+        gid = v4_compressed_kv_group_id(compress_ratio)
+        rows_per_page = v4_compressed_rows_per_page(compress_ratio)
+        table = cache.compressed_page_table(compress_ratio)[:count].cpu().tolist()
```
syuoni: This is a D2H copy sync. Could we build this from the existing CPU block-table data?

The subsequent loops also expand the entire compressed history into Python destination lists and upload them again, even for a small extension of a long cached prefix. Could we use tensor operations?

python/tokenspeed/runtime/layers/attention/dcp/comm.py
```diff
@@ -0,0 +1,82 @@
...
+def gather_query_heads(query: torch.Tensor, group: tuple[int, ...]) -> torch.Tensor:
+    """Gather only actual TP query heads after QNorm/RoPE; padding stays local."""
+    if len(group) == 1:
+        return query
+    tokens, heads, dim = query.shape
+    # The 2-D inner-dimension collective uses the existing low-latency
+    # backend where supported, with its topology/dtype/NCCL fallbacks.
+    gathered = all_gather(
+        query.reshape(tokens, heads * dim).contiguous(), group, dim=-1
+    )
```
syuoni, line 49: This allocates a symm buffer of size `(max_tokens, hidden_size)`, where max_tokens is the chunked prefill size.

However, this buffer is only used for decode, where the max_tokens is `max_decode_bs * spec_tokens`. Ideally we need a specific capacity for this all_gather op. While, currently, there is no straightforward way for this. We can revisit this issue, trying to manage symm buffer in a unified way.

syuoni (commented):
Thanks @kangguangli @FC-Li , Overall LGTM for the Python side.

Please take a look at the comments. The major concern is the introduction of `dcp_size` in `Mapping`. We can discuss the design offline.

python/tokenspeed/runtime/distributed/mapping.py
FC-Li: Thanks — agreed the two words need a precise definition, and I think that definition is what makes `dcp_size` a separate field for now rather than a reuse of `cp_size`.

What `cp_size` means today (`ENABLE_CP`): the sequence is split zigzag across CP ranks during prefill, attention projections are *replicated* (`attn_tp_size` is reinterpreted as `cp_size` and TP set to 1), `cp_group` has stride `tp_size`, and CP is a world-size dimension. Several consumers rely on "CP peers hold the same weights": `kimi_k3`, `deepseek_v41` and the DSpark drafter refuse `cp_size != 1`, and the DP-rank arithmetic in `comm_manager` / `base/comm_ops` multiplies `tp_size * cp_size`.

What this PR shards is a *subgroup of attention TP* (stride 1, no new world dimension). Its ranks hold different projection shards, so every consumer of `attn.tp_size` for weight partitioning stays correct; only the compressed-KV placement and the decode attention (Q all-gather → local partial → LSE merge → reduce-scatter back to the head owner) see the subgroup. Folding that into `cp_size` today would either trip the `cp_size == 1` guards above or require redefining the weight-sharding group as the TP×CP product, which reaches into MoE dispatch and the dense all-reduce paths.

So I would keep `dcp_size` as "the TP subgroup whose KV is sharded" in this PR, and treat the model you describe — CP partitions context, TP partitions weights, sharded projections as an explicit extension of basic CP — as a Mapping redesign in its own PR, where `dcp_size` becomes derived rather than declared. Happy to discuss offline; if you'd rather land the redesign first, this can be rebased on top of it.

python/tokenspeed/runtime/execution/distributed_initializer.py
FC-Li: Done in 3ad5e08a: the DCP group is initialized unconditionally (`init_process_group` is idempotent and a group of one is still the group the decode collectives address). That was also the last `has_dcp` branch on the runtime path.

python/tokenspeed/runtime/layers/attention/backends/specific/deepseek_v4.py
FC-Li: Done in 3ad5e08a. The scheduler tables are already staged through a pinned host buffer before their single H2D copy, so `packed_block_tables_from_forward_op` now returns those host views alongside the device ones; `CacheBatchMetadata.tables_cpu()` exposes them and the runner passes `block_tables_cpu` into `init_forward_metadata` next to `block_tables`, the same way the extend lengths travel with their `*_cpu` mirrors.

`_build_dcp_prefill_chunks` is now tensor arithmetic over that host table — owner per row, a stable argsort by owner that preserves request/row order, `bincount` for the per-owner counts, and one upload of the destinations per chunk — with no device sync and no per-row Python. The remaining per-chunk host work is `_prefill_workspace_bounds`, which the non-DCP path already does.

python/tokenspeed/runtime/layers/attention/dcp/comm.py
FC-Li: Agreed. I left a note at the call site; the decode gather only needs `max_decode_bs * spec_tokens` rows, and giving it its own capacity wants the symmetric buffers managed in one place first, so that stays for the follow-up you describe.

PR #1500, opened 2026-09-11
perf(kda): fuse prefill preparation and preserve native state layout

byshiue:
## Summary

This PR reduces KDA prefill overhead through shared metadata, fused state preparation, and native recurrent-state layouts.

### Shared metadata and fused state preparation

- Build convolution request/chunk indices once per forward with a Triton kernel and reuse them across layers.
- Share one int64 sequence-boundary tensor across KDA scans, replacing repeated per-layer casts.
- Combine history-flag generation, convolution-state copying, and recurrent-state gathering/zeroing into one Triton launch.
- Give each forward its own metadata storage so overlapping forwards can safely prepare and consume different batches.

These changes sit in the shared Mamba/GDN preparation path used by KDA. Resumed requests load their cached state, while fresh requests receive zero-initialized recurrent inputs.

### Native CuTeDSL state layout

Keep recurrent states in the `[N, H, V, K]` layout used by the NVIDIA cache and CuTeDSL kernel. Aligning the backend’s layout declaration and adapter removes the redundant V-major → K-major → V-major conversion and its associated copies.

## Performance

Compared upstream main `d3402a44` with this PR at `3ec83093`.

| Metric | Main | This PR | Latency reduction |
|---|---:|---:|---:|
| Median server prefill | 270.880 ms | 201.840 ms | **25.49%** |
| Median client E2E | 286.314 ms | 218.231 ms | **23.78%** |
| Prefill p95 | 277.660 ms | 205.530 ms | — |

The median prefill improvement is **69.040 ms**, or **1.342×**.

### Test environment

- Eight NVIDIA GB300 GPUs in one NVL72 domain.
- Full 93-layer Kimi-K3 with real NVFP4 weights.
- Attention and MoE TP8, BF16 compute, and FP8 KV cache.
- Prefill/decode CUDA graphs and overlap scheduling enabled, with eager KDA attention breaks.
- PyTorch `2.13.0+cu130`, FlashInfer `0.6.18`, and DeepGEMM 2.8.
- Unprofiled measurements with identical NVTX and request logging settings.

### Workload and measurement

The dataset contains 30 fixed synthetic continuation prompts, each appending 800 tokens to the same 50,500-token parent. Each measured request has:

- **51,300 input tokens**
- **50,432 cached tokens**
- **868 new tokens**, scheduled as **768 + 100**
- **One greedy output token**

Each version ran in two fresh server processes, ordered main → PR → PR → main. Each process measured three rounds of 30 requests, with cache reset, prefix priming, and warmup before each round: **180 measured requests per version**.

Timing covers steady-state cached incremental prefill after model loading, graph capture, and warmup. Server prefill latency includes CPU scheduling and launch gaps alongside GPU execution. The reported statistics retain every measured sample.

## Correctness validation

Validation used the CPU and GB300 environment above. Coverage combines synthetic tensor tests with full-model output comparisons on the fixed continuation dataset.

### Kernel and runtime tests

**284 tests and 56 subtests passed**, including existing regression coverage. There were 29 platform/dependency-specific skips: 23 AMD cases and 6 cases requiring optional FLA support.

The tests cover:

- Convolution indices, sequence boundaries, and metadata ownership across forwards.
- Fresh and resumed requests, shared state pages, non-contiguous tensors, and stale/NaN state handling.
- Bitwise output and final-state parity against the previous preparation sequence.
- Native state-layout handling and CUDA graph replay with updated inputs and metadata.

The full `pre-commit run --all-files` check passed.

### Full-model output equivalence

- **180/180 measured PR outputs matched main** for identical prompts.
- All four runs produced the same 500-token parent continuation from the 50,000-token synthetic prompt.
- Every measured request matched the expected cache-hit count and 768+100 scheduling.
- Graph configurations, native binary hashes, tuning configurations, and MoE fallback selections matched across runs.

These results establish output equivalence for the tested synthetic workloads and correctness coverage for the changed preparation and layout paths.


syuoni (approved):
LGTM, thanks! @byshiue

PR #1502, opened 2026-09-11
perf: materialize state checkpoints in a single prefill forward

byshiue:
## Summary

This PR lets a stateful prefill produce both an aligned checkpoint for prefix reuse and the final state for continued generation in one model forward.

Previously, an off-boundary prompt ending could require two model forwards just to preserve the aligned checkpoint. With 128-token prefix granularity, an 868-token extension would run as 768 tokens followed by 100 tokens.

The scheduler now submits all 868 tokens together when the token budget and prefix-promotion constraints allow it. Ordinary chunked-prefill limits still apply.

### What changed

- **Single-forward checkpointing.** The state backend writes the aligned convolution and recurrent checkpoints while preserving the final continuation state.
- **No repeated recurrent prefix scan.** Recurrent attention scans the 768-token body, saves its checkpoint, and continues with the remaining 100 tokens from that state. Convolution processes all 868 tokens once. This is one model forward, not one kernel launch.
- **Batched checkpoint operations.** Triton kernels handle convolution checkpoint writes and recurrent input packing/state writes across requests, replacing per-request Python loops. Single-request recurrent inputs retain a zero-copy view path.
- **Shared metadata.** Checkpoint metadata is prepared once per forward and reused across layers, including KDA's int64 sequence boundaries.
- **Cache lifecycle and capacity handling.** Admission accounts for the input state, checkpoint, continuation state, and decode growth. Prefix publication tracks the boundary actually materialized rather than assuming every crossed boundary has a reusable state.

The implementation uses the shared GDN/KDA state backend and retains the existing KDA scan adapters. It adds no prefill-mode switch and includes no FP8 quantization changes.

## Performance

The controlled recurrent-scan comparison used:

- Full 93-layer Kimi-K3 with real NVFP4 weights.
- Eight NVIDIA GB300 GPUs across two nodes, TP8, FP8 KV cache.
- Concurrency 1, prefix granularity 128, piecewise prefill and decode CUDA graphs enabled.
- Identical 51,300-token inputs with 50,432 cached tokens, an 868-token extension, and one generated token.
- A–B–B–A run order on the same allocation, with 10 warmups and 50 timing samples per run: 100 measured samples per variant.

Both variants already used one model forward. The comparison isolates replacing the earlier `768 + 868` recurrent rescan with `768 + 100` continuation.

| Metric | Repeated prefix scan | Body + tail continuation | Reduction |
| --- | ---: | ---: | ---: |
| Median server prefill latency | 181.035 ms | 176.870 ms | 2.30% |
| Median client end-to-end latency | 196.519 ms | 192.383 ms | 2.10% |
| Median profiled KDA kernel-time sum per GPU/request | 5.672 ms | 4.090 ms | 27.88% |

Server and client timings were collected with NSYS capture disabled; NVTX and the idle NSYS launcher were present equally in both arms. Kernel timings came from separate captures and are not end-to-end critical-path latency.

Twenty-request timelines verified one 868-token prefill forward per request, CUDA graph replay, and removal of the repeated 868-token recurrent scan.

These are development measurements, not a fresh comparison of the final rebased branch against the latest upstream main.

### GPU and real-model regression checks

On the full real-weight Kimi-K3 NVFP4 TP8 configuration above:

- The controlled scan comparison matched output tokens across all four runs for all 70 checked requests per run.
- Subsequent review-fix validation passed 22 checkpoint-kernel tests and 41 GDN tests, plus five subtests.
- Tests cover checkpoint continuation, batched and strided inputs, convolution-state aliasing, and CUDA graph execution.
- Scheduler tests cover checkpoint provenance, accepted-token boundaries, and capacity-limited admission.

These checks do not establish full-model accuracy for every GDN/KDA model or speculative-decoding configuration.

### AIME 2026

An earlier single-forward implementation was evaluated with the full real 93-layer Kimi-K3 NVFP4 model, TP8 on eight GB300 GPUs, FP8 KV cache, overlap, and prefill/decode CUDA graphs enabled.

Dataset: `math-ai/aime26`, all 30 problems. Evaluation used EvalScope 1.11.1, a 65,536-token context, and a 63,488-token output limit.

| Evaluation | Temperature | Correct | Accuracy |
| --- | ---: | ---: | ---: |
| Standard EvalScope, one sample per problem | 1.0 | 27/30 | 90.0% |
| Cold-cache paired check | 0 | 29/30 | 96.7% |
| Warm-cache paired check | 0 | 29/30 | 96.7% |

All 30 warm requests reused the expected aligned checkpoint. However, cold/warm final answers agreed on only 28/30 problems, and complete output-token sequences were not identical. One warm request exhausted its output budget.

Equal aggregate scores therefore do not establish cache-path equivalence or prove no accuracy regression. Follow-up diagnostics identified shape-sensitive numerical differences in the quantization, MoE, and MLA paths exposed by the scheduling change; this PR does not include repairs for those numerical differences.

### Validation after rebasing

- 459 C++ tests passed.
- 160 Python scheduler/runtime tests passed.
- Two attention-layout tests passed.
- 42 CUDA/Triton-dependent tests were skipped on the validation host.

Full-model GPU performance and AIME evaluation have not been rerun on the final rebased commit.


syuoni:
Hi @FC-Li @wangbo981016 , could you please review this PR? Thanks!

python/tokenspeed/runtime/layers/attention/backends/state/mamba.py
```diff
@@ -308,9 +364,136 @@ def _build_cu_extend_seq_lens_cpu(
...
+        rows,
+        starts,
+        lengths,
+        positions,
+        body_rows,
+        body_indices,
+        tail_indices,
+        body_query_start_loc,
+        tail_query_start_loc,
+    ) = (view.view(part.dtype) for view, part in zip(uploaded.split(byte_sizes), parts))
```
syuoni: This is a nice optimization pattern to move many small buffers from cpu to gpu. Can we make a reusable helper function for it?

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/_triton/prefill_state_checkpoints.py
```diff
@@ -0,0 +1,634 @@
...
+        + channel * state_channel_stride
+        + (state_len + relative_token) * state_position_stride
+    )
+    # Load the complete old-state window before any store. This makes the
+    # operation safe even if a caller aliases source and destination pages.
+    raw_value = tl.load(raw_inputs + raw_offsets, mask=position_live & from_raw)
+    source_value = tl.load(conv_states + source_offsets, mask=position_live & ~from_raw)
+    value = tl.where(from_raw, raw_value, source_value)
+    destination_offsets = (
+        destination * state_page_stride
```
syuoni, line 99: Please cast the loaded page IDs to tl.int64 before multiplying by the pool strides, in both checkpoint writers. These IDs are int32, so the current multiplication can overflow for large cache pools.

For example, the Kimi TP8 convolution layout has a page stride of 442,368 BF16 elements. Page 4,855 produces an offset of 2,147,696,640, which overflows signed int32 and becomes negative.

The fix should cover both source and destination offsets in the convolution writer, plus the destination offset in _write_prefill_recurrent_checkpoints_kernel. The existing causal-convolution kernel already widens its page index before this multiplication.

syuoni (dismissed):
Thanks @byshiue . LGTM for the Python side.

Let's wait for the reviews for the C++ side.

FC-Li:
LGTM. Please resolve the conflicts, then the pr is good to be merged.

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/_triton/prefill_state_checkpoints.py
FC-Li: @syuoni We also encountered the int32 overflow in triton kernels before @OftenDream. This kind of bugs are very hard to detect. Do you think it's a good practice to construct the block ids in int64 and keep them in that precision all the way to where they are used. Maybe we can do this in another PR.

borontion: cc @raikonenfnu @Yu-Zhewen this is common issue we met before for other kernels as well. let's make sure we also tested it

PR #1505, opened 2026-09-11
perf(amd): schedule causal prefill workgroups longest-first

nithinsubbiah:
## Summary

Schedules sufficiently occupied gfx1250 full-causal MHA prefill workgroups in descending q-block order.

Causal work per q block grows monotonically: on the 256x64 tile, the first block visits 4 KV tiles while the last block of a 4096-token sequence visits 64. Ascending order leaves expensive workgroups in the dispatch tail. Mapping early physical IDs to late logical q blocks lets the hardware schedule long work first and backfill CUs with shorter workgroups.

This changes only workgroup order. Per-workgroup attention math, memory accesses, K/V buffering, and output locations are unchanged.

The policy requires:

- full causal attention;
- at least 256 live workgroups, computed from the packed sequence lengths;
- more than one q tile.

Sliding-window, single-tile, and underfilled launches retain ascending order.

## Performance

gfx1250 at 2400 MHz. Every row passed the causal reference before timing and received a CLEAN exclusive-run verdict.

Five order-balanced trials on BF16, 4x4096, GQA 8/1, D128, 512 workgroups:

| order | mean TFLOP/s |
|---|---:|
| ascending | 1,357.0 |
| descending | 1,670.2 |

Gain: **+23.08%**.

Additional screens:

| case | ascending | descending | change |
|---|---:|---:|---:|
| S8192, 2 workgroups/CU | 1,472 | 1,852 | +25.8% |
| 3 workgroups/CU | 1,471 | 1,628 | +10.7% |
| H32/Hkv8, 8 workgroups/CU | 1,622 | 1,659 | +2.3% |
| 1 workgroup/CU | 1,045 | 1,052 | +0.7% |
| ragged 851/914/1053 | 620 | 675 | +8.9% |
| D64 | 878 | 1,076 | +22.6% |
| FP16 | 1,334 | 1,630 | +22.2% |
| FP8 E4M3 | 2,014 | 2,541 | +26.2% |
| FP8 E5M2 | 2,012 | 2,541 | +26.3% |
| sinks + LSE | 1,357 | 1,667 | +22.8% |

Measured exclusions: window 1024 was neutral, window 64 regressed 2.2%, and a 128-workgroup underfill regressed 3.7%.

Codegen remains spill-free at 467 VGPRs with the same 256 WMMAs, waits, barriers, and memory instructions.

## Test Plan

- gfx1250 direct hardware matrix: 15/15 passed
- narrow and wide tiles, D64/D128, full/sliding attention
- bit-identical ascending/descending outputs on ragged remainder tiles
- adversarial ragged occupancy gate: 512 rectangular groups / 47 live groups
- BF16, FP16, FP8 E4M3/E5M2, GQA/MHA, sinks, and LSE performance correctness gates

### DIFF
--- a/tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx1250/attention/mha/prefill.py
+++ b/tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx1250/attention/mha/prefill.py
@@ -61,6 +61,7 @@ class AttentionConfig:
     HAS_LSE: gl.constexpr
     WINDOW_LEFT: gl.constexpr
     TDM_WARP_HINT: gl.constexpr
+    REVERSE_Q_BLOCKS: gl.constexpr
     q_strides: InputStrides
     k_strides: InputStrides
     v_strides: InputStrides
@@ -90,6 +91,7 @@ class AttentionConfig:
         HAS_LSE,
         WINDOW_LEFT,
         TDM_WARP_HINT,
+        REVERSE_Q_BLOCKS,
         q_strides,
         k_strides,
         v_strides,
@@ -136,6 +138,7 @@ class AttentionConfig:
         self.HAS_LSE = gl.constexpr(HAS_LSE)
         self.WINDOW_LEFT = gl.constexpr(WINDOW_LEFT)
         self.TDM_WARP_HINT = gl.constexpr(TDM_WARP_HINT)
+        self.REVERSE_Q_BLOCKS = gl.constexpr(REVERSE_Q_BLOCKS)
         self.q_strides = q_strides
         self.k_strides = k_strides
         self.v_strides = v_strides
@@ -232,6 +235,8 @@ class AttentionProgram:
         batch = gl.program_id(0)
         q_head = gl.program_id(1)
         q_block = gl.program_id(2)
+        if cfg.REVERSE_Q_BLOCKS:
+            q_block = gl.num_programs(axis=2) - 1 - q_block
         kv_head = q_head // (cfg.N_HEADS // cfg.N_KV_HEADS)
         seq_base = gl.load(cu_seqlens_ptr + batch)
         seq_end = gl.load(cu_seqlens_ptr + batch + 1)
@@ -658,6 +663,7 @@ def _mha_prefill_gfx1250(
     HAS_LSE: gl.constexpr,
     WINDOW_LEFT: gl.constexpr,
     TDM_WARP_HINT: gl.constexpr,
+    REVERSE_Q_BLOCKS: gl.constexpr,
     NUM_WARPS: gl.constexpr,
     NUM_BUFFERS: gl.constexpr,
 ):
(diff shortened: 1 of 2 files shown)


borontion (dismissed):

tokenspeed-kernel/test/ops/attention/test_gluon_mha_prefill_gfx1250.py
borontion, line 1: a separated issue for @lightseekorg/kernel-team: I think we need to consider to re-organized these kernel tests - now changing tokenspeed-kernel-amd requires updating tests in tokenspeed-kernel - this will trigger all nvidia/amd ci to run which waste lots of resources.

antiagainst: Yeah I moved the tests to be under tokenspeed-kernel to share common utility and reference torch/triton computations in https://github.com/lightseekorg/tokenspeed/pull/1004. Otherwise we need to grow another set of such stuff in the -amd directory to avoid depdendency cycles.

I guess we can formalize the test/ directory structure a bit to have sub nvidia/ and amd/ directory to help CI filtering stuff?

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx1250/attention/mha/prefill.py
```diff
@@ -876,6 +894,12 @@ def gluon_mha_prefill_gfx1250(
         window_left=config.window_left,
         workgroups=math.prod(config.grid),
     )
+    reverse_q_blocks = _select_reverse_q_blocks(
+        block_m=config.block_m,
+        max_seqlen=config.max_seqlen,
+        window_left=config.window_left,
+        workgroups=math.prod(config.grid),
```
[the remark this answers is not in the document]
nithinsubbiah: Fixed in [e1a8058](https://github.com/lightseekorg/tokenspeed/commit/e1a8058). The reverse-order gate now counts live workgroups as `n_heads * sum(cdiv(seq_len, block_m))` from `cu_seqlens_cpu` instead of using the rectangular grid. The reported 1x4096 + 31x1 case now evaluates to 47 live groups (versus 512 rectangular) and keeps ascending order. Added that exact regression test; all 15 direct gfx1250 hardware tests and `pre-commit run --all-files` pass.

tokenspeed-kernel/test/ops/attention/test_gluon_mha_prefill_gfx1250.py
nithinsubbiah: Sure, makes sense this should be handled separately as a test-layout change. I will keep this performance PR focused and leave the gfx1250 test in the existing shared location.

borontion (approved):

tokenspeed-kernel/test/ops/attention/test_gluon_mha_prefill_gfx1250.py
antiagainst: Put up https://github.com/lightseekorg/tokenspeed/pull/1522 to split the directories.

PR #1508, opened 2026-09-11
perf(kimi3): add gfx1250 small-M WMMA and row-CTA decode GEMVs

Yu-Zhewen:
## Summary
- Register the existing gfx1250 dense16 WMMA kernel for `decode_gemv` at M 2..32, and let M == 1 route to the row-CTA GEMV, so K3 decode projections stop falling through to rocBLAS on CDNA5.
- Unpin `K` on `_wmma_tdm_dense_m16_kernel`. It was fixed at 7168, which excluded the shapes that dominate a decode step: o_proj at K=1536 and latent-up at K=3584.
- Add `gluon_wmma_tdm_dense_gfx1250`, a shape-generic wrapper beside the two shape-locked ones, validating the kernel's actual constraints (M <= 32, `K % 128 == 0`, `N % 16 == 0`) instead of a shape allowlist.
- CDNA5 only, via the spec's own capability gate. CDNA4 has its own tuned M == 1 Triton GEMVs and has not been swept against these.

## Performance
Kernel latency vs. main on gfx1250, cold L2 and CUDA-graph captured. `PR` is whichever kernel the registry selects: row-CTA at M == 1, dense16 WMMA at M 2..32.

o_proj 1536→7168 (93 calls/forward):
| M | Main (µs) | PR (µs) | Speedup vs. main |
|--:|----------:|--------:|-----------------:|
| 1 | 14.15 | 4.01 | 3.53x |
| 2 | 14.14 | 6.26 | 2.26x |
| 4 | 14.40 | 6.40 | 2.25x |
| 8 | 14.56 | 6.21 | 2.34x |
| 16 | 14.63 | 6.32 | 2.32x |
| 24 | 14.68 | 10.46 | 1.40x |
| 32 | 14.87 | 10.24 | 1.45x |

latent-up 3584→7168:
| M | Main (µs) | PR (µs) | Speedup vs. main |
|--:|----------:|--------:|-----------------:|
| 1 | 29.97 | 6.76 | 4.43x |
| 2 | 30.15 | 13.01 | 2.32x |
| 4 | 30.08 | 12.46 | 2.42x |
| 8 | 30.44 | 12.80 | 2.38x |
| 16 | 30.15 | 12.99 | 2.32x |
| 24 | 30.18 | 18.46 | 1.64x |
| 32 | 30.15 | 19.23 | 1.57x |

dense-mlp-down 4224→7168 (1 call/forward):
| M | Main (µs) | PR (µs) | Speedup vs. main |
|--:|----------:|--------:|-----------------:|
| 1 | 34.58 | 7.41 | 4.67x |
| 2 | 34.81 | 14.48 | 2.40x |
| 4 | 34.87 | 14.15 | 2.46x |
| 8 | 35.07 | 14.19 | 2.47x |
| 16 | 35.15 | 14.17 | 2.48x |
| 24 | 35.30 | 21.80 | 1.62x |
| 32 | 35.36 | 22.09 | 1.60x |

shared-down 768→7168:
| M | Main (µs) | PR (µs) | Speedup vs. main |
|--:|----------:|--------:|-----------------:|
| 1 | 8.24 | 3.50 | 2.35x |
| 2 | 8.03 | 4.07 | 1.97x |
| 4 | 8.37 | 4.15 | 2.02x |
| 8 | 8.62 | 4.25 | 2.03x |
| 16 | 8.77 | 4.26 | 2.06x |
| 24 | 8.92 | 6.19 | 1.44x |
| 32 | 8.92 | 7.13 | 1.25x |

M >= 48 is deliberately left unregistered and falls back to rocBLAS. The launcher tiles M in 16-row chunks and re-reads the whole weight per chunk, so the advantage decays at every chunk boundary: measured on o_proj, M=48 is 1.03x and M=64 is 0.80x.

Model throughput on the K3 toy model at TP8, 4K prompt / 1K output, against a matched control measured on the same tree in the same session. At concurrency 1 decode throughput goes from 95.6 to 108.0 tok/s (+13.0%), with decode step p50 falling from 10.47 to 9.28 ms. At concurrency 16 it goes from 735.4 to 758.5 tok/s (+3.1%), with step p50 falling from 20.80 to 20.23 ms. Concurrency 1 gains more because row-CTA at M == 1 is the larger per-call win and o_proj runs 93 times per forward; at concurrency 16 every decode GEMM is M == 16, which lands on the WMMA kernel at ~2.3x.

### DIFF
--- a/tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/triton_gemv.py
+++ b/tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/triton_gemv.py
@@ -35,6 +35,7 @@ import functools
 
 import torch
 from tokenspeed_kernel._triton import tl, triton
+from tokenspeed_kernel.platform import ArchVersion, CapabilityRequirement
 from tokenspeed_kernel.registry import Priority, register_kernel
 from tokenspeed_kernel.signature import dense_tensor_format, format_signature
 
@@ -140,6 +141,44 @@ def rowcta_gemv(
         num_warps=4,
     )
     return out
+
+
+@register_kernel(
+    "gemm",
+    "decode_gemv",
+    name="gluon_wmma_dense_gfx1250",
+    solution="gluon",
+    capability=CapabilityRequirement(
+        min_arch_version=ArchVersion(12, 5),
+        max_arch_version=ArchVersion(12, 5),
+        vendors=frozenset({"amd"}),
+    ),
+    signatures=_BF16_SIG,
+    traits={
+        "m": frozenset(range(2, 33)),
+        "k_align_128": frozenset({True}),
+        "n_align_16": frozenset({True}),
+    },
+    priority=Priority.SPECIALIZED,
+)
+def wmma_dense_gemv(
+    x: torch.Tensor, weight: torch.Tensor, out: torch.Tensor | None = None
+) -> torch.Tensor:
+    """``x @ weight.T`` for small-M decode activations on CDNA5.
+
+    Args:
+        x: ``[M, K]`` contiguous bf16 activation, K a multiple of 128.
+        weight: ``[N, K]`` contiguous bf16 weight, N a multiple of 16.
+        out: optional ``[M, N]`` destination.
+
+    Returns:
+        ``[M, N]`` output in ``x``'s dtype.
+    """
+    from tokenspeed_kernel_amd.ops.gfx1250.gemm.fp16.mm import (
+        gluon_wmma_tdm_dense_gfx1250,
+    )
+
+    return gluon_wmma_tdm_dense_gfx1250(x, weight, out=out)
 
 
 @register_kernel(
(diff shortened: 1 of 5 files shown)


tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/triton_gemv.py
```diff
@@ -142,6 +143,44 @@ def rowcta_gemv(
     return out
 
 
+@register_kernel(
```
borontion, line 146: is this for moe routing? i was thinking of adding some special moe routing api under moe family.

borontion (approved):

tokenspeed-kernel/python/tokenspeed_kernel/ops/gemm/triton_gemv.py
Yu-Zhewen: It's used in a few different places, MoE, KDA o_proj, and layer 0's dense MLP (gate_up + down).

borontion: > It's used in a few different places, MoE, KDA o_proj, and layer 0's dense MLP (gate_up + down).

yeah I was thinking of creating more specialized apis for different use cases (moe routing, proj, and mlp) given these shapes and total contributions could be different. but that's a separated issue.

PR #1509, opened 2026-09-12
feat(memory): reserve the CUDA-graph pools in the KV cache budget

nperrin-fr:
The KV cache took every byte the profile said was free, and the decode and prefill graphs were captured afterwards out of what the cache left behind. On a model whose capture is large that is a boot-time OOM, and on every model it is a budget nobody measured.

Startup now measures the graphs before it spends the budget. The backends bind a probe-sized arena, capture the largest few entries of each ladder with a driver-memory observer on each capture, and project the rest of the ladder from the marginal cost; the projection is reduced across ranks so every rank sizes its cache for the hungriest one. The executor then rebinds the real pool -- sized with the reserve deducted -- and captures the full ladder.

The rebind is the sequence the two preceding changes opened: the graph owners release what they captured, the workspace pool unfreezes, the backend trees take the replacement pool through set_cache_pool (which refuses a different geometry), the executor republishes its own pool references, and fresh owners are built for the new arena. The probe keeps the backends it measured with -- create_attn_components can build pools for backends it is handed -- so what serves is what was measured.

--disable-cudagraph-memory-reserve skips the probe and restores the old budget. So do the boots that cannot use one: --enforce-eager captures nothing, --enable-memory-saver would leave the probe's arena in a paused region the real pool no longer owns, and speculative decoding stages its verify scratch in the bound pool at the serving concurrency, which a probe arena can only satisfy by growing to serving size -- measured at 31 GiB for Kimi-K3 at --max-num-seqs 256, for four captures it then throws away.

### DIFF
--- a/docs/configuration/server.md
+++ b/docs/configuration/server.md
@@ -107,6 +107,7 @@ distributed update mode until that implementation is added.
 | `--enable-prefix-caching` / `--no-enable-prefix-caching` | Enable or disable prefix cache reuse. |
 | `--enforce-eager` | Disable device-graph execution (CUDA Graph on CUDA, ACL Graph on NPU). |
 | `--disable-prefill-graph` | Keep prefill eager while leaving decode device graphs enabled. |
+| `--disable-cudagraph-memory-reserve` | Size the KV cache from free memory instead of reserving what the device graphs will cost. |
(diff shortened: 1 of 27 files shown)


nperrin-fr:
still WIP, but I'll kick off a CI run later, please do not merge for now

PR #1520, opened 2026-09-12
perf(cache): index per-group LCM placement in BlockPool instead of scanning it

syuoni:
High cache utilization made scheduler admission and per-step page metrics scale with the number of LCM parents. This host overhead limited DeepSeek-V4-Flash decode throughput, particularly with more cache groups and smaller blocks.

`BlockPool` owns indexed occupancy; `PrefixCacheIndex` remains responsible for reclaimability. This removes hot-path scans without teaching the allocator who owns a cache entry. The runtime now consumes the new LCM gauges in this PR. Cached parents exclude active parents, while load snapshots report their sum as resident occupancy, including evictable cache.

The scheduler source version and runtime minimum advance together to **0.1.15**, preventing an older wheel from satisfying a runtime that requires the renamed API. Publish that scheduler release before installing the runtime solely from published dependencies; CI and source development install the scheduler from this tree first.

The benchmark below was measured with the first commit of this PR:

image

Validation: 464 scheduler C++ tests and 153 Python tests pass in CI, including the shutdown-leak regression; AMD runtime 1-GPU and NVIDIA runtime 2-GPU CI also pass. CI actually executes and passes both native runtime gauge cases and all 36 load-snapshot cases, covering packed/unpacked occupancy and load publication. [GB200 Slurm EPD validation](https://github.com/lightseekorg/tokenspeed/actions/runs/34740182282) passes on the fix commit: job 14479, 4 GPUs, all three image prompts correct, no scheduler exception or nanobind leak diagnostics.


syuoni:
@wangbo981016  As discussed offline, this PR removes per-iteration LcmBlock traversal.

@FC-Li @wangbo981016 Could you please help review this PR? Thanks!

FC-Li:
@syuoni thanks for this — the placement index is exactly what admission needed, and I wanted to land it quickly, so I pushed a second commit (5a97f95e) on top of yours (maintainer edits are enabled on this PR; your commit is untouched). Here is what the second commit changes and why.

**Kept: everything about `GroupPlacement`.** Fixed slots-per-parent at pool construction, the per-group free-slot count and ordered partial-parent set, the lazy shadow occupancy in `AdmissionPlanner`, and dropping `cache_blocks_per_lcm_block` from the `Acquire*` calls. This is the part that removes the admission and `cache_group_available_pages()` scans.

**Removed: the reclaimability propagation** — the `cache_owned_` bit in `CacheBlockControl`, `friend class PrefixCacheIndex` in `cache/core`, the `CacheEntry` constructor/destructor/move-assignment, the `reclaimable` / `non_reclaimable_count` fields on `LcmBlock`, and the `SetReclaimable` call from `retain()` / `release()`. Reasons:

- Its only production consumer was `available_kv_pages()`, which the engine uses for the `num_cached_pages` log gauge (the load-snapshot consumer discards that field). Admission and retraction never read it, so the cost/benefit did not work out for a gauge.
- It inverts the layering described in `docs/design/cache-concepts.md`: the refcount control block in `cache/core` gained a bit only `cache/prefix` can interpret, and a `CacheEntry`'s lifetime carried side effects into the pool's counters.
- It created a second definition of "every child is held only by the cache" next to `PrefixCacheIndex::ParentIsFullyEvictable`, which the host eviction path still uses.

**Instead:** the per-step gauge becomes `total - empty - active`. `EmptyLcmBlocks` is O(1) from the pool; `ActiveLcmBlocks` dedupes the live block tables with a bitmap instead of an `unordered_set`. `AvailableLcmBlocks` (the scan) stays as the leak check for tests and diagnostics. Scheduler accessors are renamed to LCM-block vocabulary — `AvailableLcmBlocks` (merging `PoolFreeBlocks` and `AvailableKvPages`), `EmptyLcmBlocks`, `ActiveLcmBlocks` — and the bindings expose only those names (no `*_kv_pages` aliases). The runtime's `_get_scheduler_stats` moves to `empty_lcm_blocks()` + `active_lcm_blocks()` in the same PR that bumps its scheduler pin to the release containing this change.

Also: every `group_placements_[group_id]` access now goes through a bounds-checked accessor. Group ids used to be arbitrary tags; with the vector they would otherwise be UB when out of range.

Tests: 464 C++ unit tests (the two that compared the new counter against `ParentIsFullyEvictable` became self-comparisons and were replaced by one empty-vs-available test, plus new tests for the bounds check and the empty/active partition), 149 Python scheduler tests, `pre-commit run --all-files` clean.

I have rewritten the PR title and description to match the current state. One caveat worth flagging on your measurement: the engine at its current pin still calls `available_kv_pages()` per step, so the green curve is fully reproduced only once the runtime picks up this release together with the call-site switch. If anything here conflicts with what you observed, please say so.

PR #1531, opened 2026-09-13
feat(amd): enable DeepSeek V4 on gfx1250 with portable Triton

antiagainst:
Register portable MXFP4 indexer prefill/decode kernels, graph-stable plans, and sqrt-softplus expert routing for gfx1250 while preserving the optimized gfx950 implementations. Reuse existing Triton attention, cache, and MoE kernels for functional enablement.

Organize the DSv4 Triton solution as a package, preserving its import path and registration discovery. Add helper docstrings, numerical and graph-replay coverage, and centralized kernel-selection tests. Document the solution-package layout and the existing MI350 serving recipe.

End-to-end validation on one MI450 with the official DeepSeek-V4-Flash checkpoint passed all 10 smoke checks in eager and graph modes, including concurrent requests, 2,530-token retrieval, and repeated-prefix caching. A fixed 20-question GSM8K sanity subset scored 19/20 eager and 20/20 with graphs; these are not full benchmark or bitwise-equivalence results.

Working single-MI450 recipe (gfx1250, TP1, no MTP):

```shell
tokenspeed serve deepseek-ai/DeepSeek-V4-Flash \
  --served-model-name deepseek-v4-flash \
  --trust-remote-code \
  --tensor-parallel-size 1 \
  --kv-cache-dtype fp8_e4m3 \
  --moe-backend triton \
  --attention-use-fp4-indexer-cache \
  --max-model-len 4096 \
  --max-total-tokens 8192 \
  --max-num-seqs 4 \
  --chunked-prefill-size 256 \
  --gpu-memory-utilization 0.8 \
  --disable-kvstore \
  --max-cudagraph-capture-size 4 \
  --cudagraph-capture-sizes 1 2 3 4 \
  --prefill-graph-max-tokens 256 \
  --prefill-graph-capture-sizes 128 256 \
  --host 127.0.0.1 \
  --port 8000
```

MTP is not yet validated by this recipe.


borontion (dismissed):

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/dsv4/triton.py
```diff
@@ -22,13 +22,23 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+"""Triton DSV4 kernel exports.
+
+TODO: Moe kernel implementations into their own dedicated files.
```
borontion: what does this TODO mean?

tokenspeed-kernel/README.md
```diff
@@ -182,6 +182,10 @@ value out of TokenSpeed-kernel; but one can also directly call into a
 specific solution under `ops/<family>/`, or manually `select_kernel` with
 targeted filters.
 
+DSV4 prefill and decode selection includes the query head count. FlashMLA's
```
antiagainst: This is documentation not suitable to be placed in top-level `README.md` which is only very high level kernel design. If we need this documentation, better to place it somewhere in `ops/attention` subdirectory.

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/dsv4/triton.py
antiagainst: Oh, I meant that we want to keep this top-level file more lightweight re-exports, and move kernel impl details in their corresponding files to better categorize and organize.

antiagainst:
I reverted 36babfe for now to make this pull request scoped and easy to review. 36babfe brings in several thousands LOC changes. We can do it later in its own NFC change commit.

tokenspeed-kernel/README.md
antiagainst: Added rules in `AGENTS.md`. :)

PR #1536, opened 2026-09-14
perf(PDL): improve PDL overlap in GDN, gated residual, and QSA

tuanzhangCS:
## Summary

- Improve PDL overlap in GDN, gated residual and QSA while preserving producer-read and weight-readiness ordering. QSA indexing takes an explicit `enable_pdl`; trigger timing remains a per-kernel performance choice.
- Use CuTe QSA for eligible SM100/SM103 decode and MTP verification; use FlashInfer FA2 for prefill/mixed queries, including single-token prefill. Dispatch tests use the API's `is_decode` trait.

## Validation

On `265aff02`, NVIDIA B300 (SM103): **320 tests passed, no skips** — 111 runtime and 209 kernel tests. `pre-commit run --all-files` passed.

The original 86 runtime + 147 kernel suites were also rerun on `6e014a8f`: **233 passed**. These are local GPU runs; draft CI skips the GPU jobs.

<details>
<summary>Reproduce the GPU checks</summary>

Use a clean checkout and matching scheduler/native CUDA builds. Tested with Python 3.12.3, PyTorch 2.13.0+cu130, tokenspeed-triton 3.8.10.post20260906, FlashInfer 0.6.18, CuTe DSL 4.7.1 and pytest 8.4.2.

```bash
export PYTHONPATH="$PWD/python:$PWD/tokenspeed-kernel/python"
export CUDA_VISIBLE_DEVICES=0
python -m pytest -q --tb=short --disable-warnings \
  test/runtime/test_gdn_pdl.py \
  test/runtime/test_qsa_backend.py \
  test/runtime/test_hyperconnection_kernel_boundary.py
# 86 passed
python -m pytest -q --tb=short --disable-warnings \
  test/runtime/test_qwen4_exp.py -k qsa
# 25 passed
python -m pytest -q --tb=short --disable-warnings \
  tokenspeed-kernel/test/nvidia/ops/test_qsa_pdl.py \
  tokenspeed-kernel/test/nvidia/ops/test_qsa_sparse_attention.py \
  tokenspeed-kernel/test/nvidia/ops/test_hyperconnection_capture.py \
  tokenspeed-kernel/test/ops/test_qwen4_exp_qsa.py
# 209 passed
pre-commit run --all-files
```

The existing delayed-producer graph test now covers metadata → top-k → slots, with the global PDL setting opposite to the explicit kernel argument during capture.

</details>

## Agentic benchmark

Fresh measurements on the same four B300 GPUs, run sequentially:

| Revision | Decode tok/s | Mean TTFT (ms) | E2E output tok/s |
| --- | ---: | ---: | ---: |
| Base `a4296863` | 647.79 | 469.57 | 403.03 |
| PDL `265aff02` | 679.92 | 459.19 | 418.82 |

**+4.96% decode throughput, +3.92% end-to-end throughput** against the measured base. Both revisions completed **190/190 measured requests**.

- Model: [`nvidia/Qwen3.8-Flash-Next-NVFP4`](https://huggingface.co/nvidia/Qwen3.8-Flash-Next-NVFP4), revision `fc694b54fb0174e0913e6adf86691ef85a4ead47`.
- TP4, BS1, MTP 3, BF16 KV; 81,920-token context, 8,192-token chunked prefill, GPU prefix cache enabled, host KVStore disabled, `index_share_for_mtp_iteration=true`.
- [Agentic dataset](https://huggingface.co/datasets/lightseekorg/agentic-dataset/resolve/main/agentic_dataset.json), conversations 0–7; two rounds of 95 turns, 500 output tokens per turn, greedy decoding, thinking disabled. Flush prefix cache before each round, then independently warm up on 26 turns from conversations 68–69.

Decode throughput is the reciprocal of pooled mean TPOT. End-to-end throughput uses the two measured request windows; startup and warmup are excluded. Both revisions use identical scheduler/CUDA builds and server options except the local HTTP port.


borontion:
we are doing a lot of moving of `tl.extra.cuda.gdc_launch_dependents()` from the tail to the beginning of the kernel, early enough just after `tl.extra.cuda.gdc_wait()` - is this a common pattern always benefits performance?

tuanzhangCS:
> we are doing a lot of moving of `tl.extra.cuda.gdc_launch_dependents()` from the tail to the beginning of the kernel, early enough just after `tl.extra.cuda.gdc_wait()` - is this a common pattern always benefits performance?

@borontion  Yes, `tl.extra.cuda.gdc_launch_dependents()` allows the next kernel to start executing, so its setup and independent weight loads can overlap with the current kernel's computation. The next kernel then uses `tl.extra.cuda.gdc_wait()` to ensure the data it depends on is ready before consuming it.

minedec (dismissed):

PR #1471, opened 2026-09-10
perf(kimi-k3): extend Iris all reduce window for moe

efric:
## Summary

This patch modifies the Kimi K3 all reduce variant used post MoE on AMD CDNA4: 

### TP8/EP8
```
         M    Prior Candidate                              Current candidate
  ━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
       1–4    Producer-direct Iris one-stage          Producer-direct Iris one-stage
  ─────────  ──────────────────────────────────────  ────────────────────────────────
      5–48    Producer-direct Iris two-stage          Producer-direct Iris two-stage
  ─────────  ──────────────────────────────────────  ────────────────────────────────
   49–8192    Two logical RCCL reductions, grouped    Producer-direct Iris two-stage
  ─────────  ──────────────────────────────────────  ────────────────────────────────
     >8192    Grouped RCCL                            Grouped RCCL
```
### TP8/TP8
```
          M    Prior Candidate                               Current candidate
  ━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
        1–4    Producer-direct Iris one-stage          Producer-direct Iris one-stage
  ──────────  ──────────────────────────────────────  ────────────────────────────────
       5–48    Producer-direct Iris two-stage          Producer-direct Iris two-stage
  ──────────  ──────────────────────────────────────  ────────────────────────────────
     49–292    Concatenate, then one RCCL reduction    Producer-direct Iris two-stage
  ──────────  ──────────────────────────────────────  ────────────────────────────────
   293–8192    Two logical RCCL reductions, grouped    Producer-direct Iris two-stage
  ──────────  ──────────────────────────────────────  ────────────────────────────────
      >8192    Grouped RCCL                            Grouped RCCL
```

i.e, use each prepared Iris state's producer-direct capacity when acquiring and reducing Kimi-K3 MoE outputs. This lets the measured equal-group TP8 layouts keep producer-direct Iris through 8192 tokens instead of falling back at the default 1 MiB limit. Other layouts retain the existing bounded window and RCCL fallback.

## Test Plan

  "Main" means the algorithm selected by the pre-heuristic code.

```
[...]

PR #1539, opened 2026-09-14
fix(kimi-k3): restore EAGLE3 prefill numerics

raikonenfnu:
## Summary

- restore Kimi-K3 producer-direct Iris reductions to the measured 1 MiB window (48 tokens at canonical K3 widths)
- route larger MoE joins, including the 4K prefill, back to RCCL
- retain the generic prepared-capacity support added by #1471 for other callers
- pin the performance job to EvalScope 1.11.1

## Root cause

`4ce76392` (#1471) is the exact public-CI transition. The last completed run before it measured 164.74 tok/s/user with about 3.95 decoded tokens/iteration; the first completed run with it measured 151.75 with 3.6148. Every subsequent completed run retains 3.6148. Across the available history, median latency throughput moved from 161.81 over 11 pre-change runs to 150.95 over 12 post-change runs (-6.7%).

Kimi-K3 already requested a large producer-direct symmetric buffer before #1471, but dispatch still capped producer-direct reductions at the ordinary 1 MiB window and selected RCCL first above that limit. #1471 made prepared-state capacity authoritative and probed Iris before the ordinary size gate. As a result, the large two-output K3 MoE join in 4K prefill moved from RCCL to Iris.

The collective is faster in isolation (410.63 vs 483.23 us at M=4096 in #1471), but Iris and RCCL accumulate BF16 partials in different orders. That prefill rounding difference changes the greedy target trajectory. The EAGLE3 draft then accepts about 8.5% fewer tokens per target iteration, requiring more target-model passes for the same 1K output and erasing the collective gain end to end.

The later failing head commit, `de03e84f`, adds DeepSeek V4-specific operators and is not on this path. Its 142.45 / 16.66 result is a low timing outlier within the lower post-#1471 regime; latest main passed at 149.93 / 17.48 with the identical 3.6148 acceptance trajectory.

## Fix

Cap Kimi-K3 producer-direct preparation at the original 1 MiB dispatch limit for every TP/EP group topology. The limit is byte-based, not a hard-coded row count, so it also preserves the old behavior for noncanonical hidden widths. Decode and small speculative buckets retain Iris; larger prefill joins return to RCCL. This reverts only the Kimi-K3 expansion while preserving #1471 generic backend support for explicitly prepared capacities.

The original 161 / 18.8 performance reference remains in place. EvalScope is pinned so metric semantics cannot drift independently of that reference.

## MI350 A/B result

The unchanged 4K/1K TP8/EP8 gate passed on the PR:

| Metric | Post-#1471 main | Fix |
| --- | ---: | ---: |
| Decoded tokens/iteration | 3.6148 | 3.9498 |
| Latency throughput (tok/s/user) | 150.95 median | 164.47 |
| Output throughput (tok/s/GPU) | 17.59 median | 19.04 |

The fix measured 102.2% and 101.3% of the unchanged latency and throughput references, respectively.

Failed job: https://github.com/lightseekorg/tokenspeed/actions/runs/34803815610/job/103854417190
Passing A/B job: https://github.com/lightseekorg/tokenspeed/actions/runs/34811802356/job/103874589929

## Validation

- MI350 Kimi-K3 EAGLE3 4K/1K TP8/EP8 performance gate (passed)
- `python -m pytest test/ci_system/test_eval_configs.py -q` (6 passed)


raikonenfnu:
> > That change extends the producer-direct Iris all-reduce window from 48 to 8192 tokens for Kimi-K3 TP8/EP8, so the 4K prefill moved from RCCL to Iris.
> 
> 
> 
> Shall we keep to use RCCL? @raikonenfnu 

Hi @lightseek-bot, let me do some experiment locally to double check and confirm. We can revert back to RCCL if indeed the case

raikonenfnu:
Final validation update:

- Final-head (`0b591991`) AMD runtime-1gpu suite passed, including all 13 Kimi-K3 communication arming tests: https://github.com/lightseekorg/tokenspeed/actions/runs/34813144828/job/103878429693
- The MI350 TP8/EP8 4K/1K A/B passed at 164.47 tok/s/user, 19.04 tok/s/GPU, and 3.9498 decoded tokens/iteration: https://github.com/lightseekorg/tokenspeed/actions/runs/34811802356/job/103874589929

The performance run used the immediately preceding commit with identical runtime code. The final commit only corrected the remaining TP4 mock expectation from the row-derived capacity to the byte-derived 1 MiB capacity.

PR #1543, opened 2026-09-14
perf(comms): write iris staged two-stage reductions in place

efric:
## Summary

Make the ordinary staged two-stage reduction write directly into its caller
tensor, removing the separate reduction-output allocation and copy without
changing the symmetric input staging or synchronization protocol.

Eight-GPU gfx950 graph microbenchmarks compare only the copy-out and in-place
forms of the same two-stage kernel on BF16 [M, 7168] inputs. Each entry is the
median of three paired run medians with alternating variant order, 100-160
timed replays per run, and the worst rank reported after per-rank aggregation:

```
  M     Copy-out us   In-place us   Saved us   Gain
  33         33.921        32.201      1.720   5.07%
  36         36.861        34.941      1.920   5.21%
  37         36.881        35.341      1.540   4.18%
  64         42.021        40.321      1.700   4.05%
  128        48.701        46.081      2.620   5.38%
  256        57.341        54.882      2.460   4.29%
  512        72.823        70.822      2.001   2.75%
  1024      109.504       105.123      4.381   4.00%
  2048      180.905       171.965      8.940   4.94%
  4096      337.990       315.648     22.341   6.61%
  8192      650.758       604.256     46.502   7.15%
```

(Kernel selection policy remains at 36 for Iris post attention all reduce in this PR; the comparison at larger M is just to demonstrate that gains persistent with increasing message sizes as well). 

Full model serving benchmarks at top of stack. 

## Test Plan

- Validated with the dependent fused-push layer on eight gfx950 GPUs: all 54
  existing Iris tests passed, including TP2/4/8, graph replay and mixed geometry.
- `pre-commit run --all-files` passed on the stack.


tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/iris.py
```diff
@@ -851,13 +851,9 @@ def __init__(
             if self._producer_direct_scratch_numel
             else None
         )
-        output_max_numel = max(
-            producer_direct_max_numel,
-            staged_max_numel if self._staged_two_stage_supported else 0,
-        )
         self._reduced_output_buf = (
```
raikonenfnu, line 854: Any intuition behind the drop of this?

tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/iris.py
```diff
@@ -1120,7 +1115,6 @@ def _all_reduce_two_stage(
             EXIT_BARRIER=True,
             num_warps=kernel_config.num_subgroups,
         )
-        tensor.view(-1).copy_(output)
```
raikonenfnu, line 1123: smort :)

tokenspeed-kernel/python/tokenspeed_kernel/ops/communication/iris.py
efric: this test exercised both the staged AR and producer direct AR; after this change this buffer is no longer needed for the staged AR variant.

efric: (before it was being used as a staging ground before being copied into the caller's tensor for this variant; now that we write directly into the caller's tensor it's not needed anymore)

nirvedhmeshram (commented):
I think the codex suggestion would be good to incorporate, after that this looks good.

efric:
> I think the codex suggestion would be good to incorporate, after that this looks good.

the output in the k3 caller, which is the only place we use this atm, is already aligned so it's not affected. But for now, i added a fall back to protect against misuse though it's not very satisfying. In the medium term im planning on refactoring the iris layering cleaner, including AR variants and its APi but it's not appropriate scope for this pr stack

PR #1559, opened 2026-09-15
perf(attention): skip fully masked sliding-window prefill tiles

yechank-nvidia:
## Summary

  The portable Triton prefill kernel visits every KV tile before rejecting tiles that are fully masked by a sliding window. Long inputs with small windows spend unnecessary work on those loop iterations.

  - Start the tile loop at the first tile that can intersect a positive sliding window.
  - Bound the causal upper end when there is no custom mask; custom masks can admit future keys.
  - Preserve valid-tile order and per-element masks.
  - Add ten paged attention regression cases and document the bounds.

  ## Test Plan

  Validated commit `dfb99612` on NVIDIA GB200 with BF16 inputs. Baseline comparisons use public upstream commit `eaa58e30`.

      pre-commit run --all-files
      All applicable hooks passed.

      python -m pytest tokenspeed-kernel/test/ops/test_attention_window_bounds.py -q
      10 passed

      python -m pytest tokenspeed-kernel/test/ops/test_attention.py \
        -k 'triton and bf16 and (prefill or extend)' -q
      14 passed, 190 deselected

  Both GPU suites also ran with a local comparison harness that evaluated the public-base and candidate implementations on identical inputs: all 26 calls produced byte-identical outputs and log-sum-exp (LSE) values.

  Twelve additional custom-mask cases—future-key, sparse, and empty masks at two lengths and two windows—matched the base and a Torch reference in eager execution and CUDA Graph replay: 24 comparisons passed with unchanged inputs.

  The focused environment omits the optional DeepGEMM backend. These checks qualify the portable Triton path on GB200.

  ## Performance

  GB200, Torch 2.13.0+cu130, TokenSpeed Triton 3.8.10; BF16, 6 query heads, 1 KV head, head dimension 128, page size 64, causal attention, window 511, no attention sinks or LSE output.

  Each CUDA Graph contains 20 helper launches. Values below are median per-launch times over 12 pairs with alternating execution order, after warmup and with identical fixed inputs. Output bytes matched before and after timing.

  | Input tokens | Public base, µs | Candidate, µs | Reduction |
  | --- | ---: | ---: | ---: |
  | 1024 | 48.262 | 41.778 | 13.44% |
  | 2048 | 65.753 | 42.222 | 35.79% |
  | 4096 | 196.286 | 82.419 | 58.01% |

  The 1024-token full-attention control measured 45.478 → 45.498 µs, a 0.046% increase.

  Timing excludes KV writes and the rest of the model; it does not establish serving latency or results for other GPU vendors.


tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/mha/README.md
borontion, line 1: let's remove this README.md - you can add some comments to the kernel code if needed.

borontion (dismissed):

tokenspeed-kernel/python/tokenspeed_kernel/ops/attention/mha/README.md
yechank-nvidia: Removed!

PR #1563, opened 2026-09-15
feat(kimi-k3): support MoE all-to-all for attention DP

syuoni:
## Summary

Keep Kimi-K3’s dense MoE work local under attention DP, replacing the whole-MoE all-gather with dispatch/combine around routed experts.
* Require attention DP == MoE EP == world size; replicate shared experts and latent projections.
* Support FlashInfer all-to-all and an AG/RS reference transport. Add explicit agrs and flashinfer backend choices while preserving automatic selection with none.
* Quantize NVFP4 activations before either transport. Load common activation scales as maxima across all checkpoint experts.
* Overlap shared-expert computation using the existing auxiliary-stream pattern, joining before the final accumulation.
* Simplify SiTU backend setup and trim the parallelism documentation.

## Test Plan

* [x] Runtime regression suites passed, including 310 focused tests on the final commit.
* [x] 15 kernel tests passed for the quantized-dispatch change.
* [x] 16-GPU BF16/NVFP4 dispatch/combine checks passed with uneven and idle ranks, eager execution, and CUDA graph replay.
* [x] Repository pre-commit hooks passed.
* [x] Validate the final branch with server startup and AIME26. The earlier implementation scored 28/30 (0.9333); that run predates quantized dispatch and shared-expert overlap.


python/tokenspeed/runtime/layers/moe/utils.py
```diff
@@ -64,9 +65,6 @@ def is_none(self):
     def is_deepep(self):
         return self == All2AllBackend.DEEPEP
 
-    def is_flashinfer_nvlink_one_sided(self):
```
byshiue, line 67: Is this backend removed? Or you want to rename it? 
Do we need to consider two side in the future?

syuoni: `FLASHINFER_NVLINK_ONE_SIDED` was a placeholder.

This PR added the real implementation. I didn't find the two-sided implementation in flashinfer, so just renamed it to `FLASHINFER`.

Btw, TRTLLM's two-sided MoE alltoall is less performant, so I don't think it will be integrated to flashinfer in the future.

PR #1564, opened 2026-09-15
Add Kimi K3 AgentX Slurm benchmark harness

Xiangyi1996:
Running Kimi K3 AgentX benchmarks on Slurm requires coordinating server startup, an EvalScope client, cancellation, and result collection. Add `agentx.slurm` to manage these steps for smoke and benchmark runs, with configuration and artifacts retained for inspection.

- Reuse `configs/attn_tp8_moe_tp8.sh` or `configs/attn_tp8_moe_ep8.sh` through `SERVER_SCRIPT`. Both configs accept additional arguments through `"$@"`; their existing settings remain unchanged without overrides. The documented AgentX invocation supplies the 262144-token context, checkpoint paths, seed, and other run-specific values without duplicating the server configuration.
- Support batch submission or an existing allocation, with explicit hold/release behavior. Activate the selected server environment inside the container independently of the EvalScope client environment.
- Track snapshot preparation, GPU preflight, server, client, and audit processes for cancellation. Forward client signals through GNU timeout and terminate the tracked server srun even before its step appears in squeue.
- Snapshot the executing Slurm script, selected server configuration, result auditor, and input trace. Record hashes, ordered server arguments, versions, settings, and exit status so later checkout or source-data edits do not change the recorded run.
- Audit request exports against final profiling counts and retain raw artifacts. Rejected audits preserve their reason and available counts. Smoke cancellations and profiling grace-period timeouts are rejected by this harness; a grace timeout alone does not establish an engine fault or a stalled drain.
- Document the upstream EvalScope revision with tokenizer support and benchmark grace forwarding. Pass the scenario through without dependency patches or throughput recalculation. Grace settings must be consistent across compared runs because they can change the completed request set and observation time.

The branch also includes a one-line scheduler test compatibility fix after updating from main: pass an empty materialized-state-boundary span instead of the old integer argument in the replayable-cache test. Scheduler runtime behavior is unchanged.

Validation:

- `bash -n` for the AgentX launcher and both shared server configs: passed.
- `git diff --check`: passed.
- Scheduler C++ test executable: built successfully; all five `BoundedReplayCoordinator.*` tests passed.
- The previously failing C++ test file passed local clang-tidy compilation diagnostics with an explicit check (the local LLVM wheel differs from the CI toolchain).
- CPU-only execution of both shared configs with a stub `ts` command confirmed that original arguments are retained and overrides are appended with argument boundaries preserved, including paths containing spaces.

GPU smoke validation of the config-reuse launch path is complete:

- Attention TP8 / MoE TP8: one warmup request and five profiling requests completed, with zero cancellations and zero errors; audit passed.
- Attention TP8 / MoE EP8: one warmup request and five profiling requests completed, with zero cancellations and zero errors; audit passed.
- Both runs used eight GPUs, concurrency 1, a 60-second profiling window, 600-second grace, and 262144-token context capacity. Profiling plus drain completed in 69.14 seconds and 75.39 seconds respectively; neither hit the grace timeout.
- Verified effective parallelism and argument overrides, and matched the executed harness, shared configs, and auditor hashes to the current PR files.

These smoke runs validate the launch path on the pinned runtime used for the experiments, not the newly merged main runtime or performance baselines. Keep this PR in draft for design discussion.


test/agentic_benchmark/kimi_k3/tokenspeed/configs/agentx_tp8.sh
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
```
syuoni: Why not reuse the existing `attn_tp8_moe_tp8.sh` and `attn_tp8_moe_ep8.sh`?

syuoni (commented):
Hi @Xiangyi1996 , I am marking this PR as a draft. Let's discuss offline.

test/agentic_benchmark/kimi_k3/tokenspeed/configs/agentx_tp8.sh
Xiangyi1996: Current agentx_tp8.sh mostly reused attn_tp8_moe_tp8.sh, with changes such as the 256K context length for AgentX. You mean adding argument passthrough to the existing configs, so the AgentX launcher can supply overrides without duplicating the shared settings?

Xiangyi1996: Hi @syuoni Updated in `96750224`: removed `agentx_tp8.sh` and added `"$@"` passthrough to the existing `attn_tp8_moe_tp8.sh` and `attn_tp8_moe_ep8.sh`. 
Keeping the PR in draft pending our discussion and a GPU smoke run of the updated launch path.

PR #1569, opened 2026-09-15
perf: capture KDA prefill in CUDA graphs and fix NVFP4 MoE routing padding

byshiue:
## Summary

Capture KDA prefill together with surrounding model operations to reduce CPU launch overhead and attention-boundary copies. This includes the main and tail scans used by incremental prefill with internal checkpoints.

KDA prefill graphs are enabled by default for supported `cutedsl_kda` workloads. Use `--disable-kda-prefill-graph` to disable them while retaining ordinary outer prefill and decode graph settings.

This PR also fixes uninitialized routing-map padding in FlashInfer TRT-LLM NVFP4 MoE, which caused workspace-history-dependent GEMM slowdowns during CUDA graph replay.

## KDA prefill graphs

- Capture neighboring projections, KDA main/tail computation, and post-attention operations together in the existing outer graph.
- Share one capacity-shaped metadata preparation path across supported eager execution, capture, and replay. Refresh sequence boundaries, token mappings, and state-page indices before each forward.
- Reuse captures across input lengths, request counts, and checkpoint patterns. With request capacities `[1, 2, 4]`, BS3 reuses BS4 through masked dummy slots, while BS1 and BS2 retain their exact capacities.
- Preserve scheduler-owned state and checkpoints. Dummy slots cannot update live state pages or contribute to output.
- Fuse scan-input cleanup, FP32 gate conversion, and device-side chunk planning. Restore main/tail outputs with one inverse-map gather.
- Capture configured variants at startup. Incompatible batches use the ordinary outer graph with eager KDA.
- Pass the graph setting through shared server arguments so TP workers receive consistent configuration.

### Configuration

KDA and ordinary outer prefill graphs share the same token-capacity list:

- `--prefill-graph-capture-token-sizes`: total input-token capacities per forward.
- `--prefill-graph-capture-batch-sizes`: request capacities; replay selects the smallest compatible capacity.
- `--prefill-graph-max-tokens`: maximum captured token capacity.
- `--disable-kda-prefill-graph`: disable KDA capture while preserving other graph settings.

The previous spelling, `--prefill-graph-capture-sizes`, remains a compatibility alias. Specify only one spelling.

Without explicit token sizes, the existing relative-spacing ladder is used, normally capped at 2048 tokens. Request capacity defaults to the minimum count required by each token bucket, usually `[1]`.

## NVFP4 MoE routing-padding fix

FlashInfer's routing step populates real assignments in `permuted_idx_to_token_idx`, but expert-tile padding can retain old workspace contents. The first routed GEMM consumes this map when gathering activations, making its latency depend on allocation and graph-pool history.

The kernel-package adapter initializes the complete map, including its guard entry, to `-1` before routing. A stream-ordered `cudaMemsetAsync` executes on every invocation and is recorded inside CUDA graphs, so each replay refreshes the padding before it is consumed.

The fix covers both routing from logits and precomputed top-k, using the native allocation size across token counts, expert partitions, and tile sizes. It retains the native routing and GEMM implementations.

Regression tests poison the workspace between replays, change routing distributions, check live and padded entries, and compare outputs with the upstream operator.

## Performance

Measurements use the real 93-layer Kimi-K3 NVFP4 model on eight NVIDIA GB300 GPUs in one NVLink domain, with TP8, BF16 activations, FP8 KV cache, CuTeDSL KDA, and overlap enabled. Timing runs use neither NSYS nor NVTX.

The paired OFF/ON measurements below were collected at `6a215d48`. Both modes retain outer/decode graphs and include the MoE routing-padding fix; the comparison measures the effect of KDA graph capture.

Settings:

- Token capacities: `[128, 256, 512, 1024, 2048, 4096, 8192]`.
- Request capacities: `[1, 2, 4]`.
- Prefill budget: 16,384 tokens; prefix granularity: 128 tokens.
- Kernel autotuning disabled in both modes.

These are explicit benchmark settings, rather than the default capture configuration.

### Incremental prefill

Each case has 12 measured batches per mode across two independent server processes, after warmup, in OFF–ON–ON–OFF order.

Times are full-model server prefill medians, including CPU and communication gaps within the measured interval. The artificial wait used to form each fixed-size batch is excluded.

| BS | New tokens per request | KDA graph OFF | KDA graph ON | Latency reduction |
|---:|---|---:|---:|---:|
| 1 | 868 | 128.270 ms | 105.690 ms | 17.60% |
| 1 | 1537 | 144.430 ms | 141.955 ms | 1.71% |
| 1 | 6689 | 402.720 ms | 401.140 ms | 0.39% |
| 2 | 868–869 | 163.825 ms | 148.455 ms | 9.38% |
| 2 | 1537–1538 | 235.200 ms | 230.985 ms | 1.79% |
| 4 | 64–67 | 139.480 ms | 107.480 ms | 22.94% |
| 4 | 868–871 | 251.500 ms | 246.900 ms | 1.83% |
| 4 | 1537–1540 | 415.400 ms | 410.730 ms | 1.12% |

### Request-capacity reuse

The padding validation at `19cae968` compares against `6c3f1531`, with KDA graphs enabled in both versions. Each case has six measured batches after two warmups.

| BS | New tokens per request | eager KDA fallback | BS4 graph replay | Latency reduction |
|---:|---|---:|---:|---:|
| 3 | 64–66 | 180.980 ms | 86.530 ms | 52.19% |
| 3 | 128 / 128 / 128 | 180.825 ms | 101.075 ms | 44.10% |
| 3 | 868–870 | 234.535 ms | 230.065 ms | 1.91% |
| 3 | 1537–1539 | 395.035 ms | 390.195 ms | 1.23% |

Replay records confirm BS3→BS4 selection. The large short-input gains come from capturing work that previously executed at the eager KDA break. The eleven BS1/2/4 cases changed by −0.75% to +0.27%.

These follow-up measurements are sequential runs on the same allocation; small differences warrant repeated measurements.

### End-to-end agentic workflow

The benchmark replays frozen tokenized SWE-smith conversations: four measured conversations with 11 turns each, including 40 incremental requests. Initial prompts target 50,000 tokens; later turns contain variable-length tool payloads. Each request generates 500 tokens, totaling 22,000 generated tokens per run.

The table reports median wall time over two runs per mode at `6a215d48`, including initial prefill, incremental prefill, decode, and client/server handoffs. Model loading, graph capture, and warmup are outside the timing window.

| Concurrency | KDA graph OFF | KDA graph ON | Wall-time reduction |
|---:|---:|---:|---:|
| 1 | 241.474 s | 240.737 s | 0.31% |
| 2 | 140.623 s | 140.436 s | 0.13% |
| 4 | 86.917 s | 85.744 s | 1.35% |

For the serial workflow, the summed server prefill time of the 40 incremental requests decreases from **7.491 s to 7.125 s**, or **4.89%**. These are medians of run totals.

Concurrency is a client setting rather than a guaranteed prefill batch size. Concurrent scheduling can change batching, cache hits, and generated outputs. The small E2E differences should therefore be read alongside run-to-run variation.

## Correctness and regression coverage

Validation uses GB300 GPUs with PyTorch 2.13.0+cu130, FlashInfer 0.6.18, and the CuTeDSL KDA backend.

- **OFF/ON parity at `6a215d48`:** all 38 paired responses across BS1/2/4 match for eight-token greedy generation at matching cache states. Both paired serial agentic runs also match all 44 output sequences and cached-token counts.
- **Request-padding validation at `19cae968`:** 30 cold/warm batches covering 70 requests match the unpadded baseline for generated tokens and cached-token counts. All 90 measured fixed batches match baseline outputs. The serial agentic run matches all 22,000 generated tokens.
- **Latest configuration validation at `1632fcd1`:** 189 tests and 52 subtests pass, with three expected optional/platform-specific skips. Coverage includes default-on behavior, explicit disable, shared configuration, CLI routing, and native GPU graph replay.
- **KDA state coverage:** output restoration, convolution/recurrent states, checkpoints, partial-to-full request occupancy, token-capacity boundaries, and changing replay metadata.
- **MoE coverage:** poisoned routing-map padding, changing routes, local expert partitions, PDL on/off, eager execution, graph replay, and output equality.
- **CI integration:** the KDA graph regression file is registered in `runtime-1gpu`, with a direct pytest entry point. Native replay cases run on supported NVIDIA SM100/SM103 devices.

Real-model parity compares equivalent cache states. Some cold/warm outputs differ in both graph modes; the parity checks do not assert cold/warm determinism. The SWE-smith workload supplies performance and output-parity evidence rather than a task-accuracy score.

The performance tables are revision-specific. The latest configuration changes have passed regression testing; their full-model performance has not been remeasured.


python/tokenspeed/runtime/utils/server_args.py
```diff
@@ -1918,15 +1907,41 @@ def add_cli_args(parser: argparse.ArgumentParser):
...
+            metavar="TOKENS",
             type=int,
             nargs="+",
-            help="Explicit list of token-bucket sizes to capture for the "
-            "breakable prefill graph (like --cudagraph-capture-sizes for "
-            "decode). Unset: a relative-stride ladder bounding padded compute "
-            "at ~12.5%% of any size.",
+            help="Compatibility alias for --prefill-graph-capture-token-sizes. "
+            "Specify only one spelling.",
+        )
```
syuoni, line 1939: We can merge these two args as one with alias:

```python
parser.add_argument(
    "--prefill-graph-capture-token-sizes",
    "--prefill-graph-capture-sizes",
    dest="prefill_graph_capture_sizes",
    metavar="TOKENS",
    type=int,
    nargs="+",
    help="...",
)
```

python/tokenspeed/runtime/execution/prefill_graph.py
```diff
@@ -167,6 +170,31 @@ def _prefill_bucket_step(size: int) -> int:
...
+    Counts outside a particular bucket's range are skipped, not padded with
+    empty sequences. Invalid configured counts fail before capture. The result
+    is sorted and deduplicated; backend support is checked separately.
+    """
+    context = max(1, int(config.context_len))
+    minimum = -(-token_bucket // context)
+    maximum = config.max_num_seqs // config.data_parallel_size
+    sizes = config.prefill_graph_capture_batch_sizes
+    if sizes is None:
+        sizes = [minimum]
```
syuoni, line 189: Just to double check: Since normally we have context >> token_bucket, so the default bs buckets seems `[1]`, only for bs=1. Correct?

python/tokenspeed/runtime/execution/prefill_graph.py
```diff
@@ -232,6 +260,9 @@ def __init__(
         self._embed_tokens = getattr(self.inner_model, "embed_tokens", None)
         self._input_embeds_buf: torch.Tensor | None = None
         self.attn_backend = attn_backend
+        # (token capacity, exact BS) -> (capture, outputs, metadata bindings).
+        # Checkpoint request counts are refreshed metadata, not another key.
+        self._inline_captures = {}
```
syuoni: Can we move this to next to the `self._captures`? Also, please add the corresponding type annotation.

May need a comment to explain the "inline" meaning. E.g., "Captures with supported prefill attention inlined in the prefill CUDA graph."

python/tokenspeed/runtime/execution/prefill_graph.py
```diff
@@ -622,9 +682,19 @@ def replay(
                 ib.mrope_positions_buf[:, num_tokens:bucket].zero_()
             else:
                 ib.positions_buf[num_tokens:bucket].zero_()
-        with self._padded_to(ctx, bucket):
-            self._captures[bucket].replay(valid_rows=num_tokens)
-        hidden_states, aux_hidden_states = self._outputs[bucket].sliced(num_tokens)
+        cap, output = self._captures[bucket], self._outputs[bucket]
+        with ExitStack() as stack:
+            if self.attn_backend.step_counter is None:
```
syuoni: Seems it's conflict with PD disagg -- the KDA-inlined prefill graph cannot be enabled on the prefill instance. We may revisit this.

docs/design/kda-prefill-subgraphs.md
```diff
@@ -0,0 +1,165 @@
...
+- Data parallelism and layerwise prefill/decode cache transfer retain the
+  ordinary outer route.
+- Forwards above the largest token bucket run eager. Requests with longer
+  prompts can still use graphs for individual scheduled chunks that fit.
+- Other KDA backends retain their existing execution behavior.
+
+Within an ordinary outer graph's pure-prefill attention break, the same switch
+can also enable a separate per-layer KDA subgraph. It uses the outer token
+bucket and captures lazily: the first eligible call warms the kernels, the
+second captures and replays, and later calls replay. It requires compatible
```
syuoni: Does this mean it captures graphs lazily during serving time? I would expect capturing all graphs during the startup time, and if the bs is not compatible with the captured graphs, fallback to the outer graphs.

Capturing graphs lazily also will increase the memory consumption during the serving time, as the server experiences more and more combinations of bs and tokens, potentially leading to OOM.

syuoni (commented):
Thanks @byshiue !

When I started reviewing this PR, I was considering if it's possible to make MLA capturable so we can capture full graph for the prefill forward. However, the current prefill MLA requires a chunk loop over the prefix cache lengths, which means it at least introduce a 3rd dimension of the captured graphs. So it's not feasible in a straightforward way. We can revisit this.

syuoni:
Hi @nperrin-fr , could you please review this PR? Thanks!

python/tokenspeed/runtime/execution/prefill_graph.py
byshiue: Correct.
This limits the default merged KDA captures, not the batch sizes the server can process. For merged captures at BS 1, 2 and 4, users need to set --prefill-graph-capture-batch-sizes 1 2 4. Other batch sizes retain the ordinary outer-graph path.

python/tokenspeed/runtime/utils/server_args.py
byshiue: Yes — we can use exactly that. It is the standard, simpler way to define aliases, and both populate prefill_graph_capture_sizes.

The only behavioral difference is when both appear:
```
--prefill-graph-capture-token-sizes 1024 --prefill-graph-capture-sizes 2048
```
Argparse accepts this and uses [2048] — the last value wins.

The separate declarations enforce the previously agreed “using both raises an error” rule. If that restriction is unnecessary, I can update the code.

python/tokenspeed/runtime/execution/prefill_graph.py
byshiue: Yes, when layerwise transfer is active on the prefill instance. The guard checks step_counter is not None, rather than the disaggregation mode itself.
The ordinary outer path preserves the per-layer cache-ready notifications on every forward. Simply removing this guard would bypass that contract. Supporting inline KDA here requires replay-safe cache-ready signaling with the same transfer ordering. I’d keep the guard until that is implemented and validated.

python/tokenspeed/runtime/execution/prefill_graph.py
byshiue: Agreed. _inline_captures belongs next to _captures and _outputs, with an explicit type for the (token_capacity, exact_bs) key and the capture/output/bindings entry.
Your suggested comment also explains “inline” more clearly: supported attention executes inside the prefill graph segment rather than at an eager attention break.

docs/design/kda-prefill-subgraphs.md
byshiue: Yes—the paragraph describes the separate per-layer fallback cache. The merged outer+KDA variants are captured at startup, but this fallback can still capture during serving.
Its schedules are keyed by batch size, token bucket, stream and PDL setting. There is no explicit cache budget or eviction policy. Shared scratch pools reduce duplication, but retained graphs, metadata and input references can still increase memory usage as new schedules appear.
I suggest removing this lazy fallback: capture the configured merged variants at startup, and use ordinary outer graphs with eager KDA for incompatible batches. That would make graph allocation predictable at startup. Fallback coverage and performance should be rechecked after the change.

nperrin-fr:
The `(token bucket x batch size)` capture matrix looks right to me. My question is about `inline` being a mode rather than the only shape.

`inline` today means two things: skip the attention break at KDA, and swap `backend.forward_metadata` for a capacity-shaped object. The second is the real content and the first follows from it. The mode exists only because KDA keeps its live-shaped metadata as the default, so `KdaOuterGraphBinding.bind()` has to switch between two shapes, `prefill_graph_inline` has to decide whether `@break_point` applies, and `PrefillGraph` has to carry `_captures` and `_inline_captures` side by side. My concern is less this PR than the precedent: the next backend that wants to be captured arrives with its own binding, its own flag and its own graph cache.

Note that everything else in the model already runs padded. The outer graph pads every bucket through the projections and the MoE, as the design note says. KDA is the only op that insists on live shapes, which is why the break exists for it at all.

So would this be simpler: make the capacity shape KDA's only shape, and let the outer breakable graph capture padded bs x padded tokens with nothing KDA-specific in it? Padded requests would write their state to a sink page, the way padded rows are already handled elsewhere. The request dimension then becomes symmetric with the token dimension (a few captured points, pad up to the nearest) instead of an exact-match lookup, so `--prefill-graph-capture-batch-sizes` would not be needed either.

Most of the machinery already seems to be here: `prepare_capacity_scan`, the fixed body/tail slots with negative-index masking, and a host-side plan that already depends only on the capacity (`_route_for_workspace` is fed synthetic `bounds = i * token_capacity`, chunk counts are `cdiv(capacity, block) + N - 1`). On the engine side `state_write_padding_pool_index` already exists for padded rows.
(or we can pad to very few bzs since it should only affect small prefills)

What I cannot tell from reading is whether there is a concrete blocker: are zero-length requests benign in the scan and the convolution, and is there a sink block for `ssm_states[state_out_blocks]`? If those hold, this direction would retire `prefill_graph_inline`, `KdaOuterGraphBinding`, `compatible()`, `_inline_captures` and `KdaPrefillGraphCache`, while keeping the parts that carry the value: the capacity-planned scan and the masked checkpoint slots.

nperrin-fr:
A precedent for this in-tree: the Inkling sconv backend already takes the padded shape as its only shape. The `causal_conv1d` Triton kernel accepts `pad_slot_id` as a first-class argument and returns early on it (`causal_conv1d.py:103`). `InklingAttnBackend.init_prefill_graph_state` reserves a PAD request row (`_pfg_max_bs = min(max_bs, conv_pool.num_slots - 2)`), keeps that row's table at all `-1` and its cache index at `PAD_SLOT_ID`, and then pads both dimensions per step: `_pfg_qsl[bs + 1:].fill_(total)` turns surplus request slots into empty chunks, `_pfg_seq_idx[total:].fill_(max_bs)` assigns surplus tokens to the PAD request. Admission is by bound rather than exact match (`extend_total <= _pfg_seq_idx.shape[0] and bs <= _pfg_max_bs`), and its docstring notes that once armed, every extend (eager, capture and replay) routes through the same persistent buffers. No inline mode and no backend-specific graph were needed there.

The caveat is that this was acceptable for sconv because padding measured as nearly free for it, and as far as I know that has not been measured for KDA. There is a reason to expect it is not free here: `KdaPrefillCapacity` bounds the compact plan at `ceil(token_capacity / chunk_size) + num_sequences - 1` chunks, so each padded request slot adds a chunk, and the relative cost is worst at small token counts, which is exactly where this PR's largest wins are (BS 4 at 64 to 67 new tokens). Padding the token dimension has the same shape of cost, bounded by the bucket ladder.

Also worth noting that a padded request is not representable today: `KdaPrefillCapacity.validate` raises on `(lengths <= 0).any()`, so zero-length sequences are rejected rather than treated as inactive.

So the concrete question is just how much padding costs the KDA scan. If it is cheap, the Inkling shape applies directly and the mode machinery goes away. If it is not, the middle ground still retires the mode: keep a small ladder of request-count buckets and pad up to the nearest one, symmetric with the token buckets, instead of an exact-match lookup. Either way the outer graph enumerates and nothing is KDA-specific. A sweep of scan time over (live tokens, token capacity, request count) would settle it, and I am happy to run one if that is useful.

byshiue:
@nperrin-fr 

Thanks for the Inkling reference. I’ve implemented the metadata cleanup locally, keeping request-capacity padding as a separate follow-up.

Supported KDA prefill forwards now use a common capacity-shaped metadata preparation path for eager execution, capture, and replay. This removes KdaOuterGraphBinding, the temporary prefill_graph_inline flag, and the separate inline-capture table. Startup retains the buffers needed by captured shapes; uncaptured shapes use temporary metadata without creating new graphs. Exact-BS matching remains unchanged.
The padding sweep suggests being cautious about a larger shared request capacity. With token capacity held fixed, padding BS1 to four slots increased KDA-body time from 75.699 to 96.205 µs (+27.09%) for a 64-token increment. For BS2 with 868/869-token increments, padding to four slots increased it from 147.520 to 159.181 µs (+7.90%). These are single-GPU, synthetic KDA-body measurements using TP8 per-rank geometry, with CUDA graphs on both sides—not full-model latency measurements.

A nearest-fitting [1,2,4] policy would preserve exact shapes for BS1/2/4 and add BS3→4 coverage. I’m leaving that for the next step, after addressing padding costs and validating masked state writes.
GPU regressions for the refactor passed: 309 tests, plus 148 subtests, with three skips. Coverage includes changing lengths, checkpoint patterns, state pages, fallback selection, and pool rebinding. Full-model performance and AIME validation are still pending.

test/runtime/execution/test_kda_prefill_graph_cache.py
```diff
@@ -0,0 +1,764 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+"""Regression coverage for the capacity-based prefill graph owner."""
+
+from types import SimpleNamespace
+
+import pytest
```
nperrin-fr, line 27: This file still has no `register_cuda_ci(...)` and no `unittest.main()` entry point, so the suite runner skips it (compare `test_breakable_cuda_graph.py`), which leaves the new padding and capacity-selection coverage out of CI.

nperrin-fr (dismissed):
Conditional approve.

The metadata unification and the removal of the serving-time per-layer graph cache address what I raised: eager, capture and replay now share one capacity-shaped path, and the outer graph enumerates `(token bucket, request count)` with the ordinary capture as the fallback column. The request padding is guarded correctly as far as I can follow it: dummy slots carry one token rather than a zero-length scan, and `_merged_capture_bs` requires `input_num_tokens + bs - ctx.bs <= bucket`, so a batch that already fills its bucket declines the padded capacity instead of overflowing the plan. Thanks for running the padding sweep; nearest-fitting `[1, 2, 4]` keeps BS 1 and 2 on exact shapes, which avoids the case your numbers flag.

The one thing I would like before merge is the CI registration noted inline: the file that covers the new padding logic is still not collected by any suite.

Two non-blocking follow-ups, either here or later: the feature switch is still a bare `os.environ.get` in the backend constructor rather than a server arg, and the capture count depends on it per process, so ranks that disagree diverge during capture; and the FlashInfer NVFP4 routing-padding fix is still bundled here despite having no code coupling to KDA.

test/runtime/execution/test_kda_prefill_graph_cache.py
byshiue: Thank you to remind. I have fixed it and will try to launch a CI to see if the new test is covered.

byshiue: The issue if fixed and I have confirmed that the CI has run the new unit tests.

python/tokenspeed/runtime/utils/server_args.py
syuoni: Makes sense; let's keep it. Thanks for the explanation.

syuoni (approved):
LGTM, thanks @byshiue !

tokenspeed-kernel/python/tokenspeed_kernel/thirdparty/flashinfer/trtllm_moe.py
```diff
@@ -0,0 +1,159 @@
...
+    lines = (
+        "// Initialize tile padding and any guard entry before routing writes live rows.",
+        "// Graph-pool reuse can overwrite this storage: initialization must replay.",
+        "CHECK_CUDA_ERROR(cudaMemsetAsync(",
+        "    permuted_idx_to_token_idx.data_ptr(), 0xff,",
+        "    static_cast<size_t>(permuted_idx_to_token_idx.numel()) * sizeof(int32_t),",
+        "    get_stream(hidden_states.device())));",
+    )
+    initialization = "\n" + "\n".join(indent + line for line in lines)
+    return source[: match.end()] + initialization + source[match.end() :]
```
syuoni, line 65: This introduces a memset for all decode layers.

Is it a flashinfer bug? As a solution, this is a really hacky one. I would expect all positions of `permuted_idx_to_token_idx` should be correctly written by the sort kernel. If it's a flashinfer bug, let's fix it in flashinfer and upstream the fix.

PR #1578, opened 2026-09-15
fix(amd): keep gfx1250 MXFP4 32-bit operands narrow following TDM requirements

jerryyin:
This PR addresses a few compiler failures on the gfx1250 MXFP4 MoE kernels, all the same shape: a value formed in the kernel's index type (int64 when indices are upcast) feeding an instruction operand that is 32-bit regardless.
  - TDM descriptor row extent and column offset fail to compile because the descriptor is an SGPR tuple and only takes 32-bit scalars.
  - `_enforce_wave_uniform_i32` on the scatter path fails because its `v_readfirstlane_b32` takes a single VGPR, so a widened `pid_n` arrives as a register pair and the assembler rejects it.
  - The no-scatter output store hand-built an i32 element offset for `buffer_store`. Replaced with `tdm.async_store` on a `shape=(eM, yN)` descriptor: the descriptor extent becomes the bound, so both masks and the offset disappear.

  Testing: 
 - `test_gluon_mxfp4_situ_gfx1250.py`, added a case that runs MoE apply narrow, forces `should_upcast_indices` true, and runs it again 
 - Added `num_tokens=65`, the smallest size above the prefill  `BLOCK_M` of 64 and so the only one that evaluates `off_m` away from zero.


tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx1250/moe/mxfp4/decode.py
```diff
@@ -326,13 +328,17 @@ def _matmul_decode(
     out = out.to(Y.dtype.element_ty)
     out = gl.convert_layout(out, BLOCKED_LAYOUT_Y)
 
+    OUTPUT_SHARED_LAYOUT: gl.constexpr = gl.SwizzledSharedLayout(
```
knwng: As discussed offline, we need to use padded shared layout here

borontion (approved):

PR #1584, opened 2026-09-16
fix(cache): retain exact decode checkpoint publication evidence

byshiue:
## Summary

This fixes a decode checkpoint-publication issue as part of the replay-SSM work.

Decode can write an aligned state checkpoint before the scheduler's conservative prefix-publication frontier reaches it. If another aligned endpoint arrives first, retaining only the newest checkpoint loses the older one. A later request then misses a reusable prefix.

## Changes

- Keep every unpublished exact checkpoint in request cache progress. Accepted feedback records aligned endpoints, including when multiple results land before the next admission.
- Publish all recorded checkpoints covered by the new hashes, even when the frontier jumps past several boundaries. Remove entries only after successful admission so a failed attempt can retry.
- Keep crossed-but-unwritten boundaries uncached, and preserve the existing conservative allocation and retention rules.
- Update the design notes and regression coverage for admission, finish, and retraction publication.

## Validation

Validated commit `2fe5c968` on the branch's main snapshot `764a549f`, using a CMake Release build on Linux aarch64, GCC 13.3 and Python 3.12.13.

Synthetic scheduler tests cover the original 128-token boundary/four-token decode case and four-token boundaries with decode widths 3, 4, and 12. They check live and finished parent requests, overlap depths 0 and 1, multiple pending checkpoints, skipped boundaries, failed-admission retry, and state pages smaller than the prefix granularity. The new reuse regression fails on the previous revision and passes with this fix.

- 485 C++ scheduler tests passed.
- 153 Python scheduler-binding tests passed using the rebuilt extension.
- `pre-commit run --all-files` and `git diff --check` passed.

Validation is CPU scheduler-level; no new model accuracy or GPU performance result is claimed.


FC-Li:
Thanks for tracking this down — the missed reuse is real, and I could reproduce it with your scenario tests. I think the root cause sits one level lower than the single-record overwrite, though, and fixing it there makes most of the new machinery unnecessary.

## Root cause: the decode frontier is not "conservative", it is wrong

`scheduleDecode` and `retractVictim` derive the computed-token count as `TokenSize() - decode_input_tokens` ([forward.cpp#L513](https://github.com/lightseekorg/tokenspeed/blob/764a549f/tokenspeed-scheduler/csrc/scheduler/operations/forward.cpp#L513), [#L659](https://github.com/lightseekorg/tokenspeed/blob/764a549f/tokenspeed-scheduler/csrc/scheduler/operations/forward.cpp#L659)). Feedback lands the accepted tokens plus one sampled token, so the exact frontier is `TokenSize() - 1`; the finish path already uses exactly that ([outside_event_handler.cpp#L113](https://github.com/lightseekorg/tokenspeed/blob/764a549f/tokenspeed-scheduler/csrc/scheduler/outside_event_handler.cpp#L113)). The formula dates from #447, whose comment reads "Decoding's pending tail is not yet computed" — a model where the container's last T tokens were the pending inputs, which stopped being true once feedback carried variable-length acceptance. With T = 1 the two agree; with speculation on, hashing lags the real endpoint by up to T − 1 tokens.

That lag is the whole reason an aligned accepted endpoint is not hashed in the admission that follows it, and therefore the reason a second aligned endpoint can arrive before the first is published. The retained list compensates for the lag rather than removing it.

Evidence: I ported your two scenario tests (`RetainsExactDecodeBoundaryUntilConservativePublication`, `PublishesEveryPendingDecodeBoundary`) onto `main` and changed only the frontier — no list, no coordinator loop, the single-int record and equality check untouched. On `main` every aligned-endpoint case fails; with the frontier fix alone all of them pass, including the T = 12 cases and "crossed S_4 but never wrote it".

## What still needs per-landing evidence

One case survives the frontier fix: the overlap schedule plans step k+1 before step k commits, so when a request sits out a round, two results land back to back before its next admission. If both endpoints are aligned (4 then 8 with P = 4), the older one has written its state and its page is hashed in that admission, but a record resolved from the current token count only names 8. I added a scenario test for exactly this; it fails with the frontier fix alone and passes with landing-time recording. So the direction of recording in `ExtendTokens` is right — the motivation is multiple landings per admission, not a lagging frontier.

## Branch

https://github.com/lightseekorg/tokenspeed/compare/main...fix/exact-decode-frontier — two commits, both crediting you as co-author since the tests and the list structure are adapted from here:

- [8f3deef5](https://github.com/lightseekorg/tokenspeed/commit/8f3deef590730295d99875f3aa7eedc777dff9e4) `Request::NumComputedTokens()` as the one frontier for prefix hashing and retention (scheduled window end while prefilling, `TokenSize() - 1` while decoding), used by `scheduleDecode`, `schedulePrefill` and `retractVictim`. Your two scenario tests, adapted, as the regression coverage. Docs stop describing the frontier as conservative.
- [2b7fdff1](https://github.com/lightseekorg/tokenspeed/commit/2b7fdff1ca7d5f8ac5e62ae61cad53f429dee7b9) `materialized_state_boundaries` recorded at landing, plus the double-landing test. Differences from this PR worth calling out:
  - Non-closed groups keep **one** publication routine: the existing lookback path, iterated per boundary. State selects which boundaries qualify instead of getting its own `keys.last(1)` loop.
  - The remote-prefill aligned endpoint still counts as evidence. `Request::MaterializedStateBoundaryTokens()` used to claim `PrefillDone`'s aligned window end regardless of source; recording only for `kLocal` at first-chunk admission silently drops that on the D role, whose endpoint slot holds the transferred snapshot.
  - `DiscardHashedStateBoundaries` rather than "published": the list tracks hashing, and the coordinator decides what to publish.

Validation on that branch: 486 C++ scheduler tests and 153 Python binding tests pass with the rebuilt extension; `pre-commit run --all-files` clean.

One thing to keep an eye on when merging either version: with the exact frontier, sliding-window retention reclaims up to T − 1 tokens earlier than today. `ExpiredBlocksAt` keeps positions ≥ C − W + 1 and `SwaMatcher::pagesNeededToResume` wants the window's last W − 1 tokens, so the two agree and T = 1 has always run at zero slack — but it is CPU-level reasoning, so a CI pass over the SWA + MTP model tests would be good before this lands.

Happy to go either way: rebase this PR onto 8f3deef5 and trim the conservative-frontier wording from the docs, or I open the branch as a PR and you review it. Which do you prefer?

FC-Li:
Opened #1597 with the branch above so the two can be compared side by side; requested your review there, @byshiue.

byshiue:
> Opened #1597 with the branch above so the two can be compared side by side; requested your review there, @byshiue.

Thank you. I have reviewed and approve the PR. 

Mark this PR as dupliacated of https://github.com/lightseekorg/tokenspeed/pull/1597

PR #1589, opened 2026-09-16
fix(cache): allow uneven ordinary cache groups

yechank-nvidia:
## Summary

  Ordinary cache setup rejects valid uneven attention groups. A 1:4 full/sliding layout requires a padding fraction of 3.0, exceeding the fixed 1.0 limit.

  Remove this limit through the existing recipe property. Geometry validation and capacity sizing continue through the shared cache code and profiled memory budget.

  ## Test Plan

  Extend the existing MHA cache-budget test with one 1:4 regression case:
  - Before: cache setup fails with a padding-limit error.
  - After: the cache pool is created within the byte budget, including the null parent.

  Validation:
  - `python -m pytest -q test/runtime/test_cache_setup.py`: 24 passed.


yechank-nvidia:
@codex review

PR #1595, opened 2026-09-16
feat(dsv41): support basic prefill/decode disaggregation with DSpark for DeepSeek V4.1 Flash

wangbo981016:
## Summary

DeepSeek V4.1 Flash now runs under prefill/decode disaggregation, with the same-checkpoint DSpark drafter enabled on both roles. The draft's per-stage context windows are KV-cache state, so prefix reuse, PD transfer and the L2 tier carry them together with the target KV.

Until now V4.1 shared the V4 drafter, which keeps those windows in a request-private ring and rebuilds them after a prefix hit by replaying the last 128 prompt tokens; that path stays as is for V4.

### Cache recipe (`recipes/deepseek_v41.py`)

- The four history groups carry `full_suffix` transfer policies under PD. No V4.1 field is head-sharded, so unequal prefill/decode TP works through the replicated path of the cache-transfer planner.
- Sliding windows protect two verify widths on every role. The window is peer-contract state; a prefill node schedules without overlap and a decode node with it, so retention cannot depend on the local schedule depth.
- With DSpark, the SWA group carries three extra fields on the last target layer, `layer.39.dspark_kv{0,1,2}`: one BF16 `head_dim` row per position and draft stage, addressed by the target's SWA slots and cached, transferred and evicted together with the SWA rows. The LCM parent grows to 1,571,328 bytes and the other groups repack (r2 22, r1 66, tail 62 pages per parent; padding 1.5% / 4.5% / 4.5% / 3.1%). Without DSpark the layout is unchanged.

### Backend and pool

- `DeepseekV41AttentionBackend` exposes `window_slots(group_id, ...)`, the 128-row window resolver the decode path already used for SWA.
- `DeepseekV41CachePool.zero_new_blocks` clears freshly admitted pages (a decode node waits on it before publishing destination pages) and exposes `dspark_kv(stage)`.

### DSpark drafter

- `DeepseekV41DSpark` (new) writes each stage's rows at the target's SWA write slots for the token rows the target just processed and gathers its 128-row history through `window_slots`. It holds no per-request state. `DeepseekV41ForCausalLMDSpark` dispatches to it; the V4 drafter is unchanged.
- V4.1 advertises a zero prefix-replay requirement. Same-checkpoint DSpark accepts KVStore and disaggregated serving; the engine decides from the resolved draft config.

### Other

- EPD encoder facts resolve the `vision` module and take its dtype from the parameters, so a multimodal V4.1 prefill node starts.
- `docs/design/cache-concepts.md` records where a same-checkpoint DSpark draft keeps its context (extra fields of the target's SWA group) next to the existing block-drafter rule.
- CI: `test/ci/ut/deepseek-v4.1-flash-pd-1p1d.yaml` (b200-4gpu, TP2 + TP2, manual trigger until validated on that runner), `test/ci_system/serve_deepseek_v41_flash_pd_1p1d.sh`, `test/runtime/distributed/test_deepseek_v41_pd_1p1d.py`.

## Launch

Both workers share the model flags. The prefill worker adds the bootstrap port; the decode worker disables prefix caching (its sliding-window groups land only their retained tail, so local prefix probes cannot match).

```bash
COMMON="--model deepseek-ai/DeepSeek-V4.1-Flash --trust-remote-code \
  --tensor-parallel-size 4 --moe-backend marlin --dtype bfloat16 \
  --max-model-len 32768 --max-total-tokens 262144 --max-num-seqs 16 \
  --chunked-prefill-size 8192 --gpu-memory-utilization 0.95 \
  --max-cudagraph-capture-size 8 --disable-prefill-graph --disable-kvstore \
  --engram-host-table --language-model-only \
  --speculative-algorithm DSPARK \
  --disaggregation-transfer-backend mooncake --disaggregation-layerwise-interval 0"

# prefill
CUDA_VISIBLE_DEVICES=0,1,2,3 python -m smg_grpc_servicer.tokenspeed $COMMON \
  --world-size 4 --port 18346 --dist-init-addr 127.0.0.1:12579 \
  --disaggregation-bootstrap-port 8998 --disaggregation-mode prefill

# decode
CUDA_VISIBLE_DEVICES=4,5,6,7 python -m smg_grpc_servicer.tokenspeed $COMMON \
  --world-size 4 --port 18347 --dist-init-addr 127.0.0.1:13580 \
  --disaggregation-mode decode --disable-prefix-caching

# router
python -m smg launch --pd-disaggregation \
  --prefill grpc://127.0.0.1:18346 8998 --decode grpc://127.0.0.1:18347 \
  --port 18345 --model-path <model> --tokenizer-path <model> \
  --reasoning-parser deepseek_v31
```

On Blackwell use `--moe-backend mega_moe --enable-expert-parallel`; the CI script selects this for `MOE_BACKEND=mega_moe`. Let Mooncake choose its transport; do not force the intra-node NVLink transport.

### Unit tests

`test_deepseek_v41_cache.py`, `test_deepseek_v41_dspark.py` (including the CUDA graph and drafter tests), `test_prefix_replay_scheduler_config.py`, `test_dspark_proposal.py`, `test_cli_config_compat.py`, and the PD contract / planner / manifest tests pass on an 8× H20 node.

### End to end

8× H20 (143 GB), prefill TP4 + decode TP4, Mooncake over RDMA. evalscope 1.11.1 gsm8k, 100 samples, greedy, concurrency 8, `max_tokens` 30000.

| Configuration  | gsm8k accuracy | Mean accept length |
|----------------|:--------------:|:------------------:|
| fused          | 0.98           | –                  |
| fused + DSpark | 0.98           | 3.65               |
| 1P1D           | 0.98           | –                  |
| 1P1D + DSpark  | 0.98           | 3.69               |

Each 1P1D + DSpark request transfers 72 field pages — SWA 43 (40 layers + 3 DSpark stages), global 8, compressor tail 21 for a 43-token prompt — 1.81 MB per rank.

### CI smoke test

`test_deepseek_v41_pd_1p1d.py` (TP4 + marlin on the same node) passes: all four answers match, and the decode node accepts draft tokens (27 of 35 completion tokens on the longest prompt).

## Notes

- V4.1 transfers the cache once per completed prompt. `forward_v41` records no per-layer cache step, so the config rejects a non-zero `--disaggregation-layerwise-interval` on PD roles.
- With a drafter-private context (V4 DSpark), KVStore is refused whether or not prefix caching is enabled; V4.1's cache-resident context accepts it.
- Decode-side prefix caching stays off in the CI topology; enabling it is correct but yields no hits for sliding-window groups.


wangbo981016:
@codex review

FC-Li:
Restacked this PR on #1594 (base changed to `feat/v41-bounded-replay-runtime`; your commit is rebased unchanged, `f698ee3b` adds the glue) so V4.1 gets bounded replay and PD together. What the stack changes for this PR:

* **Drafter rows follow the decoder view.** With CED narrowing the target's taps are each prompt's kept tail plus the decode rows, so `DeepseekV41DSpark.run` now writes one window row per row of `backend.decoder_view()` instead of the raw extend lengths; the kept tail is exactly the 128-row history the first decode reads. The fake-backend test seeds a 5-token extend whose decoder kept 3 rows and checks the dropped rows are never written.
* **The SWA group is replayable, so a hit regenerates the DSpark rows with it.** Nothing to do here beyond the above — the replay forward captures the kept rows and the drafter writes them.
* **Replay covers the whole retention window.** The transfer's `full_suffix` ships every page intersecting the last `sliding_window_tokens − 1` positions; with `protection = 2 × verify width` that is wider than the 128-token attention window, and the first 1P1D run failed with `table 'v41.swa' logical slot 185 has invalid block ID 0` (the page below the replay start was a hole). The recipe (#1594) now declares `replay_window_tokens = sliding_window_tokens` for the SWA and tail groups, the backend validates the equality, and the scheduler (#1593) requires it on the P and D roles. `validate_cache_peer_layout` also compares the field.

Validated on 8×H20, prefill TP4 + decode TP4, Mooncake over RDMA (no forced NVLink), marlin, DSpark on both roles: the three quality checks pass with 39 draft tokens accepted on the decode node; prefix hits on the prefill node replay (`#cached-token: 11892` = 12032 − 140) and answer identically to a fresh prompt, including a hit followed by a 9K-token suffix (two chunks); GSM8K 100 samples 1.00, mean accepted 4.0 on the decode node. `--engram-host-table` fails here with `cudaHostRegister` on this container, so the local run kept the Engram tables on GPU.

PR #1596, opened 2026-09-16
perf(mla): pipeline FP8 prefill on gfx950

qedawkins:
## Summary

Overlap adjacent FP8 MLA tiles with eight-wave workgroups, unscaled K64 matrix instructions, separate K/V buffering and VGPR matrix accumulators. Zero-fill masked loads directly and reduce output relayout work. Keep the 512-workgroup launch fixed across query lengths, placing useful work at low block IDs for underfilled calls and exiting unused blocks before metadata loads. Preserve online softmax, preallocated output strides and the FP16/BF16 path.

## Testing Plan

Extend reference coverage to ragged and prefixed requests, GQA and 12/128 heads. Add FP8 cases for online-maximum changes across buffer wraps, poisoned/empty KV tails, guarded strided outputs through FP64, persistent query cycles and multiple batch groups. Cover graph replay with changing request boundaries and empty requests, a fixed launch configuration, and compiled-kernel reuse across query sizes.

## Benchmarks

Kimi-K3 MXFP4, 8x MI355X, TP8/EP1, FP8 KV, EAGLE3 (3 steps, topk 1, 4 draft tokens), Gluon MLA, 8192-token prefill chunks and decode graphs.

EvalScope streaming completions with identical 50,000-input/500-output prompts, one request per concurrency slot, prefix caching and KVStore off. Two disjoint warmup sweeps, then three measured waves per concurrency. Values are medians of the three request means in milliseconds; all requests are retained.

```text
| C  | Main TTFT | New TTFT | Change | Main TPOT | New TPOT |
|----|-----------|----------|--------|-----------|----------|
|  4 |  13854.12 | 13238.30 | -4.45% |     10.44 |    10.19 |
|  8 |  29283.00 | 28832.46 | -1.54% |     14.31 |    12.68 |
| 16 |  60535.45 | 58326.67 | -3.65% |     19.09 |    18.30 |
```

These are serving metrics from repeated waves on one server process per arm, not isolated decode-kernel timings or independent server restarts.


tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/attention/mla/prefill.py
```diff
@@ -40,6 +40,27 @@
 async_copy = cdna4.async_copy
 
 
+@gluon_builtin
+def _mfma_unscaled_fp8(a, b, acc, *, _semantic):
+    # Keep scales absent in IR. mfma_scaled materializes unit-scale tensors,
+    # preventing LLVM from selecting the unscaled K64 instruction.
+    fmt = "e4m3" if a.dtype == gl.float8e4nv else "e5m2"
+    output = _semantic.dot_scaled(
```
raikonenfnu, line 50: Any thoughts on just using non scaled mfma variant? per `V_MFMA_F32_{*}_FP8_FP8`?

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/attention/mla/prefill.py
```diff
@@ -504,6 +542,319 @@ def compute_tile(
...
+                is_pure=True,
+                pack=1,
+            )
+            updated += (value,)
+        values = updated
+    return values
+
+
+@gluon.jit
+def _fp8_pipeline_phase(
```
raikonenfnu: NIT: Can we get a little bit more intuitive naming for this?

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/attention/mla/prefill.py
```diff
@@ -886,7 +1261,10 @@ def _mla_prefill_kernel(
             cu_seqlens_kv_ptr,
         )
         if active:
-            process_query_block(program, k_smem, k_pe_smem, v_smem)
+            if cfg.IS_FP8:
+                process_query_block_fp8(program, k_smem, k_pe_smem, v_smem)
+            else:
+                process_query_block(program, k_smem, k_pe_smem, v_smem)
```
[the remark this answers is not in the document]
qedawkins: https://github.com/lightseekorg/tokenspeed/blob/26e2b9f7c464f9512050a43eedd777293602b90d/AGENTS.md#design-principles appears to be talking more about runtime/model schedule than individual kernels and how they handle different datatypes. It's expected that different datatypes can have dramatically different kernel paths. The FP8 implementation uses 4 V buffers and overlaps differently than the FP16 case so forced code sharing isn't necessarily a win here.

raikonenfnu (commented):
Overall LGTM! just minor NIT & Q. :)

qedawkins:
Darn, having some accuracy issues switching to the wide `V_MFMA_F32_32x32x64_FP8_FP8` instruction.

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/attention/mla/prefill.py
qedawkins: (according to agent) this is the way to emit the unscaled variant. Comment was very misleading though, good catch, updated.

raikonenfnu (approved):

raikonenfnu:
> Darn, having some accuracy issues switching to the wide `V_MFMA_F32_32x32x64_FP8_FP8` instruction.

OK, we can fix later then :)

qedawkins:
> > Darn, having some accuracy issues switching to the wide `V_MFMA_F32_32x32x64_FP8_FP8` instruction.
> 
> OK, we can fix later then :)

Actually I checked some earlier CI runs and they also fell under the threshold on unrelated changes so might be fine.

PR #1597, opened 2026-09-16
fix(scheduler): publish state checkpoints against the exact computed frontier

FC-Li:
## Summary

Alternative fix for the missed Mamba/SSM state checkpoint reuse reported in #1584, addressing the root cause: the decode admission frontier lagged the real computed endpoint by up to `decode_input_tokens - 1`, so an aligned accepted endpoint was not hashed in the admission that followed it.

`scheduleDecode` and `retractVictim` computed `num_computed_tokens` as `TokenSize() - decode_input_tokens`, a leftover from #447 where the container's last T tokens were the pending inputs. Feedback lands the accepted tokens plus one sampled token, so the exact frontier is `TokenSize() - 1` — which the finish path already used. With speculation on, prefix hashes therefore advanced behind the endpoint, and a state checkpoint written at an aligned endpoint waited for a later admission; if another aligned endpoint landed first, the single materialized-boundary record was overwritten and the earlier checkpoint was never published.

## Changes

- `Request::NumComputedTokens()` is the one frontier for prefix hashing and retention: the scheduled window end while prefilling, every token but the last while decoding, exact under any verify width. `scheduleDecode`, `schedulePrefill` and `retractVictim` use it; the publish-before-`ReclaimExpired` order inside `Admit` is unchanged.
- `CacheProgress::materialized_state_boundaries` keeps every aligned boundary written but not yet hashed. A landed result records its accepted endpoint in `ExtendTokens`; a scheduled prefill records the checkpoint it materializes (local prefill its latest internal aligned boundary, a remote landing its endpoint when aligned). This is what the overlap schedule needs when two results land back to back before the request's next admission; merely crossing a boundary still records nothing.
- Non-closed groups keep one publication routine: the existing lookback path, iterated per resumable boundary. A sliding window publishes its newest hashed boundary; state publishes each recorded boundary inside the newly hashed range. Entries leave the list only after the admission that hashed them succeeds, so a failed attempt retries.
- `Request::MaterializedStateBoundaryTokens()` is removed; evidence lives where it is produced.
- `docs/design/scheduler.md` and `docs/design/cache-concepts.md` describe the exact frontier and the evidence list; the "conservative frontier" wording is gone.

Relation to #1584: the two scenario tests and the list structure are adapted from there (co-authored). Differences: the frontier is fixed rather than compensated for; the coordinator does not grow a state-specific publication loop; the D role's transferred endpoint snapshot still counts as evidence when the prompt ends aligned.

## Validation

CPU scheduler-level only; no model accuracy or GPU performance result is claimed.

- The two scenario tests adapted from #1584 (prefix granularity 128 with verify width 4; granularity 4 with widths 3, 4, 12; live and finished parents; overlap depth 0 and 1; a boundary crossed but never written) fail on `main` for every aligned endpoint and pass with the first commit alone.
- A new scenario test plans two decode steps back to back under overlap depth 1, lands both results, and checks that both aligned endpoints resume a later prompt; it fails with the frontier fix alone and passes with the second commit.
- A `CacheProgress` unit test covers landing-time recording and deduplication, a failed admission leaving the list intact for retry, and state pages smaller than the prefix granularity publishing only the interval's endpoint slot.
- 486 C++ scheduler tests pass; 153 Python scheduler-binding tests pass against the rebuilt extension; `pre-commit run --all-files` clean.

One behavioural consequence to watch in CI: with the exact frontier, sliding-window retention reclaims up to `decode_input_tokens - 1` tokens earlier than before. `ExpiredBlocksAt` keeps positions `>= C - W + 1` and `SwaMatcher::pagesNeededToResume` needs the window's last `W - 1` tokens, so the two agree and the non-speculative case has always run at zero slack, but the SWA + MTP model tests are the ones that exercise it end to end.


byshiue (approved):
LGTM

PR #1598, opened 2026-09-16
perf(amd): widen the SiTU warp-decode bound to cover speculative decode widths

nirvedhmeshram:
## Summary

Kimi-K3 decode was running its MoE on the prefill kernels. Two commits: the
first removes a dispatch failure that capped how far the bound could move, the
second moves it.

**1. `_dynamic_fp8_quantize` reduced the per-tensor amax in a single
workgroup**, materialising `next_power_of_2(numel)` elements in one program. It
degrades sharply with size and eventually cannot be dispatched. Measured on
gfx950 with the device idle:

| padded elements | one workgroup | grid reduction |
|---|---|---|
| 16,384 | **14.0 us** | 25.8 us |
| 65,536 | **13.6 us** | 26.4 us |
| 131,072 | 167.8 us | **26.7 us** |
| 262,144 | 1156.3 us | **26.6 us** |
| 458,752 | `HSA_STATUS_ERROR_OUT_OF_RESOURCES` | **25.6 us** |

Reduce over a grid above the size where one workgroup stops paying off. The
combine folds into the apply pass, so the grid form costs two launches rather
than three -- at these sizes the launch, not the arithmetic, dominates. `N`
becomes a runtime argument rather than a `constexpr`, so the JIT no longer
recompiles per distinct token count.

**2. `gluon_mxfp4_fp8_precomputed_situ` sent anything wider than 16 rows to
package prefill.** That bound predates speculative decoding, when a decode step
was one row per request. With EAGLE3 the decode width is
`concurrency x num_draft_tokens`, so every concurrency above 4 landed on the
prefill kernels.

Those kernels tile routed rows into `BLOCK_M=128` groups, which pays off once a
batch fills the tiles. Speculative decode does not fill them: a batch of 8 is 32
rows spread over hundreds of experts, so the tiles are mostly padding. The MoE
block costs ~246 us per layer that way, against ~34 us for the warp-decode
kernels:

```
  92x  18.9 us  _warp_decode_precomputed_situ_stage1_kernel
  92x   9.7 us  _warp_decode_stage2_fp8_mxfp4_kernel
  92x   4.9 us  _moe_partial_reduce
```

## Measurements

End to end, 8x gfx950, TP8/EP1, EAGLE3 (3 steps, 4 draft tokens), 50,000 in /
500 out, one warm-up then one measured pass per point, device otherwise idle.
TPOT in ms, and output token throughput in tok/s, from the same runs:

| | concurrency 8 (32 rows) | concurrency 16 (64 rows) |
|---|---|---|
| TPOT before | 12.49, 12.56 | 18.29 |
| TPOT after | **10.57, 10.34, 10.33** | **17.45, 17.64** |
| | **-17%** | **-4.0%** |
| throughput before | 102.00, 99.95 | 107.25 |
| throughput after | **104.79, 104.36, 104.57** | **107.84, 107.35** |
| | **+3.6%** | **+0.3%** |

Baseline spread at concurrency 8 is 0.6% on TPOT, so that move is roughly 25x
the run-to-run noise.

**Throughput moves far less than TPOT, and that is expected at this workload
rather than a sign the TPOT figure is soft.** With 50,000 input and 500 output
tokens, time to first token is 83% of wall time at concurrency 8 and 87% at
concurrency 16, so output throughput -- total tokens over total wall time -- is
mostly measuring prefill. The two are consistent: predicting benchmark duration
from the TPOT delta alone (`TPOT x 499` per request) gives 38.14 s against 38.25
measured at concurrency 8, and 74.17 s against 74.18 at concurrency 16. The
decode improvement is fully accounted for; it is simply a small share of this
benchmark's wall clock.

TTFT is unchanged, as it should be for a decode-path change: 32.5 s before and
after at concurrency 8, 64.9 s at concurrency 16.

**Controls.** With the bound at 32 rather than 64, concurrency 16 does not cross
it and does not move (TPOT 18.29 -> 18.32, throughput 107.25 -> 107.69). Only a
concurrency whose width crosses the bound changes, and only when it crosses.
Concurrency 4 (16 rows) is below the bound in both configurations and takes an
identical code path by construction; measured once as an invariance check at
7.20 ms / 99.65 tok/s.

Concurrency 16 could not be measured at all before the first commit: the
one-workgroup quantizer failed to dispatch at that width, on every rank.

## Caveats

* **Speculative acceptance differed between arms** (3.556 / 3.710 / 3.632).
  TPOT is per output token, so higher acceptance flatters it. Removing that
  denominator, the concurrency-8 gain is about -13% rather than -17%, and
  concurrency 16 about -2%. Both still well outside noise, but the headline
  figure is not purely a speed effect. The acceptance numbers are per-server
  aggregates over mixed runs, so the correction is approximate.
* **One measured pass per point**, after a warm-up. The concurrency-8 result is
  far outside the baseline spread; the -4.0% at concurrency 16 is smaller and
  would benefit from repeats.
* **The warp-decode kernel documents itself as "Small-M (M<=4)"** and its own
  wrapper enforces `WARP_DECODE_MAX_M = 4`. The SiTU entry point already called
  it directly at up to 16 rows, bypassing that wrapper; this raises it to 64.
  The measurements say it is correct and faster there, and the new tests check
  numerics against the reference at 17, 32 and 64 rows -- but the kernel's
  stated contract has not been updated to match how it is used, and somebody who
  owns that kernel should say whether 64 is a bound it wants to stand behind.

## Tests

* Both package-prefill tests picked token counts one row past the old bound
  (17 and 33), so raising it would have silently stopped them testing package
  prefill. They now derive their token count from the named constant.
* New reference-comparison coverage at 17, 32 and 64 rows on the warp-decode
  side. Nothing covered those widths before -- the existing decode tests stop at
  16, exactly where the bound used to be. They also assert the routing, since a
  silent fall back to package prefill would still produce correct numbers.
* New quantizer coverage across both strategies, including every width that
  previously aborted the dispatch, plus a host-side test pinning the strategy
  choice itself.

`pre-commit run --all-files` clean. AMD MoE suite: 216 passed, 15 skipped.
Kernel selection suite: 272 passed, 61 skipped.

## Note on splitting

The two commits are independent in review terms even though the second needs the
first to be measurable at concurrency 16. Happy to split them if that is easier
to review -- the quantizer change stands on its own as a crash fix.


raikonenfnu (approved):

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx950/moe/mxfp4/fused/moe.py
```diff
@@ -137,6 +137,20 @@ def _swiglu_activation(alpha: float, limit: float, beta: float) -> FusedActivati
 _ROUTE_OWNED_DECODE_MAX_M = 2
 
 
+# Widest activation the precomputed-SiTU entry point serves with the
+# warp-decode kernels; anything wider goes to package prefill.
+#
+# Package prefill tiles routed rows into BLOCK_M=128 groups, which pays off once
```
qedawkins, line 143: Did you check different options for BLOCK_M? At M = 128 with BLOCK_M=128 this is still mostly padding in every case (because of top-k). Should we reduce the block size?

nirvedhmeshram: taking a look, will send a follow up with what I find.

Edit: AI response below.

nirvedhmeshram: Good question, and no I hadn't — so I swept it. BLOCK_M turns out not to be the
lever, on the K3 shape (896 routed experts, top-k 16, hidden 3584, intermediate
384), median us for the whole SiTU package-prefill call:

| M | 32 | 64 | 128 |
|---|---|---|---|
| 65 | 621.1 | 619.4 | 618.9 |
| 128 | 644.1 | 643.2 | 643.3 |
| 256 | 633.3 | 644.8 | 646.0 |
| 1024 | 693.1 | 694.3 | 694.9 |
| 2048 | 1012.3 | 1011.6 | 1010.7 |
| 8192 | 2102.1 | 2102.9 | 2101.7 |

Under 2% everywhere, mostly 0%, and identical numerics across the three.

You're right that the tiles are nearly all padding -- at M=128 it is 114,688
padded tile rows for 2,048 real ones. It just isn't what costs. 126x the rows
(65 -> 8192) buys only 3.4x the time, on a floor of ~620 us, and that floor is
what it takes to stream all 896 experts' weights once: 1.85 GB, ~617 us at
3 TB/s effective against 619 us measured at M=65. At low occupancy the kernel is
weight-bandwidth-bound, so shrinking the block shrinks padding that was already
free.

Which also reframes the bound this PR moved: warp-decode does not win there by
tiling better, it wins by not touching every expert's weights. So the lever at
low density looks like expert weight traffic rather than block shape -- a bigger
piece of work than a selector tweak, and I don't have a follow-up PR to point at
from this.

Two caveats on the numbers: the harness routes each token to a different set of
experts, which maximises how many are touched, so the absolute times are a
worst-case spread rather than the model's real MoE cost -- all three arms share
inputs, so the comparison holds even though the magnitudes do not. And note
BLOCK_M=32 only exists on the E4M3 activation path; stage1 rejects it for E2M1.

PR #1601, opened 2026-09-16
[Speculative Decoding] add synthetic acceptance length

edwingao28:
## Summary

Add benchmark-only synthetic acceptance length via CLI/JSON across five samplers, preserving draft prefixes, target tokens and bootstrap limits. Forward passes remain enabled; lengths use floor/ceil sampling. Generated text is unsuitable for accuracy evaluation.

## Test Plan

At `e9778b3` (base `d1464d8`): pre-commit passed; H100 suite: 107 tests + 52 subtests passed, no skips. Qwen3.5-0.8B BF16/native MTP, width 4; 288/288 requests, 128 reported output tokens each, concurrency 1/4/8.

| TP | Requested AL | Eager mean | Graph mean |
|---|---|---|---|
| 1 | off | 2.204028 | 2.219733 |
| 1 | 1 | 1.000000 | 1.000000 |
| 1 | 2.6 | 2.608479 | 2.605853 |
| 1 | 4 | 4.000000 | 4.000000 |
| 2 | 2.6 | — | 2.607098 |

AL statistics include warmup. Post-request graph execution confirmed, including both TP2 ranks. Stochastic/penalty/logprob/DP/AMD/other-drafter combinations remain unqualified. No speedup claim.


Dogacel (changes requested):
Instead of gating `verify...` methods, we should modify the verify functions directly. We can override the accepted indices inside those functions *after the calculation* to make sure we run the actual drafting as well.

For simulation we can probably do gaussian instead of ceil / floor, I would like to learn what is the distribution for accepted tokens in real inference for reference.

edwingao28:
> Instead of gating `verify...` methods, we should modify the verify functions directly. We can override the accepted indices inside those functions _after the calculation_ to make sure we run the actual drafting as well.
> 
> For simulation we can probably do gaussian instead of ceil / floor, I would like to learn what is the distribution for accepted tokens in real inference for reference.

thanks for the review @Dogacel! The draft and target model forward passes still run in this implementation; the synthetic path only replaces the acceptance decision inside the sampler’s verify(). vLLM and ATOM use a similar approach, while SGLang and TRT-LLM compute verification results before overriding them. If your concern is preserving the original verification kernel’s overhead, I’m happy to discuss that separately.
For the distribution, floor/ceil is intentional: it matches the synthetic acceptance modes used by other engines inside our repo(https://github.com/SemiAnalysisAI/InferenceX/blob/main/golden_al_distribution/README.md). Switching only TokenSpeed to Gaussian would change the distribution even at the same mean AL. Investigating real acceptance distributions sounds useful, but we’d keep this mode aligned with the existing cross-engine benchmark methodology.

edwingao28:
just resolved file conflicts and updated our h100 validation, could you take a look @Dogacel thanks!

Dogacel:
> > Instead of gating `verify...` methods, we should modify the verify functions directly. We can override the accepted indices inside those functions _after the calculation_ to make sure we run the actual drafting as well.
> > For simulation we can probably do gaussian instead of ceil / floor, I would like to learn what is the distribution for accepted tokens in real inference for reference.
> 
> thanks for the review @Dogacel! The draft and target model forward passes still run in this implementation; the synthetic path only replaces the acceptance decision inside the sampler’s verify(). vLLM and ATOM use a similar approach, while SGLang and TRT-LLM compute verification results before overriding them. If your concern is preserving the original verification kernel’s overhead, I’m happy to discuss that separately. For the distribution, floor/ceil is intentional: it matches the synthetic acceptance modes used by other engines inside our repo(https://github.com/SemiAnalysisAI/InferenceX/blob/main/golden_al_distribution/README.md). Switching only TokenSpeed to Gaussian would change the distribution even at the same mean AL. Investigating real acceptance distributions sounds useful, but we’d keep this mode aligned with the existing cross-engine benchmark methodology.

Yes, I favor preserving verification kernel overheads, this could help us plan ahead what speculator to train as well. How hard would it be to update this PR to do that instead of skipping drafting?

Matching distribution, I think doing empirical data is fine. But I expect DFlash's acceptance to be bimodal, if first token is accepted, it is more likely for next tokens to be accepted. I saw InferenceX only measured MTP and EAGLE, which are same AR family. Though I'm fine with merging this existing implementation and later extend the format together.

edwingao28:
@Dogacel thanks just updated all five backends now run normal verification before overriding the acceptance result, preserving the verification kernel overhead

For Dflash distribution I would kept floor/ceil for this PR but happy to explore in a followup!

PR #1605, opened 2026-09-17
fix(runtime): preserve pending drains on empty memory resume

yechank-nvidia:
## Summary

Keep new requests paused when an empty memory-resume request arrives while memory release is still waiting.

## Problem

Before releasing memory, the scheduler stops accepting new requests and waits until it can clear the cache.

During that wait, `resume_memory_occupation` may have no memory regions to resume. The current code still lets new requests in, even though the memory release has not finished. Both omitted tags and an empty list can trigger this.

## Solution

Add a four-line check in the existing `MemoryOccupationController`. If there are no regions to resume and a drain is pending, return success and keep the scheduler paused. The operation already waiting for the drain will decide when to accept new requests again.

Add both cases to the existing memory-release test and document this rule.

## Test Plan

Run from the repository root with the runtime dependencies installed:

```bash
  test/runtime/test_memory_occupation.py \
  test/runtime/test_event_loop_pause_hooks.py
pre-commit run --all-files
```

- Before: both new test cases fail because the scheduler accepts new requests while memory release is still waiting.
- After: all 11 tests pass. The resume request succeeds, new requests stay paused, and memory release finishes once the cache can be cleared.
- All pre-commit checks passed. These are CPU tests; full serving was not tested.


yechank-nvidia:
@codex review

PR #1606, opened 2026-09-17
perf(moe): pass unpacked routes to FlashInfer

yechank-nvidia:
## Summary

Pass expert IDs and routing weights directly to FlashInfer. At batch 1, routed MoE call time drops from 25.22 to 20.59 us, an 18.4% reduction.

## Problem

The wrapper packs each expert ID and BF16 routing weight into one integer before calling FlashInfer. The casts and bit operations create temporary tensors and add GPU work to every call.

The pinned FlashInfer 0.6.18 API already accepts separate IDs and weights, so this packing is unnecessary.

## Solution

Pass a tuple of contiguous `(int32 expert IDs, BF16 routing weights)` to the same FlashInfer API. Keep the BF16 conversion so the weights round the same way as before.

Reuse the existing weight preparation and output handling. Extend the existing test to compare packed and unpacked inputs, including the path that runs finalization separately.

## Performance

Measured on NVIDIA GB200 with BF16 inputs and expert weights: 16 experts, top-4 routing, hidden size 4096, intermediate size 1024, and finalization enabled. Routing weights come from a strided FP32 tensor.

Both versions use the same FlashInfer call. The old version builds packed score/ID integers; the new version passes IDs and weights separately.

After warmup, measure old/new/new/old. Each CUDA graph contains 32 calls. For each measurement, take the median of 15 graph runs, then take the median of the two measurements for each version. Times below are per call.

| Batch | Packed (us) | Unpacked (us) | Latency reduction | Speedup |
| ---: | ---: | ---: | ---: | ---: |
| 1 | 25.22 | 20.59 | 18.4% | 1.22x |
| 8 | 71.35 | 66.60 | 6.7% | 1.07x |
| 33 | 79.56 | 72.30 | 9.1% | 1.10x |

At batch 8, the old version took 70.22 and 72.48 us in the two measurements. The new version took about 66.60 us in both. Outputs matched exactly at every measured batch size.

These numbers measure one routed MoE call. Full serving performance was not measured.

## Test Plan

Build the matching native kernel extensions, then run:

```bash
  tokenspeed-kernel/test/nvidia/thirdparty/test_moe_finalize_fuse_shared.py
  tokenspeed-kernel/test/test_kernel_api_selection.py \
  -k 'moe and (unquant or precomputed)'
pre-commit run --all-files
```

- 23 native tests pass on GB200. They check exact results for packed and unpacked inputs, strided routing weights, separate finalization, PDL on/off, and graph replay after changing weights.
- Eight kernel-selection tests pass; five are skipped on this platform.
- All pre-commit checks passed.


yechank-nvidia:
@codex review

PR #1607, opened 2026-09-17
perf(sampling): skip unused fallback-index reduction in target-only v…

yechank-nvidia:
## Summary

Skip an index calculation that target-only chain verification does not use. At batch 1, chain length 1 and vocabulary 131072, kernel time drops from 47.44 to 39.77 us, a 16.2% reduction.

## Problem

The shared sampling helper finds `last_valid_id` for every vocabulary chunk. Target-only chain verification uses `sampled_id` and never reads `last_valid_id`, so that work is wasted. Larger vocabularies need more chunks and pay more of this cost.

Other callers still use `last_valid_id` as a fallback, so they need to keep the calculation.

## Solution

Add one compile-time flag to the existing helper. Turn it off for target-only verification and keep it on for the other six callers.

The helper keeps the same probability sums, CDF-based token selection, synchronization and fallback token. Add a case with sparse probabilities to the existing tests to check selection at a CDF boundary.

## Performance

Measured on NVIDIA GB200 with FP32 probabilities, deterministic mode on, PDL off, no draft probabilities, and `threshold_single=threshold_acc=1.0`.

Build the old and new native libraries separately. Use the same inputs and output buffers for both. Warm up both graphs, then collect six measurements per version, switching the order between old/new and new/old.

Each CUDA graph contains 256 calls. For each measurement, take the median of 31 graph runs. The table shows the median of the six measurements for each version, as time per call.

| Batch | Chain length | Vocabulary | Before (us) | After (us) | Latency reduction | Speedup |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 1 | 4096 | 3.91 | 3.63 | 7.1% | 1.08x |
| 1 | 1 | 131072 | 47.44 | 39.77 | 16.2% | 1.19x |
| 8 | 1 | 131072 | 44.92 | 37.78 | 15.9% | 1.19x |
| 64 | 1 | 131072 | 49.96 | 41.40 | 17.1% | 1.21x |
| 8 | 4 | 131072 | 52.59 | 43.61 | 17.1% | 1.21x |

These numbers measure the native sampling kernel. Full serving latency and other GPU architectures were not measured.

## Test Plan

Rebuild the native kernel package from this checkout, then run:

```bash
  tokenspeed-kernel/test/nvidia/thirdparty/test_cuda.py \
  -k 'TestChainSpeculativeSamplingTargetOnly or TestVerifyChainGreedy'
pre-commit run --all-files
```

- 26 tests pass, including the sparse-probability CDF-boundary case with deterministic mode both on and off.
- Old and new outputs match exactly in 96 comparisons. Values written back to the draft buffer also match. Cases cover batch sizes 1/3/16, chain lengths 1/4, vocabularies 4096/131072, with and without draft probabilities, both deterministic modes, and PDL on/off.
- All pre-commit checks passed.


yechank-nvidia:
@codex review

PR #1608, opened 2026-09-17
fix(runtime): aggregate cache flush results across data parallel schedulers

yechank-nvidia:
## Summary

Return success from `flush_cache()` only when all data-parallel schedulers clear their caches.

## Problem

The code waits for all schedulers to reply but returns only the first result. For example, `[success, failure]` returns success even though one scheduler failed. Changing the reply order changes the result.

`AsyncLLM` has its own copy of this method with the same bug. Fixing only the base method would leave that copy unchanged.

## Solution

Use `all(result.success ...)` in `SchedulerControlClient.flush_cache()` to check every reply. Remove the duplicate method in `AsyncLLM` so it uses the existing base method.

Keep the same communicator, which already waits for every scheduler. This fix does not depend on the deferred-flush or shutdown changes.

## Test Plan

Run from the repository root with the runtime dependencies installed:

```bash
  test/runtime/test_scheduler_communicator.py \
  test/runtime/test_async_llm_protocol.py
pre-commit run --all-files
```

- Test `AsyncLLM.flush_cache()` with the existing queue-based communicator. Cover one scheduler, all-success replies, and mixed replies in both arrival orders. Also check that the call waits for the last reply.
- Before: `[success, failure]` returns success and fails the new test.
- After: 5 tests and 4 subtests pass. Mixed replies return failure in either order. All-success replies return success.
- All pre-commit checks passed. These tests run on the CPU.


yechank-nvidia:
@codex review

PR #1618, opened 2026-09-17
refactor(cache): own the pool-sizing capacity model in the C++ scheduler

FC-Li:
## Summary

The per-request cache working set — the decode reservation, the overlap-protected step (`protected_tokens`), a snapshot-state group's checkpoints and banked growth, a sliding group's lookback and resident window — was defined twice: in `Scheduler::singleRequestLcmBlocksRequired` for the C++ startup bound, and again in the Python recipes (`compute_cache_group_page_counts` plus hand-written overrides in the Kimi-K3 and GLM-5.3-Flash recipes) that size the pool before it exists. The two copies agreed only by convention, and drift surfaced as a silently clipped `max_model_len` or a mis-sized pool rather than a failure.

This PR moves the model into one config-only C++ class, `CapacityModel`, exports it through the scheduler binding, and makes the recipes call it. Neither side restates the other's formula anymore.

## Design

**`csrc/scheduler/capacity_model.h`** — built from a `SchedulerConfig` alone; reads no `total_pages`, validates the sizing subset via the new `SchedulerConfig::ValidateCapacityInputs()`, and answers both questions from the same helpers:

- `SingleRequestGroupPages(L)` / `MaxSingleRequestTokens` — the startup bound, moved verbatim; the `Scheduler` now calls it.
- `ConcurrentGroupPages(T, ctx)` — per-group demand for `max_batch_size` live requests, which the Python recipes previously computed on their own.
- `LcmBlocksNeededFor` — moved off the coordinator (it only read packing).

`MakePrefixMatcher(spec)` is factored out of `MakeCoordinator` so the model asks the same matcher the coordinator builds for prefix-closedness and lookback. `SchedulerConfig::Validate()` / `CacheGroupConfig::Validate()` each become "capacity inputs + page-count checks", so no rule is stated twice.

**`recipes/scheduler_bridge.py`** — the recipes' side of the scheduler contract: the one spec → `CacheGroupConfig` conversion (`engine/scheduler_utils.pool_to_cache_groups` now delegates to it), `SchedulerLimits`, and `capacity_model()`. `CacheRecipe.parents_needed` builds an unsized config, asks the model, and folds by packing. `scheduler_limits` now also carries the scheduler role and the prefix-cache flag, so P/D deployments size with the inputs the `Scheduler` bounds with.

Deleted: `compute_cache_group_page_counts`, `NULL_PAGES`, the K3 `parents_needed` override, `glm53_flash_parents_needed`.

## The invariant that ties the two callers together

For one live request of `L` tokens, `ConcurrentGroupPages(L, L)` is never below `SingleRequestGroupPages(L)` in any group — a pool sized for the configured concurrency admits every request the bound accepts. The C++ suite sweeps roles × chunk sizes × decode widths × overlap depth × prefix-cache state over four group sets and checks it for every `L`; the binding test sizes a pool from the model, constructs a `Scheduler` on it and asserts `max_single_request_tokens() >= L`.

## Sizing changes (deliberate, all toward what the scheduler actually admits against)

- **Kimi-K3: unchanged** (98 parents at 128K / +9 with a full-block overlapped verify width). Its override already equalled the C++ state peak.
- **GLM-5.3-Flash linear-attention state groups** drop from `N + ceil(chunk / 64) + protected` pages to `N × 4`. The old term counted a full chunk of state slots the scheduler never allocates; KV capacity goes up accordingly. Worth a look from whoever owns that recipe.
- **Base-recipe state groups (Inkling, Qwen4-Exp)** go from `2N` to the C++ peak (four blocks per request for an unaligned finishing chunk) — the shortfall `docs/design/scheduler.md` already described.
- **Full-history and sliding groups** round the tail and the protected tokens together (`ceil((g - 1 + protected) / g)` per request) and the sliding term adds its lookback; typically ±1 page per request.

Tests that pinned the deleted formula now derive their expectations from the model, explained in place. The DCP contract test compares per-parent packing instead of raw page counts: the parent count can differ by the capacity search's rounding without any geometry difference.

## Compatibility

- `tokenspeed-scheduler` bumps to **0.1.17** for the `CapacityModel` export; the runtime pins `>= 0.1.17`. The runtime CI needs the new wheel published first.
- `scheduler_limits` reads `server_args.disaggregation_mode` and `server_args.enable_prefix_caching`; test fixtures that fake `server_args` for K3 / V4 / V4.1 / GLM recipes gained those two fields.

## Verification

- C++: 516 tests pass (5 new in `test_capacity_model.cpp`).
- Scheduler binding: 182 tests pass (26 new in `test_capacity_model.py`).
- Runtime: the recipe / sizing suites (`test_kimi_k3_cache_spec`, `test_glm53_flash_cache_spec`, `test_multi_window_page_counts`, `test_v4_sliding_window_groups_smoke`, `test_unified_kv_slab_pool`, `test_dcp_cache_contract`, `test_dsv4_cache_arena`, `test_kimi_k3_integration`, `test_cache_setup`, `test_qwen4_exp` sizing tests) pass. `test_v4_pd_recipe_and_readiness_follow_cache_producers` fails identically on `main` (fixture lacks `dcp_size`), unrelated.
- `docs/design/scheduler.md` §1.4 and `docs/design/cache-concepts.md` record the shared model and the invariant.


FC-Li:
**CI triage after the first round**

`model-test / eval-glm-5.3-flash-fp8-mtp-aime26-amd` failed, and it is caused by this PR — the one place the sizing change is user-visible. Same job on `main`:

| | `main` | this PR |
|---|---|---|
| `Cache profile` | `parents=596`, arena 8.7 GB | `parents=308`, arena 4.5 GB |
| reported `token_capacity` | 131072 | 131072 |
| `retract … for cache capacity` | 0 | 224 |
| outcome | pass | evalscope timeouts, job cancelled |

Why: the GLM-5.3-Flash recipe sizes `num_lcm_blocks = min(budgeted, parents_needed(--max-total-tokens))`. The old linear-attention state formula charged one state slot per 64-token slice of the prefill chunk (3 groups × 128 pages = 384 parents) that the scheduler never allocates — and because LCM parents are shared across groups, full-attention history consumed them at runtime. So `--max-total-tokens 131072` was effectively ~460K tokens of history, and the eval (16 concurrent requests × 27K–31K-token answers) depended on that. The capacity model honours the limit exactly (2080 history pages ≈ 4 requests at a time) and the remaining requests thrash.

The state demand itself is right (four blocks per live request, what the scheduler admits against); the eval config was stating a budget it never actually ran within. `f0e4e6a7` raises both GLM-5.3-Flash aime26 evals to `--max-total-tokens 524288` (16 × 32K, ~9.5 GB of arena). The NVIDIA variant had not reached the eval yet and would have failed the same way.

The other three failures are not related to the change:
- `eval-deepseek-v4.1-flash-gsm8k-amd`: runner out of disk while downloading weights (`No space left on device` in `hf_hub_download`); server never started.
- `eval-deepseek-v4.1-flash-gsm8k` (b200): `apt-get update` failed on the CUDA apt mirror mid-sync during dependency install.
- `eval-kimi-k3-nvfp4-tp8-two-node-aime26-gb300-slurm`: score 0.867 vs threshold 0.9 on 30 AIME problems; the last four `main` runs scored 0.9 / 0.967 / 0.967 / 0.9. K3 sizes its pool from the byte budget (not `parents_needed`), so its pool and admission are identical to `main` — one-problem noise.

PR #1620, opened 2026-09-17
feat(amd): add gfx1250 DSV4 Gluon decode

antiagainst:
Implement fused page-planar FP8 dequantization, BF16 WMMA attention, online softmax, split reduction, padded LDS, TDM query staging, and double-buffered async cache staging. Select the async path by partition geometry and preserve a register-prefetch fallback, with Proton FLOP and byte metadata.

Benchmark Flash and Pro shapes with 64/128 heads across batches 1/16/64/256, 8K-32K inputs, 1K-8K outputs, and CSA/HCA layers. All 192 cases matched the portable reference; geometric-mean speedup was 6.99x cold and 8.26x reused.


tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/README.md
borontion, line 1: actually what's the purpose of this file? should be the "Contract" specified in the API docstring or somewhere?

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx1250/attention/dsv4/decode.py
```diff
@@ -0,0 +1,1125 @@
...
+            partial_out.stride(2),
+            partial_lse.stride(0),
+            partial_lse.stride(1),
+            partial_lse.stride(2),
+            output.stride(0),
+            output.stride(1),
+            NUM_KV_SPLITS=num_splits,
+            HEAD_DIM=512,
+            num_warps=1,
+            waves_per_eu=1,
```
borontion: is this `waves_per_eu` something still needed?

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx1250/attention/dsv4/decode.py
```diff
@@ -0,0 +1,1125 @@
...
+        extra_kv_cache,
+        extra_slots,
+        extra_lens,
+        extra_page_size,
+        out,
+    )
+    tokens = q.shape[0]
+    swa_width = swa_slots.numel() // tokens
+    extra_width = extra_slots.numel() // tokens if extra_slots is not None else 0
+    if q.shape[1] < 1:
```
borontion: nit: this check is not needed

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx1250/attention/dsv4/decode.py
antiagainst: Good catch. I was copied over from gfx950 version I believe. It's hurting performance. So dropped.

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/README.md
antiagainst: Good question. I was actually debating a bit about this. Right now leaning towards having this given it is more a quicker guidance for humans to understand the contract and algorithm of those AMD kernels, consider that the AMD package is on its own and registration (which carries the API contracts) happens in the tokenspeed-kernel/ side.

borontion (approved):

PR #1622, opened 2026-09-17
perf(amd): fuse gfx1250 MXFP4 MoE operand loads with partial TDM

jerryyin:
Add a partial_tdm option to the gfx1250 MXFP4 MoE kernels. It assigns each TDM descriptor load a disjoint warp bitmask so a pair of operand loads can issue as one fused TDM operation instead of two.

 - The gather path cannot participate: async_gather takes no warp hint, so its x load keeps every warp and the fusion moves onto w + w_scale. 
 - The scatter path fuses x + w. Both SiTU projections therefore drop one outstanding TDM operation per batch, which NUM_LOADS_IN_BATCH accounts for so async_wait stays in step.

Measured under the functional model at num_tokens=65, num_warps=4: tensor_load falls from 9 to 6 in each projection, and the SiTU output is bit-identical to the whole-warp control. Runtime is not measured, so the option defaults off at every call site.

Credit to original author: @knwng


borontion (dismissed):

tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx1250/moe/mxfp4/_common.py
```diff
@@ -52,6 +52,28 @@ def get_scaled_dot_format_string(dtype: gl.dtype):
...
 
 
+@gluon.constexpr_function
+def partial_tdm_warp_hint(num_warps: int) -> int:
+    # Alternating bits so the mask and its reversal stay disjoint, which is what
+    # lets two loads merge into one `async_load_fused` operation.
+    if num_warps == 8:
+        return 0b01010101
+    if num_warps == 4:
+        return 0b0101
```
knwng: Actually according to my recent experiment, `0b11110000` and `0b1100` shows better TDM latency regarding warp balance.

jerryyin: Interesting, I'd imagine this doesn't matter. Do you know why? Something like cache locality?

borontion (approved):

PR #1626, opened 2026-09-17
refactor(kernel): unify PyTorch references under numerics/reference

antiagainst:
Register every PyTorch reference as the "reference" solution and move the ones that lived under ops/ (decode_gemv, MoE routing, AttnRes) into numerics/reference/, replacing the ad-hoc "torch" solution. The dispatchers that fell back to "torch" now request "reference" through the same select_kernel path; moe_softmax_topk drops its private shortcut.

Make the REFERENCE priority band ground truth only: select_kernel and explain_selection skip it unless the caller passes solution="reference" or override=, so a missing real kernel raises NoKernelFoundError instead of silently running a quadratic torch path. PORTABLE-band references keep their last-resort fallback role.

Consolidate the hand-rolled test references for MHA/MLA/GDN attention, unquantized MoE apply and mHC pre-mapping into registered reference kernels and have the tests request solution="reference" through the public APIs. Add self-checks for the new references and selection-rule tests.


borontion:
I was wondering should we better to switch to use `ops/<family>/torch.py` or `ops/<family>/reference.py` for them. the reasons are

1) the current `ops/` tree is becoming a lot more complex, especially for different attention and moe variants, and it will very likely keep growing after new model support. if we put reference torch implementation under `numerics/reference`, we may either mess it up or need to duplicate the kernel organization there
2) another thing is for low precision kernels involving fp4, it is even tricky to implement a torch reference. instead, using triton can be much easier. so in certain case we need to compare against a fallback triton implementation. if so I think it is better to organize these triton kernels under `ops/` to follow what we are doing now.

PR #1627, opened 2026-09-17
perf(kernel-amd): drop gl.barrier() calls the compiler already emits

antiagainst:
The Gluon compiler's membar analysis tracks every LDS read, write, atomic, async copy, and scratch-backed op (layout conversions, reductions, atomic result broadcasts) and inserts a CTA barrier right before the first conflicting access, including across loop back-edges. It also emits a barrier right after every async_copy.wait_group, and the lowering of a release/acq_rel atomic emits one before it (an acquire atomic emits one after). Manual gl.barrier() calls next to any of these are duplicate s_barriers, or land earlier than the compiler's minimal placement and pin the schedule.

This removes 48 of the 66 manual gl.barrier() sites in the AMD Gluon kernels. Every removal was verified by compiling each affected kernel for gfx950 and gfx1250 with tokenspeed-triton 3.8.10.post20260906 (the version the package pins) and diffing the amdgcn, with each s_barrier attributed to its source line via debug locations. Compilation is deterministic (284/284 kernels byte-identical across two baseline builds), so the diffs are exact. For every removed site the compiler either re-inserts an equivalent barrier at the first consuming op, or an existing barrier already covers the hazard:

```
  kernel (arch)                        s_barrier before -> after
  KDA fused decode (950, 1250)          4 -> 3
  DSv4 decode split stage (950)         6 -> 6 static, 4 -> 3 per tile iter
  DSA top-k persistent (950)           48 -> 43
  DSA top-k oneblock (950)          30-32 -> 25-27
  DSA top-k wave32 (1250)              22 -> 18
  MoE stage1 e4m3, non-async LDS path  14 -> 7
  MoE stage2 1x2, non-async LDS path   12 -> 5
  iris two-stage reduce                 9 -> 8 (-1 per tile iteration)
  iris push / stage attnres             9 -> 7 each
  KDA prefill, MLA decode, DSA dense attention, MoE async paths:
                                        same count, barrier relocated to
                                        the consumer
```

VGPR count, occupancy, and scratch are unchanged in every kernel; the remaining text diffs are instruction reordering around the moved barrier.

Notable cases:
- KDA fused decode: the barrier before output_shared.load was already covered by the CTA barrier inside the gl.sum reduction immediately before it.
- DSv4 decode: the loop-end barrier was covered by the barrier the compiler places before the convert_layout at the top of the next iteration.
- DSA top-k: the compiler already emits one barrier per tile iteration before the LDS atomic_scatter_add, which subsumes the store-then-loop barriers; the pass-boundary barriers move to the histogram load.
- MoE non-async e4m3 path: the wait_group calls still execute, and the compiler's post-wait barriers cover the helper's read-after-write; it keeps exactly one write-after-read barrier per loop.
- iris: the removed barriers sit directly before a release atomic_xchg or after an acquire atomic_add and were literal back-to-back s_barrier pairs in the ISA; the two-stage in-loop barrier is covered by membar for LDS and by the release atomic's barrier after the loop.

The 18 kept sites are the ones the compiler cannot see:
- Global-memory ordering across threads of a workgroup: DSA top-k arrival and pass_done protocol (sparse_mla 446/467) and the IS_DECODE re-read of `out` that other threads just wrote; the gfx1250 small-M route zero-init before atomics; moe_align_fused init stores before the overlapping scatter; the lamport sentinel stores before the epoch publish (LLVM already drops that one for the single-wave kernel).
- load_shared_relaxed pipelines (MoE stage1 slot reuse, stage2 slot reuse, and all nine sites in the fused pipelined program): that load opts out of the compiler's async-copy hazard tracking, so the write-after-read barrier before the buffer_load_to_shared that reuses the slot must stay explicit. Removing them in an experiment made the barriers vanish with no replacement.

The kept sites without a rationale now carry one, and the ops README documents the barrier contract.

Follow-up for triton: membar treats local_atomic_scatter_rmw against itself on the same buffer as a hazard, so every histogram tile iteration in the top-k kernels still carries an s_barrier before the LDS atomic. A membar filter for atomic-vs-atomic pairs would save one barrier per tile.


antiagainst:
This change can be very tricky so let's wait for at least two folks to approve before landing.

qedawkins (dismissed):
LGTM, I had an agent help compare barrier placement in the asm before and after for each and verified that in all the cases that weren't direct drops of duplicate barriers, other barriers were covering for the removed ones. Less familiar with the gfx1250 code but looking at the before and after the removals make sense.

antiagainst:
Hmm, this somehow [regresses perf](https://github.com/lightseekorg/tokenspeed/actions/runs/35394514800/job/105787529441?pr=1627). It doesn't quite make sense.

PR #1635, opened 2026-09-18
feat(kimi-k3): add NVFP4 MegaMoE and opt-in autotuning

syuoni:
## Summary

MegaMoE kernels are vendored from TensorRT-LLM: https://github.com/NVIDIA/TensorRT-LLM/blob/d1166c722ad3b81d7809c9ea98c05e0f1f860248/tensorrt_llm/_torch/moe/fused_moe/mega_moe/mega_moe_cute_dsl.py#L361

Adds `--moe-backend mega_moe` for Kimi-K3 NVFP4 with attention DP and MoE EP on Blackwell. The backend replaces the separate routed dispatch/GEMM/combine pipeline with TensorRT-LLM’s CuTeDSL MegaMoE implementation.

* Vendor the kernel sources under tokenspeed-kernel, including licensing and upstream provenance.
* Preserve quantization before dispatch, SiTU activation, routing weights after FC2, and BF16 outputs.
* Add optional FlashInfer autotuning over 36 tactics using CUDA graph timing. Enable with MEGAMOE_TACTIC_AUTOTUNE=1; heuristic selection remains the default.
* Reuse one maximum-sized workspace with stable offsets across token capacities and tactics.
* Unify DeepEP and MegaMoE process-group plumbing through the optional process_group argument.
* Add four-node DP16 AIME26 CI coverage and update TP8 to EAGLE3. These CI configurations use flashinfer_trtllm.
MegaMoE requires --all2all-backend none, since it owns dispatch and combine. At EP16 with an 8,192-token capacity, its workspace reserves approximately 21.5 GB per rank.


tokenspeed-kernel/python/tokenspeed_kernel/thirdparty/cute_dsl/mega_moe/LICENSE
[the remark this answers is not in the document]
syuoni: Done. Thanks for the reminder.

tokenspeed-kernel/python/tokenspeed_kernel/ops/moe/mega_moe.py
```diff
@@ -0,0 +1,165 @@
...
+import torch
+from tokenspeed_kernel.platform import ArchVersion, CapabilityRequirement
+from tokenspeed_kernel.registry import Priority, register_kernel
+from tokenspeed_kernel.signature import format_signatures
+
+
+def trtllm_mega_moe_process_weights(plan: dict, w: torch.nn.Module) -> None:
+    from tokenspeed_kernel.thirdparty.cute_dsl.mega_moe.blocked_scale import to_blocked
+
+    del plan
```
byshiue, line 32: Why we pass plan into the function and delete it directly?

syuoni: It’s a registered weight-preprocessor callback.

The dispatcher always calls: `preprocessor(plan=plan, w=w)`. So the callback must accept plan, including that parameter name because it’s passed by keyword.

PR #1638, opened 2026-09-18
feat(kimi-k3): fuse sharded latent down proj with NVFP4 quantize

syuoni:
## Summary

Fuse the Lamport gather after Kimi-K3’s latent down projection with NVFP4 activation quantization, removing an intermediate BF16 write/read and a separate kernel launch before the MoE.
- Add a CuTe DSL consumer that quantizes ready mailbox fragments directly into packed NVFP4 values and block scales, then resets the consumed slots.
- Integrate into the existing latent-down operator and pass the standard (values, scales) tuple to the TRT-LLM MoE backend.
- Enable automatically for supported Blackwell configurations with up to 1,280 tokens, preserving the existing fallback elsewhere. Support PDL and prepare the kernel before CUDA graph capture.

Eight-GPU GB200 TP8 microbenchmarks with PDL showed 6.6–19.0% lower combined down-projection and quantization latency across 8–1,280 tokens.

Validated with 28 kernel tests and 168 runtime tests, plus eight-GPU bitwise comparisons against the unfused baseline in eager execution and CUDA graph replay.


syuoni:
All touched UT and K3 e2e tests passed locally, merging.

PR #1641, opened 2026-09-18
perf(amd): give gfx1250 MXFP4 decode's spare warp bit to N

jerryyin:
`get_wmma_layout` always spent one warp bit stepping M. At the BLOCK_M the decode launcher picks for low expert occupancy (16 or 32), one warp's registers already span every row the block has, so that bit addressed rows past the tile and half the warps issued WMMA over nothing.

Spend it on N instead whenever block_m is within the rows one warp's registers cover; prefill at BLOCK_M=64 is unchanged. Warp bases now double out from tiles_per_warp, continuing the sequence the register bases start, which also replaces the hardcoded 4/8 branch.

At BLOCK_M=16 this halves v_wmma_* in _matmul_decode, 16 -> 8. The payoff is tuning-dependent: paired decode timing puts this layout 9.00% ahead of the same kernel without it at BLOCK_K=512.


tokenspeed-kernel-amd/python/tokenspeed_kernel_amd/ops/gfx1250/moe/mxfp4/_common.py
```diff
@@ -298,17 +298,24 @@ def get_wmma_layout(num_warps, packed, use_wmma_scaled, scale_preshuffle):
...
+    # Warp bits continue the doubling the register bits started, so the first
+    # one steps N by the tiles_per_warp tiles a warp's registers already cover.
+    warp_bases = [
+        [0, tiles_per_warp << bit] for bit in range(num_warps.bit_length() - 1)
+    ]
+
+    # One warp's registers already span tiles_per_warp tiles of M. A warp bit
+    # stepping M beyond that covers rows the block does not have, so below the
+    # threshold every bit goes to N instead.
+    split_m = block_m > tiles_per_warp * WMMA_INSTR_SHAPE[0]
```
knwng: Should we put heuristics somewhere out of kernel and leave this function to be a pure layout generator?

knwng, line 315: We can also apply similar logic to choose `tiles_per_warp` depending on `block_m`. When `block_m` is small we don't need `tiles_per_warp` > 1.

tokenspeed-kernel/test/amd/ops/moe/test_gluon_mxfp4_contracts_gfx1250.py
```diff
@@ -264,3 +265,40 @@ def test_partial_tdm_hints_exactly_the_operands_it_fuses(
...
+    """Tile coordinates reached by flipping any subset of ``bases``."""
+    reached = {(0, 0)}
+    for base in bases:
+        reached |= {(m + base[0], n + base[1]) for m, n in reached}
+    return reached
+
+
+@pytest.mark.parametrize("num_warps", [4, 8])
+@pytest.mark.parametrize("block_m", BLOCK_M_VALUES)
+def test_no_wmma_warp_starts_past_the_block(block_m: int, num_warps: int) -> None:
```
knwng, line 292: Should we also check block_n?

borontion:
@jerryyin are you still working on this?

borontion (approved):

PR #1652, opened 2026-09-20
feat: support tp shard on MoE shared expert 

byshiue:
# Add independent shared-expert TP with ordered communication overlap for Kimi-K3

## Summary

This PR lets Kimi-K3 shard its BF16 shared-expert MLP independently of attention and routed experts. In DEP16, attention stays at TP1/DP16 and routed experts stay at TP1/EP16. The main benefit is lower replicated weight memory and more room for cache. End-to-end latency depends on the workload; the TP4 measurements below do not show a consistent speedup.

For example, enable TP4 shared experts with:

```bash
export TOKENSPEED_KIMI_K3_SHARED_EXPERT_TP_SIZE=4
```

Unset or `1` retains the original shared-expert path. An enabled TP size must be greater than one, smaller than world size, and divide both world size and the shared MLP intermediate width. DEP16 can therefore use TP2, TP4, or TP8. Ranks agree on the raw setting before creating TP groups, including disabled or malformed settings, so mismatches fail together rather than leaving peers waiting in group creation.

Sharding uses contiguous TP groups and loads gate/up and down weight shards directly. Attention and cache ownership do not change. The current integration requires attention TP1/DPworld, routed MoE TP1/EPworld, and PP1.

## Implementation

- Run `AllGather → gate/up → activation → down → ReduceScatter`, restoring each rank's local token ownership before adding the shared and routed outputs. There is no intermediate A2A or duplicate down-projection reduction.
- Start shared AllGather on the auxiliary stream while the main stream prepares routed inputs. An event after AllGather allows routed dispatch to proceed while shared GEMMs finish.
- Finish shared GEMMs before routed BMM to avoid competing GEMM workloads. Run shared ReduceScatter on the auxiliary stream after dispatch, overlapping routed BMM, and join it before combine.
- For MegaMoE, which owns dispatch/combine inside the expert call, finish shared ReduceScatter before that call instead of overlapping its communication.
- Keep shared collectives ordered with routed dispatch/combine. Separate buffers alone do not prevent peer-polling kernels from starving each other of GPU resources.
- Prepare one `SharedExpertCommunication` instance, reused by sequential shared-expert layers, and stream events before CUDA-graph capture and cache budgeting. It owns padding, backend selection, and restoration of local token ownership; its scratch is separate from other communication workspaces.
- Use TRT-LLM one-shot AllGather and Lamport ReduceScatter for up to 128 padded rows per rank on CUDA-IPC-accessible groups. Native kernels support TP2/4/8/16; this wrapper requires positive, 128-aligned hidden widths and enforces Lamport capacity limits. Group size, buffer dimensions, and token counts come from runtime parameters. AllGather derives an aligned view width within the native 2112-element limit, rather than special-casing Kimi-K3's hidden size.
- Use NCCL for larger batches or unsupported TP/width combinations. Empty token owners still join an active subgroup; an entirely empty subgroup skips both collectives. Native initialization errors are not silently converted into rank-local fallbacks.
- Keep the plain collective wrappers in the existing `tokenspeed_kernel.ops.communication.trtllm` module: `TrtllmAllGatherState`, `TrtllmReduceScatterState`, `trtllm_allgather`, and `trtllm_reduce_scatter`. Each state owns an independent IPC workspace rather than sharing the existing fusion workspace. AllGather returns borrowed storage; ReduceScatter returns owned output. Their `stateful_allgather` and `stateful_reduce_scatter` registry modes stay separate from stateless auto dispatch.

No scheduler or cache-layout changes are required. Deployment guidance is in `docs/recipes/kimi-k3-shared-expert-tp.md`, with a configuration note in `docs/serving/parallelism.md`.

## Benchmark setup

These are prior TP4 measurements, collected before workspace cleanup, runtime-geometry generalization, and the latest wrapper consolidation. E2E performance and memory have not been remeasured on the current revision (`7ab54b8b`). The tables remain useful reference results, but do not validate the current code or the newly enabled TP2/TP8 native paths.

- 16 NVIDIA GB300 GPUs; each DEP16/TP4 pair runs sequentially on the same allocation.
- Real Kimi-K3 NVFP4 checkpoint; shared-expert weights remain BF16. FP8 KV cache, CUDA graphs enabled, no speculative decoding or autotuning, FlashInfer routed-MoE communication.
- Attention TP1/DP16, routed MoE TP1/EP16. Only shared-expert TP changes from 1 to 4.
- `gpu_memory_utilization=0.83`, 8,192-token prefill chunks, fixed inputs and per-rank request affinity.
- Cold input: 1,024 tokens. Incremental input: 1,024 cached tokens plus 256 new tokens. Both generate 128 output tokens per request.
- One warmup and three measured rounds; 10-ms request-wave pacing. No profiler is attached during timing.

C32/C64/C128 mean requests **per rank**, or 512/1,024/2,048 requests total.

C32 uses the full 93-layer model. Full-model C64 failed cache admission in both variants, so C64/C128 were explicitly tested with a **48-layer dummy model retaining real checkpoint weights**. This is the nearest approximately half-depth configuration compatible with the existing cache layout; it is not a random-weight model. Full-model C128 was not run. The dummy configuration is a benchmark-only exception, not a new default.

## End-to-end performance

Workflow latency includes admission, prefill and decode, excluding model loading, JIT compilation and artifact writes. Values are three-round means.

| Requests/rank | Model | Cold DEP16 | Cold TP4 | Incremental DEP16 | Incremental TP4 |
|---|---|---:|---:|---:|---:|
| 32 | Full, 93 layers | 13.14 s | 14.07 s | 9.17 s | 9.10 s |
| 64 | Dummy, 48 layers | 13.00 s | 11.04 s | 8.24 s | 8.58 s |
| 128 | Dummy, 48 layers | 18.61 s | 19.27 s | 13.26 s | 12.42 s |

Do not use these mixed-depth rows to infer concurrency scaling or estimate full-model C64/C128 latency. The full-model C32 reference used a different allocation from the dummy-model runs.

Decode measurements, in milliseconds per step:

| Requests/rank | Model | Cold DEP16 → TP4 | Incremental DEP16 → TP4 |
|---|---|---:|---:|
| 32 | Full, 93 layers | 51.06 → 49.02 | 48.70 → 46.99 |
| 64 | Dummy, 48 layers | 30.95 → 30.85 | 31.40 → 31.26 |
| 128 | Dummy, 48 layers | 42.00 → 42.51 | 42.01 → 42.51 |

The C32 reference uses steady-decode intervals inside the common active-request window. Dummy runs use worker-local full-batch intervals inside each node's active-request window, excluding local prefill/mixed dispatch. Node clocks were offset, so timestamps were not compared across nodes. All 16 ranks contributed valid intervals in every measured dummy-model round. These are serving-loop measurements, not isolated shared-MLP kernel timings.

Full-model C32 decode improves by approximately 3.5–4.0%, but its cold workflow is slower. Dummy-model C64 decode is effectively tied, and C128 TP4 decode is approximately 1.2% slower. Workflow variance is substantial: C64 incremental latency ranges from 6.42–11.34 s for DEP16 and 6.40–10.26 s for TP4. These sequential three-round tests do not establish a blanket E2E speedup.

## Memory savings and cache capacity

All values below are per GPU. Weight totals include attention, routed MoE, dense/shared MLP, embeddings and normalization weights.

| Configuration | Weight memory DEP16 → TP4 | Weight savings | Allocated cache DEP16 → TP4 | Cache increase |
|---|---:|---:|---:|---:|
| Full model, C32 | 161.28 → 144.30 GiB | **16.98 GiB** | 60.87 → 75.09 GiB | **23.4%** |
| Dummy, C64 | 85.46 → 76.78 GiB | **8.68 GiB** | 137.50 → 143.51 GiB | **4.4%** |
| Dummy, C128 | 85.46 → 76.78 GiB | **8.68 GiB** | 137.50 → 143.51 GiB | **4.4%** |

Usable LCM parent blocks increase from **414 to 511** for the full model and **1,874 to 1,956** for the dummy model. These blocks provide storage for attention cache and recurrent states; their increase is not a direct percentage increase in maximum request concurrency.

At a fixed memory-utilization setting, the runtime uses much of the saved weight memory for cache. Communication buffers and other workspaces consume part of the savings. Exact device-free memory immediately before cache allocation was not separately recorded.

| Configuration | Peak device memory DEP16 → TP4 |
|---|---:|
| Full model, C32 | 259.59 → 259.73 GiB/GPU |
| Dummy, C64 | 272.54 → 273.84 GiB/GPU |
| Dummy, C128 | 273.20 → 275.35 GiB/GPU |

Peaks are sampled once per second across loading, warmup and measured rounds, taking the maximum observed across all GPUs. They include cache, allocator pools and non-Torch allocations and can miss shorter spikes. Similar or higher peak use does not negate the measured weight savings.

### Coverage in the current code

`test/runtime/distributed/validate_kimi_k3_shared_expert_tp.py` consolidates the shared-expert communication checks into a GPU validator; the separate CPU communication test and single-use mapping helper were removed. The validator covers:

- Real cross-rank setting agreement, including disagreement with disabled or malformed ranks.
- Exact checkpoint weight shards and numerical comparison against the unsharded BF16 MLP, with maximum cross-rank relative L2 error below 1%.
- Balanced, uneven, and empty owners/subgroups; retained-output lifetime; repeated auxiliary-stream CUDA-graph replay with changing inputs and delayed peers.
- Actual collective participation and native/NCCL selection at the 128/129-row boundary, observed during eager warmup without substituting collective results.
- Basic AllGather/ReduceScatter correctness at hidden widths 128, 2304, and 7200, using one token per rank in eager mode. These cases exercise a narrow view, row subdivision, and an unaligned NCCL fallback without repeating the full boundary/graph matrix.

The existing model orchestration and stream-ordering tests remain. They check call parameters and event ordering; they are not substitutes for real GPU communication tests. Run the distributed validator with explicit `--model MODEL_DIR --layer 1 --tp-size 4`, and repeat with TP2/TP8 on a compatible 16-GPU topology.

### What has been verified

Full `pre-commit run --all-files` checks passed on the current revision. AST comparison confirmed that the wrapper move/rename preserved its computation and left existing TRT-LLM function/class definitions unchanged. GPU correctness, CUDA-graph execution, and E2E performance have **not** been rerun after the recent changes. TP4/H7168 has prior GPU validation; newly enabled native geometries still need validation on the target topology.

For the reported dummy-model runs, execution checks passed for 8,192 requests per variant at C64 and 16,384 per variant at C128, including warmup. Every request produced 128 output tokens with finite log-probabilities; streamed and final token sequences agreed. Every incremental request reused 1,024 cached tokens. All four dummy-model cases completed without OOM.

These are execution-integrity checks, not a dataset-accuracy evaluation or a claim of identical cross-variant generation. Sharding changes BF16 reduction rounding. No AIME or other task-accuracy score is claimed for these benchmark runs.

Full-model C64's failure was insufficient configured cache/state admission capacity after loading weights, not a weight-loading CUDA OOM. Higher full-model concurrency remains unvalidated with this configuration.


byshiue:
@lightseek-bot Can you help to review this feature? 
I will add this feature for attention later in another PR.

PR #1671, opened 2026-09-20
refactor(kernel): trim benchmark validation to the trust boundaries

antiagainst:
Generalize the benchmark documentation so it no longer anchors on the initial GEMM suite, generator, validator, or introducing pull request.

Keep input validation where untrusted data enters (suite JSON, generator parameters, the PR comment artifact) and drop the layers that re-checked in-process data produced by our own code:

- harness: take a `GraphTimer` instead of a config/timer either-or; stop cross-checking the generator's selected registration; drop the `backend_unavailable` status that only the gemm generator's registry re-query could produce.
- validation: `runs` lives on `PreparedValidation` instead of each output spec, so the harness no longer computes a maximum and slices per output; the `close` validator reads its tolerances instead of re-validating them; drop the arbitrary 100-run cap and the float-only dtype list.
- graph timer: allocate a fresh side stream per measurement and release the graph best-effort on failure instead of tracking stream state and attaching exception notes.
- worker: drop the self-consistency check that a successful result echoed the suite's timer and environment, and the temp-file dance for one JSON.
- coordinator: validate only what it keys on (schema, revision, timer, environment, case ids); the worker already guarantees sample shape, registration, and correctness of a success. A merge base without the suite is a bootstrap regardless of whether it has the runner, so adding a second suite file no longer fails as "not an initial bootstrap".
- gemm generator: parse dtype as a name only, map every selection miss to `not_applicable`, and stop re-deriving tensor layouts to check them against the registration that was already selected on those traits.


antiagainst:
Before landing, @Max191 please take a look of this change. It tries to drop some excessive error checking and over abstraction--this infra is meant for internal use anyway so we can avoid some overheads.

Max191 (dismissed):

PR #1678, opened 2026-09-21
refactor(kernel): name Gluon kernels after their registration

antiagainst:
Proton names a launch after the @gluon.jit function it compiled, while override= strings, describe_kernel and the dispatcher scopes use the register_kernel(name=...) value. The two never matched: the AMD kernels were private `_foo_kernel` defs and the Python launcher around each one carried the registration name instead.

Make one name thread the whole stack for Gluon kernels: the registration name, the registered Python def, and the @gluon.jit (or @triton.jit) kernel that does the op's work are now identical. The AMD launcher is renamed to `launch_<name>` so a function named like a registration in tokenspeed_kernel_amd is always the kernel. Companion kernels launched only by that op insert a role before the arch suffix
(gluon_dsv4_decode_reduce_gfx1250, gluon_mha_prefill_sliding_gfx950). Kernels shared by several registered ops (radix top-k, DSA dense attention, MLA decode, routing/sorting/quantize and the MoE stage pipelines) keep their descriptive names.


borontion (commented):
shall we just use `launch_<name>` instead of `launch_gluon_<name>`? I assume all kernel launchers there are gluon anyway. just want to keep the naming compact.

antiagainst:
> shall we just use `launch_<name>` instead of `launch_gluon_<name>`? I assume all kernel launchers there are gluon anyway. just want to keep the naming compact.

Simplified consistency rule is preferrable here as it's very clear we just add a `launch_` prefix. Also a bit verbose here helps to really see more context in various profilers for both CPU side (which shows `launch_gluon_*` names) and GPU side (which shows `gluon_*` names). This is subtle ergonomics for development but it pays dividends. :)

borontion (dismissed):

test/ci_system/install_deps.sh
```diff
@@ -189,7 +189,14 @@ pip_install_with_retry pip3 install tokenspeed-kernel/python/ --no-build-isolati
 # ============================================================
 echo "=== Step 5: Install TokenSpeed Scheduler ==="
 pip_install_with_retry pip3 install cmake ninja
-pip_install_with_retry pip3 install tokenspeed-scheduler/
+# Scheduler changes intentionally accumulate without a version bump. Build in
```
borontion, line 192: is this change intended?

antiagainst: Yes. Some runtime tests were broken without this.

PR #1683, opened 2026-09-21
perf(dsv41): select Reindex rows from the compacted candidate pool

wangbo981016:
## Motivation

V4.1's hierarchical sparse indexer designates one candidate-source layer that scores the full history and publishes its best 2048 blocks; the later Reindex layers are meant to select only within that pool.

On Hopper the restriction was implemented by masking: a Reindex pass scored the whole history, allocated a full-width boolean table, wrote a masked copy of every score with non-candidates set to `-inf`, and then ran Top-K across all 126016 columns to find 512 rows that could only ever come from 16384 of them. The mask keeps the result correct, but it does not shrink anything the selector has to look at, so the hierarchy bought no time on sm90 — a Reindex pass cost the same as a candidate-source pass.

For reference, the other three solutions already avoid part or all of this: the portable Triton scan and the AMD Gluon scan bound their scan width by the pool, and the Blackwell solution compacts the pool before selecting. Hopper was the one path that did neither.

## Changes

**Select from the pool, not from a masked history.** `candidate_scores` gathers the candidate blocks' scores into a `[tokens, blocks * 8]` table in one Triton kernel, replacing the boolean table and the masked copy. Over a 126K history with a 2048-block pool, Top-K now reads 16384 columns instead of 126016. Column `c` holds row `candidates[c // 8] * 8 + c % 8`, so winners are resolved arithmetically and no id table is materialized.

**Fuse the selection tail.** `_finish_topk` ran six elementwise ops plus a sort over `[tokens, 512]`, entirely launch bound. One Triton kernel now sorts, packs the valid prefix, pads with `-1`, counts the survivors, and resolves pool columns to row ids. It is shared with the portable solution, so both paths benefit.

**Write the empty candidate pool.** With `candidate_topk == 0` — every Reindex pass — the Hopper path never wrote `blocks` or `block_lengths`, so a caller supplying uninitialized buffers read back garbage. The portable, AMD and Blackwell solutions all write them; the runtime only escaped this by zeroing its own lengths buffer. Found by the new test.

## What this does not change

Scoring. A Reindex pass still scores the full history, because DeepGEMM ships no sm90 sparse paged MQA logits kernel (only an sm100 one). Two alternatives were measured and rejected:

- Gathering the pool into a dense buffer for the dense scorer costs 277 us for 8x16384 rows, roughly eight times the 33 us of scoring it would save: the gather is fully scattered (62 GB/s) and it destroys the sharing of K across queries that MQA scoring relies on.
- Routing Reindex to the portable scan, which does bound its width by the pool, runs at 194 us against 83 us for the current path — it reads eight times fewer rows and is still 2.3x slower.

The remaining fix is an sm90 sparse paged MQA logits kernel upstream.

## Results

sm90, 8 query tokens over a 126K history, timed under CUDA graph replay:

| | before | after |
|---|---|---|
| Reindex `index_topk` | 214.7 us | **83.1 us** |
| Candidate-source `index_topk` | 282.1 us | **200.4 us** |
| Selection per decode step (1 source + 4 Reindex) | 1141 us | **533 us** |

Per-call kernel times in a served decode with speculative decoding:

| kernel | before | after |
|---|---|---|
| FlashInfer Top-K | 51.47 us | 27.46 us |
| `radixSortKVInPlace` (two variants) | 24.65 / 26.29 us | replaced |
| fused selection tail | — | 4.28 us |
| paged MQA logits (scoring) | 22.91 us | 22.92 us |

Cross-check: the served average moves less than the microbenchmark because only four of the nine Top-K calls per step are Reindex passes. Backing that out gives 54 us saved per Reindex Top-K, against 56 us measured in isolation.

These are kernel-level numbers. The indexer runs on an auxiliary stream overlapped with the main stream, so no end-to-end latency claim is made here.

## Tests

`tokenspeed-kernel/test/ops/test_attention_dsv41.py`: 103 passed, 1 skipped.

Adds `test_reindex_maps_every_candidate_block_order_back_to_row_ids`, parametrized over both index formats. It feeds an unsorted pool with `-1` padding in the middle and a Top-K capacity above the pool size, so every visible candidate row must come back, sorted — pinning the pool-column to row-id mapping on each format's scorer independently of scores.

Accuracy against earlier runs on the same configuration:

| benchmark | before | after |
|---|---|---|
| GSM8K (100, greedy) | 0.97-0.98 | 0.98 |
| GPQA Diamond (198, sampled) | 0.8939 | 0.8990 |

GPQA runs in the official instruct setting, which samples at temperature 1.0, so its standard error over 198 questions is around 2 points. These establish no detectable regression rather than an improvement. Selection is unchanged in which rows it can return; the only thing this patch can perturb is the order among equal scores, which `index_topk` already documents as unspecified.


wangbo981016:
@codex review

PR #1691, opened 2026-09-21
perf(kernel): wire PDL into the activation Triton kernels

tuanzhangCS:
## Summary
Add Programmatic Dependent Launch support to the eight activation kernels that lacked it: fused_gate_sigmoid_mul_add, sigmoid_mul, silu_and_mul, swiglu_oai, situ_and_mul, add3, attnres_partial and attnres_partial_dual.

Each kernel gains an ENABLE_PDL constexpr that runs gdc_wait followed immediately by gdc_launch_dependents at entry, so a dependent grid can launch early and overlap its prologue with the producer grid. Each wrapper accepts enable_pdl (defaulting to the platform setting) and passes launch_pdl when enabled.

## Test Plan
- Add PDL-vs-serial bitwise parity coverage for every wired kernel.


tuanzhangCS:
@claude review it

borontion:
I think you can check whether pdl is enabled using this in tokenspeed-kernel without adding it as an argument for ever call

https://github.com/lightseekorg/tokenspeed/blob/bce6d20a0e64138c8a9efa83a1f522d7a6a671b7/tokenspeed-kernel/python/tokenspeed_kernel/platform.py#L333

tuanzhangCS:
> I think you can check whether pdl is enabled using this in tokenspeed-kernel without adding it as an argument for ever call
> 
> https://github.com/lightseekorg/tokenspeed/blob/bce6d20a0e64138c8a9efa83a1f522d7a6a671b7/tokenspeed-kernel/python/tokenspeed_kernel/platform.py#L333

good suggestion

borontion (approved):

PR #1693, opened 2026-09-21
perf(kernel): Add an optional split-KV floor to MLA decode

wzhao18:
## Summary
Add optional parameter `min_split_kv` to `tokenspeed_mla_decode` to force a minimum number of KV splits. The default value is `1` and thus does not change the existing heuristics. Forcing a minimum number of KV splits shows to improve by up to ~4x in microbenchmark, and more than 30% speedup in e2e agentic benchmark.

### Microbenchmark
Kimi K3 TP=1 attention shapes:
 - “Balanced” means every request in the batch has the same context length:
 - “Imbalanced” uses geometrically decreasing context lengths. For 256K with batch 8: [256K, 128K, 64K, 32K, 16K, 8K, 4K, 2K]
mla_min_split_kv_q5_sweep_20260921

### E2E Agentic Benchmark
Kimi K3 DCP8 prefill + DEP16 decode (vllm):
| Validation | Automatic | Min-8 | Improvement |
|---|---:|---:|---:|
| Throughput TPS/GPU | 10,277 | 13,840 | +34.67% |
| Mean ITL | 34.77 ms | 18.57 ms | −46.59% |
| p90 ITL | 40.12 ms | 20.74 ms | −48.31% |


tokenspeed-mla/python/tokenspeed_mla/mla_decode.py
```diff
@@ -601,8 +615,19 @@ def tokenspeed_mla_decode(
...
 
+    # Public FP8 auto-splitting is bounded by 64 (M64) or 32 (M128).
+    # Both values are in the compile cache key, including across batches.
+    reducer_max_splits = 256
+    if is_fp8:
+        reducer_max_splits = 64 if mma_m_tile == 64 else 32
+        if min_split_kv > reducer_max_splits:
+            reducer_max_splits = 64
+        if min_split_kv > 64:
+            reducer_max_splits = 256
```
dishengbin, line 634: Can you please remind me why `reducer_max_splits` should be 256 instead of 128 when `min_split_kv > 64`?

dishengbin:
Overall, it looks good to me. Thanks for the changes.

tokenspeed-mla/python/tokenspeed_mla/mla_decode.py
wzhao18: I don't expect to use very large values (e.g., > 64) for `min_split_kv` as we just want to make sure split_kv is not too small. Given we still accept `min_split_kv` within 256, the code just makes sure  `reducer_max_splits` is no less than that.

dishengbin: Makes sense, thanks

wzhao18:
@dishengbin Please help start the CI testing. Thanks!

PR #1702, opened 2026-09-21
fix(kernel): remove deepgemm BF16 JIT patch

borontion:
## Summary

Experimental change to drop the patch of deepgemm mega moe.


borontion:
hi @XucSh, I want to clean up this this ad-hoc deep gemm patch. and as shown in the ci, removing it  does not regress accuracy. but let me know if there is something not covered in ci. thanks.

PR #1718, opened 2026-09-22
feat(dcp): support GPU MLA and DSA decode context parallelism

OftenDream:
## Summary

Extend GPU decode context parallelism (DCP) to ordinary MLA, hybrid MLA/KDA, and DSA, and shard the existing DeepSeek V4 Index-K cache. Reuse the existing physical cache pools and scheduler-owned cache groups instead of introducing parallel sharded-pool implementations.

This is a draft for design review. GPU kernel/reference checks pass, but model-level numerical validation is incomplete, especially for DSA. This is not a claim of bitwise equivalence across DCP sizes.

## Backend scope

| Attention path | DCP implementation | Limits |
| --- | --- | --- |
| Dense MLA with `flashmla` | Sharded decode; bounded history reconstruction for prefill | Ordinary MLA and MLA/KDA; KDA state stays replicated. Dense FlashMLA still needs supported hardware; this PR does not add Blackwell dense FlashMLA support. |
| GPU DSA, Triton attention | Sparse prefill/decode with local output + LSE | Supports the registered regular and packed latent-cache formats. |
| GPU DSA, FlashMLA attention | Sparse prefill/decode with local output + LSE | Prefill accepts regular BF16 KV; sparse decode requires the supported packed FP8 format. Regular BF16 decode continues to use its matching implementation, e.g. Triton. |
| Existing `deepseek_v4` | Extend Index-K placement/selection to independent context shards | Compressed attention KV remains sharded; SWA and compressor states remain replicated. This is not the separate `deepseek_v41` backend. |

Index scoring uses DeepGEMM for supported NVIDIA SM90+ packed-FP8 configurations, with a portable Triton implementation behind the same candidate API.

**Blackwell is backend-dependent, not a blanket GPU support claim.** The FlashInfer/TRTLLM sparse attention path currently advertises `return_lse=False`, so it cannot supply DCP partials. Selection must find a compatible LSE-capable implementation; registering a Hopper+ kernel is not an end-to-end Blackwell model validation. Dense FlashMLA coverage remains SM90-only in the reported tests.

PD cache transfer and KVStore are not supported with DCP. The new FlashMLA/GPU DSA paths reject speculative execution. This PR does not extend Ascend DCP.

### Placement and storage

- Recipes declare each cache group's `shard_count`. The arena remains the source of physical storage geometry; the scheduler continues to allocate virtual blocks through the existing cache-group contract.
- `CachePlacement` describes ownership and virtual-to-local translation. `resolve_cache_slots` returns physical slots plus an ownership mask. Writers mask both source loads and destination stores: foreign rows must not write to a dummy/null slot.
- Keep the existing `set_mla_kv_buffer` / `get_mla_kv_buffer` pool interfaces and physical pool implementations. Placement and collective reconstruction live above storage, avoiding separate DCP/non-DCP cache-pool classes.
- Shared page-table metadata supplies the layout a kernel needs: position-preserving tables with `-1` holes, or compact owned pages with local lengths and original logical positions. Compaction does not change causal positions.
- Keep recurrent KDA state replicated. Give DSA Index-K its own cache group and ownership geometry; do not assume its virtual IDs equal attention-KV IDs. Capacity follows physical demand, so replicated groups can remain the limiting resource.

### Prefill and decode

**Dense MLA:** prefill reconstructs only the requested, bounded history chunk in logical order, then runs the existing TP-local attention path. Decode gathers actual query heads after normalization/RoPE, attends to local KV pages, and merges context partials.

**DSA:** both prefill and decode select a global sparse candidate set, map it to each rank's owned cache slots, gather query heads, and compute local sparse attention. This is context sharding, not query-context parallelism (QCP).

For local attention outputs `O_r` and natural-log LSE `L_r`, the shared merge weights each partial by `exp(L_r - logsumexp(L))`. It gathers FP32 LSE, weights partials in FP32, and reduce-scatters to the TP head owner before restoring the output dtype. Empty shards contribute zero output and `-inf` LSE; a sink, when present, is applied once after reconstruction. FP32 merge arithmetic does not undo rounding in a BF16 local output. No experimental FlashMLA FP32-output kernel is included.

### Distributed Indexer Top-K

Each rank scores the same queries against its local Index-K pages and retains up to K candidates. Candidate positions are mapped back to global logical offsets; positions and FP32 scores are all-gathered, then **one global `torch.topk`** selects K from the union. No full Index-K replication is required for this selection.

Local Top-K is sufficient because an entry excluded from a rank's local K cannot be required for a global top-K set, apart from interchangeable equal-score ties. Equal-score ordering is intentionally unspecified. Invalid/NaN candidates, padded entries, valid counts, causal tails, and forced initial/local windows are handled explicitly. Bounded query tiling limits temporary score memory for both prefill and decode.

## Validation

The port's recorded validation for `0140ec34` includes full `pre-commit run --all-files`, 297 runtime/cache tests, 63 GPU kernel tests, and a two-rank distributed Top-K plus attention/LSE comparison against unsharded references. Eight runtime tests were skipped, including SM90-only dense FlashMLA tests on Blackwell.

The following **model measurements are from the predecessor implementation `5d9460cd`, before the port to main**, not an end-to-end validation of this PR head. H200, TP4, DCP1 vs DCP4, eager, fixed continuation tokens; no FP32-output experiment patches:

| Model/path | Samples / scored tokens | Target-token logprob MAE | Max absolute error |
| --- | ---: | ---: | ---: |
| DeepSeek V4 | 10 / 320 | 0.000758 | 0.044113 |
| LongCat DSA | 18 / 768 | 0.005642 | 0.199098 |


FC-Li:
Pushed ccda097d to fix the two real CI failures:

- **ut-runtime-1gpu (b200)**: `FlashMLATileScheduleTest` builds its backend stub with `__new__`, so the new DCP fields (`dcp_group` etc.) were never declared and `init_cuda_graph_state` raised `AttributeError`. The stub now mirrors the `__init__` declarations with single-rank values. Verified locally: 24/24 pass.
- **ut-tokenspeed-kernel (amd-mi35x)**: `torch.cuda.is_available()` is also true on ROCm, so `test_flashmla_dsa_dcp.py` ran on the AMD runners and failed with `NoKernelFoundError`. It is now gated on `platform.is_nvidia and platform.is_hopper_plus`, matching the kernel registration predicate and the sibling tests in `test/nvidia/`.

The **mi450-sim** failure was an infra flake (the runner stalled in `pip install` for 35 minutes and got canceled before any test ran); the new push reruns it.

Review to follow.

FC-Li (commented):
Full review of the diff at head `ccda097d`. Overall the design direction is right — `CachePlacement` as address views over the existing pools, cache groups staying scheduler-owned, DCP remaining a parameter of the one decode path — and the guards for the unsupported combinations (PD/KVStore via server_args, speculation in `AttnConfig`, hybrid non-K3 in the registry) all check out. The items below are ordered by severity.

## Must fix

1. **`flashmla.py` — `_forward_absorbed_extend` has no DCP guard.** `forward_extend`'s absorbed branch (`use_ragged=False`) plans the flashinfer prefill wrapper on the router-resolved (virtual) page table and attends the *local sharded* `k_buf` with no LSE merge. DeepSeek/Kimi happen to bypass it through the model-level chunked path, but nothing enforces that; the DSA sibling raises in the analogous case. Reaching this path with `len(self.dcp_group) > 1` should raise rather than silently produce garbage prefill logits.

2. **`dsa/_triton/topk.py:411` — unguarded OOB load under `EXPLICIT_ROWS`.** `base = tl.load(seq_lens + req)` runs unconditionally, but under `EXPLICIT_ROWS` `seq_lens` is the per-token causal-lens buffer (length = tile tokens) while `req` is a request ID that can exceed it. The loaded value is dead (overwritten by the `EXPLICIT_ROWS` branch) so today this relies entirely on Triton DCE-ing the load; if any Triton version keeps it, it is an illegal memory access on the last tile. Move the `base` load inside `if not EXPLICIT_ROWS`.

3. **`recipes/ordinary.py:185` — `_max_packing` is not a shard count.** `num_lcm_blocks` uses `shard_count = self._max_packing(layout)`, which is `max(packing × shard_count)` over groups. Any future ordinary-family layout that packs >1 CacheBlock per parent *without* DCP (e.g. an MHA target+draft two-geometry plan) silently flips the token-limit math from `_capped_parents` floor semantics to ceil-over-"shards". Use `self._shard_counts[FULL_ATTENTION]` like `kimi_k3.py`'s `token_capacity` does — it states the intent and is immune to packing changes.

4. **`dcp/metadata.py:177` — silent realloc on refresh-contract violation.** When `previous.local_page_table` mismatches the new `page_table` in shape/dtype/device/contiguity, `refresh_dcp_page_table_metadata` silently sets `previous = None` and allocates fresh buffers. Under CUDA-graph replay that means the graph keeps reading the *old* buffers while new refreshes write the new ones — a silent wrong-result mode. Every sibling check in the same function (layout/topology/geometry) raises `ValueError` for exactly this class of misuse; this one should too.

5. **`flashmla.py:841` — DCP decode returns a mismatched `(output, lse)` pair.** In the DCP branch `output` is the merged TP-local-head result but `lse` stays the pre-merge gathered-head local LSE. The sole current caller discards `lse`, but any future consumer (MTP verify, sink merge like `forward_extend_chunked`) gets an LSE whose head count and normalization do not match the output. Return the merged LSE or `None`.

## Questions

6. **FP8 query through the DCP all-gather** (`dsa.py` decode ~755 / prefill ~583): when `data_type == float8_e4m3fn` the query is cast to fp8 *before* `gather_query_heads`. Which collective backend covers fp8 here? If the low-latency backend falls back to torch.distributed/NCCL on some topology, does the fp8 all-gather survive? If it needs a `view(uint8)` bridge, better to put it in `gather_query_heads`.

7. **`deepseek_v4.py` `_forward_sharded_indexer` never dispatches DeepGEMM.** It calls `triton_dsv4_index_candidates` / `triton_dsa_index_candidates` directly, so the V4 sharded indexer always scores with the portable Triton kernel — the PR description says "Scoring uses DeepGEMM where supported, with Triton fallback", which only seems true for the GLM DSA path. It also re-implements the tile+merge loop that `dcp/indexer.select_dsa_topk` already provides. Can the non-FP4 branch route through the `dsa_index_candidates` dispatcher and share the loop?

## Nits

8. `dsa/__init__.py` docstring claims Triton dense-KV partials "retain FP32 output with LSE enabled", but `_run_dense_kv`/`_run_packed_kv` always allocate bf16/q.dtype output — the PR's own tests say partials are rounded to BF16 before the FP32 merge. One of doc or kernel is wrong.
9. `dcp/indexer.py:82` — `select_dsa_topk` re-derives the ownership translation of the full page table per layer per step; the FlashMLA path refreshes `CompactDCPMetadata` once per step instead. Worth converging on the refresh-once contract (`docs/design/unified_path.md`) in a follow-up.
10. `dcp/indexer.py:50` — `merge_index_candidates` launches two all-gathers (offsets, scores) per tile; they are same-shape 4-byte tensors and can travel in one collective.
11. `glm5.py:612` — the decode DCP branch hard-codes `max_logits_bytes = 64 MiB` while prefill reads the server-arg-backed budget; the decode bound should use the same knob.
12. `deepseek_v3.py:1262` — `attn_backend.cache_placement(self.attn_mha)` is loop-invariant in the chunked-KV history loop; hoist it above the `for loop_idx ...`.

E2E validation on 8×H20 (SM90) is in progress on my side (GLM-5.2 GPU-DSA DCP1-vs-DCP4 logprob parity, deepseek_v4 Index-K sharding next); will post results here.

FC-Li:
E2E numerical validation at head `ccda097d` on 8×H20-3e (SM90), eager, TP as noted, greedy decoding. Method: fixed prompt set (4 short + 4 long, up to ~12K tokens including a pathological 400×-repeated-sentence prompt), per-token logprobs from `/generate`, plus GSM8K-200 via evalscope at temperature 0.

**Determinism baseline** — same server (DCP1) probed twice with a cache flush in between: 512/512 tokens bit-identical, logprob MAE exactly 0. Every DCP1↔DCP4 difference below is therefore attributable to the topology change alone.

**DeepSeek-V4-Flash-0731-FP8 (deepseek_v4 path, sharded Index-K), TP4, DCP1 vs DCP4:**
- Realistic prompts incl. 1K–8K-token contexts: 64/64 greedy tokens identical, logprob MAE 0.010–0.019 on matching tokens (max 0.52).
- GSM8K-200: **0.985 (DCP1) vs 0.980 (DCP4)** — one question, noise level.
- Short high-entropy prompts and the 400×-repeated-sentence prompt diverge at near-tie tokens (both sides' logprobs in the −0.5…−1.0 band, both continuations coherent). The repeated-text case is consistent with the declared unspecified equal-score ordering in the sharded Top-K: hundreds of near-identical index scores mean the selected candidate *set* legitimately differs per topology.

**GLM-5.2-FP8 (GPU DSA path, BF16 KV, Triton sparse attention), TP8, DCP1 vs DCP4:**
- More greedy divergences than V4-Flash, but every divergence point sits at logprobs −0.55…−0.96 on both sides (genuinely uncertain tokens) and both continuations stay factually correct — e.g. both answer "fox" for "the first animal mentioned" over ~12K tokens of context. Logprob MAE 0.039 over matched prefixes.
- GSM8K-200: **0.990 (DCP1) vs 0.985 (DCP4)** — one question, noise level.

Also note CI's own `eval-deepseek-v4-flash-dcp4-mtp-gsm8k` (B200, 4 GPU, with MTP) passes on this head, which covers the deepseek_v4 DCP path on Blackwell.

**Coverage gaps that remain:** dense MLA (FlashMLA) DCP and the K3 hybrid path were not e2e-validated here — no non-DSA MLA checkpoint was reachable from this cluster (K3 needs 16 GPUs). Given the FlashMLA dense path is entirely new code, it would be good to attach the SM90 numbers mentioned in the description, or add a CI model-test for it like the V4-Flash DCP4 one.

Operational note: the mi450-sim job failure is unrelated to this PR — both attempts died on Triton *compile* timeouts (pytest-timeout 300 s inside `make_ttgir`/`_init_handles`) across a varying set of tests untouched by this diff, on a runner that failed to restore its compile cache; the same job passes on other current PRs (e.g. #1775).

